PR #40, opened 2016-04-06
stubs ExecutionContext and adds LocalSubprocessExecution

### LINKED ISSUE #35: stub execution context
This will first be necessary for the cli, and then for other interfaces (it wasn't necessary for q2d3 because users initiates execution themselves).

```python
>>> e = LocalSubprocessExecution()
>>> e(workflow_instance, params to workflow_instance.to_script) # this would call subprocess
```

gregcaporaso:
fixes #35.

From ``q2d3/demo/analysis-dir`` this can be tested as follows:

```python
from qiime.sdk.execution import SubprocessExecutor
from qiime.sdk import Workflow

w = Workflow.from_markdown('/Users/caporaso/code/diversity/diversity/workflows/feature_table_to_pcoa.md')

e = SubprocessExecutor()
f = e(w, {'feature_table': 'table.qtf', 'phylogeny': 'phylogeny.qtf'}, {'metric': 'unweighted_unifrac', 'depth': '50'}, {'distance_matrix': 'uu-dm.qtf', 'pcoa_results': 'uu-pc.qtf'})
print(f.done())
f.result() # When f.done() == True, this will be a subprocess.CompleteProcess
```

### DIFF
--- a/qiime/sdk/execution.py
+++ b/qiime/sdk/execution.py
@@ -1,0 +1,46 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016--, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import os
+import tempfile
+import shutil
+import subprocess
+from concurrent.futures import ProcessPoolExecutor
+
+
+class SubprocessExecutor:
+
+    _python_executable = "python"
+
+    def __call__(self, workflow, input_artifact_filepaths,
+                 parameter_references, output_artifact_filepaths):
+        input_artifact_abs_filepaths = \
+            {k: os.path.abspath(v)
+             for k, v in input_artifact_filepaths.items()}
+        output_artifact_abs_filepaths = \
+            {k: os.path.abspath(v)
+             for k, v in output_artifact_filepaths.items()}
+        job = workflow.to_script(input_artifact_abs_filepaths,
+                                 parameter_references,
+                                 output_artifact_abs_filepaths)
+        temp_dir = tempfile.mkdtemp()
+        pool = ProcessPoolExecutor(max_workers=1)
+        py_filename = os.path.join(temp_dir, 'job.py')
+        with open(py_filename, 'w') as py_file:
+            py_file.write(job.code)
+        # TODO: handle subproccess exceptions
+        future = pool.submit(subprocess.run,
+                             [self._python_executable, py_filename],
+                             stdout=subprocess.PIPE,
+                             stderr=subprocess.PIPE)
+        # TODO: handle callback exceptions
+        # TODO: make sure that tempdir is cleaned up even if there is an
+        # exception in pool.submit or the callback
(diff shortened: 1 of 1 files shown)


qiime/sdk/execution.py
```diff
@@ -0,0 +1,44 @@
...
+        job = workflow.to_script(input_artifact_abs_filepaths,
+                                 parameter_references,
+                                 output_artifact_abs_filepaths)
+        with tempfile.NamedTemporaryFile(mode='w', delete=False) as of:
+            of.write(job.code)
+            of.close()
+            cp = subprocess.run([self._python_executable, of.name],
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
+            os.remove(of.name)
```
wasade: does `subprocess.run` guarantee the `fork` starts before it's return? if not, there may be a race condition

wasade: what about `delete=True` and allow the context manager to handle the delete? the race condition may still be a possibility though

gregcaporaso: That won't work, as I need to refer to it by name. It's probably too early for a review of this PR now though. We're discussing some design stuff related to this tomorrow.

qiime/sdk/execution.py
```diff
@@ -0,0 +1,44 @@
...
+        input_artifact_abs_filepaths = \
+            {k: os.path.abspath(v)
+             for k, v in input_artifact_filepaths.items()}
+        output_artifact_abs_filepaths = \
+            {k: os.path.abspath(v)
+             for k, v in output_artifact_filepaths.items()}
+        job = workflow.to_script(input_artifact_abs_filepaths,
+                                 parameter_references,
+                                 output_artifact_abs_filepaths)
+        with tempfile.NamedTemporaryFile(mode='w', delete=False) as of:
```
jairideout: This is nitpicky, but can you use a more descriptive variable name than `of`? Also, why are you using a tempfile context manager but closing and deleting the file within the context?

qiime/sdk/execution.py
```diff
@@ -0,0 +1,44 @@
...
+            of.write(job.code)
+            of.close()
+            cp = subprocess.run([self._python_executable, of.name],
```
jairideout: Same comment for `cp`, use better name

qiime/sdk/execution.py
gregcaporaso: The closing needs to happen so all of ``job.code`` gets written before trying to execute the script (I'm referring to it by ``name`` on the next line). Similar for deleting - if I use the default of ``delete=True``, the file no longer exists when I try to execute it. I guess this defeats the purpose of the context manager - think I should just do it without the context manager?

qiime/sdk/execution.py
gregcaporaso: @wasade, re: your first question ``subprocess.run`` blocks until the subprocess complete, so I think the answer has to be yes. This is now the [recommended](https://docs.python.org/3/library/subprocess.html) way of using ``subprocess`` in Python 3.5 and later, so I was going with that. Let me know if there is a problem with it though.

wasade: Ah, fantastic

Issue #31, opened 2016-03-11
create a mechanism for importing data into artifacts


gregcaporaso:
I have an [old gist](https://gist.github.com/gregcaporaso/edbe4135dd870bb606f9) that illustrates how to do this - that code may or may not still work, but the idea is the same. 

As of now, on import ``provenance`` should be ``None``, but we likely want to provide some provenance information for these, such as when the file was uploaded, the source file path, etc.

jairideout:
Copying these instructions for importing an artifact from q2d3 README.md since they're about to be deleted:

```python
from feature_table.artifact_types import FeatureTable, Frequency

import biom
biom_path = "./otu-table.tsv"
table = biom.load_table(biom_path)

from qiime.sdk import Artifact
Artifact.save(table, FeatureTable[Frequency], None, "./example-table.qtf")
```

gregcaporaso:
This might makes sense as a workflow in the ``q2_types`` plugin.

jairideout:
👎, turning data into artifacts is a general operation (regardless of semantic type or what plugins you have installed) so makes sense in the framework's SDK. A developer should be able to import data into artifacts without having the domain-specific `q2_types` plugin installed.

Issue #26, opened 2016-03-10
plugin version, workflow, and website should be added to job provenance


gregcaporaso:
should also include citation(s), workflow code (in case the job code is not read-only), ...

gregcaporaso:
Adding a citations list (list is necessary, as there might be more than one) that can be defined when a plugin is instantiated should be done as part of the next release. This can be ``None`` by default for backward compatibility, but we should get this in now as we really need this to support transparency.

PR #42, opened 2016-04-11
TST: add basic framework unit tests

### LINKED ISSUE #41: add framework unit tests
@gregcaporaso @ebolyen and I are extreme programming this today.

jairideout:
Fixes #41. Pair-programmed with @gregcaporaso.

### DIFF
--- a/qiime/sdk/artifact.py
+++ b/qiime/sdk/artifact.py
@@ -17,6 +17,7 @@ import time
 import datetime
 
 import qiime.plugin
+import qiime.sdk
 
 
 class Artifact:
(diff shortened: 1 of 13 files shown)


gregcaporaso:
Good to go once the tests pass.

qiime/sdk/artifact.py
```diff
@@ -185,8 +202,16 @@ def _parse_uuid(self, string):
         return uuid.UUID(hex=string)
 
     # TODO implement provenance parsing
-    def _parse_provenance(self, string):
-        return string
+    def _parse_provenance(self, provenance):
+        if isinstance(provenance, str):
```
wasade: `provenance` can be expressed as `json`, right? if so, would it make sense for this branch to attempt a decode?

(i realize this is merged, asking in the event this is worth opening an issue to discuss)

jairideout: This is handling the case where there's no provenance in the artifact. Currently if there is no provenance, the serialized artifact provenance is the string `'No provenance available because this artifact was generated independently of QIIME.'`. We're still working out provenance and importing data into an artifact so this is just a placeholder for now (#26, #31).

wasade: Cool cool, thanks!

PR #50, opened 2016-05-20
ENH: adding view type annotation

gregcaporaso:
Note that this just adds the annotations - we're not doing anything with the view types yet.

### DIFF
--- a/qiime/sdk/workflow.py
+++ b/qiime/sdk/workflow.py
@@ -37,6 +37,7 @@ class Signature:
         outputs : collections.OrderedDict
             Named output to semantic type.
 
+
         """
         self.name = name
         self.input_artifacts = {}
@@ -77,15 +78,22 @@ class Workflow:
             metadata, template = frontmatter.parse(fh.read())
 
         type_imports = metadata['type-imports']
+
         input_types = {}
+        input_views = {}
         for name, type_expr in metadata['inputs'].items():
-            input_types[name] = cls._parse_type(type_imports, type_expr)
+            semantic_type, view_type = cls._parse_type(type_imports, type_expr)
+            input_types[name] = semantic_type
+            input_views[name] = view_type
 
         output_types = collections.OrderedDict()
+        output_views = collections.OrderedDict()
         for output in metadata['outputs']:
             # TODO validate each nested dict has exactly one item
             name, type_expr = list(output.items())[0]
-            output_types[name] = cls._parse_type(type_imports, type_expr)
+            semantic_type, view_type = cls._parse_type(type_imports, type_expr)
+            output_types[name] = semantic_type
+            output_views[name] = view_type
 
         name = metadata['name']
         signature = Signature(name, input_types, output_types)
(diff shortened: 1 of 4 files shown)


qiime/sdk/workflow.py
```diff
@@ -155,6 +174,16 @@ def from_function(cls, function, inputs, outputs, name, doc):
             doc=doc, import_path=import_path, function_name=function_name,
             parameters=parameters, results=results)
         signature = Signature(name, inputs, outputs)
+
+        # just to show that we've got the views
+        input_views = {input: function.__annotations__[input]
+                       for input in inputs.keys()}
+        output_view_types = qiime.core.type.util.tuplize(
+                           function.__annotations__['return'])
+        output_views = dict(zip(outputs.keys(), output_view_types))
```
wasade: i dont think it is necessary to use `keys` here and on line 180

qiime/sdk/workflow.py
```diff
@@ -155,6 +174,16 @@ def from_function(cls, function, inputs, outputs, name, doc):
             doc=doc, import_path=import_path, function_name=function_name,
             parameters=parameters, results=results)
         signature = Signature(name, inputs, outputs)
+
+        # just to show that we've got the views
+        input_views = {input: function.__annotations__[input]
+                       for input in inputs.keys()}
+        output_view_types = qiime.core.type.util.tuplize(
```
wasade: that's going a bit deep. would you be supportive of the creation of an issue oriented around import aliases?

wasade:
Function annotations will open all new exciting doors for obfuscating code. Absolutely in support of using them, but just as an example of some of the interesting things that can be done:

```python
09:52:26 (daniel@sandbar):diversity> a = np.arange(5)

09:52:32 (daniel@sandbar):diversity> def foo() -> a:
    return "asd"
   ....:

09:52:37 (daniel@sandbar):diversity> foo()
Out[13]: 'asd'

09:52:39 (daniel@sandbar):diversity> foo.__annotations__
Out[14]: {'return': array([0, 1, 2, 3, 4])}

09:52:42 (daniel@sandbar):diversity> def bar() -> foo:
   ....:     return 123
   ....:

09:54:02 (daniel@sandbar):diversity> bar()
Out[16]: 123

09:54:03 (daniel@sandbar):diversity> bar.__annotations__
Out[17]: {'return': <function __main__.foo>}
```

It's a mix of awesome and scary.

ebolyen:
@wasade Ideally, the annotations will be MyPy compatible, the things annotated will be a "view", so we could probably check to see if there even exists a registered view, throwing an error otherwise. (These are independent from the semantic types)

qiime/sdk/workflow.py
ebolyen: I wouldn't be opposed :+1:

qiime/sdk/workflow.py
gregcaporaso: fixed, thanks!

PR #58, opened 2016-06-20
ENH: SubprocessExecutor.__call__ returns Job

### LINKED ISSUE #55: Job object is not accessible through the use of SubprocessExecutor
`.sdk.execution.SubprocessExecutor` creates a `.sdk.job.Job`, and writes the code to a file to be ran, but doesn't allow access to the object information itself. This was noticed while trying to use the `Job.uuid` as the identifier for tracking the current processes being ran by qiime-studio, and realizing that only the `Future` object was being returned.

Found while working on [PR#32 in qiime-studio](/qiime2/qiime-studio/pull/32#discussion-diff-67222461)

jairideout:
Fixes #55.


ebolyen:
:+1:

PR #62, opened 2016-06-27
ENH: added parsing and basic error handling

gregcaporaso:


qiime/metadata.py
```diff
@@ -32,5 +52,9 @@ def __init__(self, series):
     def __repr__(self):
         return self._series.to_json()
 
+    @staticmethod
+    def load(path, category):
```
gregcaporaso: Let me know if you prefer a different method name here.

gregcaporaso:
@jairideout, @ebolyen this one is ready.

qiime/metadata.py
```diff
@@ -14,12 +16,30 @@ def __init__(self, dataframe):
     def __repr__(self):
         return self._dataframe.to_json()
 
+    @staticmethod
+    def load(path):
+        try:
+            df = pd.read_csv(path, sep='\t', index_col=0, dtype=object)
```
wasade: @eldeveloper, @josenavas @antgonza, wanted to draw attention here in the event any of the pandas + qiita experiences are relevant as I know there were a few unexpected issues encountered.

antgonza: Thanks @wasade. 

I think the only concerns, which might not be are:
- values with leading/following empty spaces
- blank rows or columns
- if the first row is gonna be used like the index that means that it can have any label and this might completely confuse users. However, not hard feelings about this one.
- in past versions of pandas if you index on read, it will try to auto cast the values of the index (even ignoring the dtype parameter), not sure if this is still the case with the version that skbio uses

wasade: Thanks, @antgonza. The last point is particularly annoying with AG samples as most are numeric-like (e.g., 10317.000004210) but are not safe to treat as floats.

qiime/metadata.py
```diff
@@ -14,12 +16,30 @@ def __init__(self, dataframe):
     def __repr__(self):
         return self._dataframe.to_json()
 
+    @staticmethod
```
jairideout: use classmethod so that subclasses work

qiime/metadata.py
```diff
@@ -14,12 +16,30 @@ def __init__(self, dataframe):
...
+    def load(path):
+        try:
+            df = pd.read_csv(path, sep='\t', index_col=0, dtype=object)
+        except pd.io.common.CParserError:
+            raise ValueError(
+                'Metadata file format is invalid for file %s. Currently only '
+                'QIIME 1 sample metadata mapping files are officially '
+                'supported. These can be validated using [Keemei]'
```
jairideout: Markdown link in an error message is odd, how about `...validated using Keemei: http://keemei.qiime.org`?

qiime/metadata.py
```diff
@@ -14,12 +16,30 @@ def __init__(self, dataframe):
     def __repr__(self):
         return self._dataframe.to_json()
 
+    @staticmethod
+    def load(path):
+        try:
+            df = pd.read_csv(path, sep='\t', index_col=0, dtype=object)
+        except pd.io.common.CParserError:
```
jairideout: Does this error get raised if `path` doesn't exist or has the wrong permissions?

qiime/metadata.py
```diff
@@ -6,6 +6,8 @@
 # The full license is in the file COPYING.txt, distributed with this software.
 # ----------------------------------------------------------------------------
 
+import pandas as pd
```
jairideout: Now that pandas is a dep, add to setup.py and .travis.yml

jairideout:
@gregcaporaso done reviewing

qiime/metadata.py
gregcaporaso: Confirmed that that's an issue, and fixed (note that it's not a bug in pandas though, ``dtype`` here is specifically for the columns, not the index).

qiime/metadata.py
```diff
@@ -14,12 +16,33 @@ def __init__(self, dataframe):
     def __repr__(self):
         return self._dataframe.to_json()
 
+    @staticmethod
+    def load(path):
+        try:
+            df = pd.read_csv(path, sep='\t', dtype=object).\
```
antgonza: Thanks. Just want to confirm that we are fine allowing empty columns/rows, which are a problem when working in Excel and I'm not sure if Google spreadsheets also sends these.

gregcaporaso: I'm good with it for now, unless there are specific issues. This is just a first pass for alpha. We need to have a discussion about this file format in the (post-SciPy) future.

gregcaporaso:
I think all comments have now been addressed. My battery is about to die, and I'm going to be tied up in a meeting for most of the day today, so if there are other issues and this needs to go in, feel free to make changes and push from another branch.

jairideout:
@gregcaporaso the new pandas dependency is missing from setup.py and .travis.yml. I'll merge this since master is currently broken, and include the pandas dep in my PR to fix master and add unit tests.

gregcaporaso:
Ah, sorry, think I must have forgotten to git add before my git commit
--amend

Sent from my phone, please excuse typos and brevity.
On Jun 28, 2016 9:46 AM, "Jai Ram Rideout" <notifications@github.com> wrote:

Merged #62 <https://github.com/qiime2/qiime2/pull/62>.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/qiime2/qiime2/pull/62#event-706618240>, or mute the
thread
<https://github.com/notifications/unsubscribe/AALvdNC83Fio12xhGYsPqkdym_OADOw4ks5qQU_lgaJpZM4I_m5p>
.

jairideout:
No problem

PR #65, opened 2016-06-29
ENH: allowed definition of Properties in plugins

gregcaporaso:
@ebolyen, can you review this one? I tested that this works with a new type that defines properties via the q2_types unit tests.


ebolyen:
:-1:!

PR #67, opened 2016-06-29
BUG: support command-line zipped archives and nested data directories

jairideout:
This fixes the two archiver bugs @ebolyen, @jakereps, and @ElDeveloper found. Not adding tests in this PR as I'm refactoring all tests in a separate branch, and the current tests are broken. I'll include test cases for these bugs in that PR.


ebolyen:
:+1: nice!

PR #68, opened 2016-06-29
ENH: adds parse_type and singleton PluginManager

ebolyen:
Pair programmed with @jairideout and @jakereps

### DIFF
--- a/qiime/core/util.py
+++ b/qiime/core/util.py
@@ -24,27 +24,65 @@ def overrides(cls):
     return decorator
 
 
-def parse_type(string):
+def parse_type(string, expect=None):
+    """Convert a string into a TypeExpression
+
+    Parameters
+    ----------
+    string : str
+        The string type expression to convert into a TypeExpression
+    expect : {'semantic', 'primitive', 'visualization'}, optional
+        Will raise a TypeError if the resulting TypeExpression is not a member
+        of `expect`.
+
+    Returns
+    -------
+    TypeExpression
+
+    Raises
+    ------
+    ValueError
+        Raised when `expect` has an invalid value
+    TypeError
+        Raised when the expression contains invalid characters
+    TypeError
+        Raised when the expression does not result in a member of `expect`
+    UnknownTypeError
+        Raised when unkown types are present in the expression.
+
+    """
     # Avoid circular imports
     import qiime.sdk
-    from .type import Visualization, Properties
+    import qiime.core.type as qtype
 
-    if len(string.split('\n')) != 1:
-        raise TypeError("Found multiple lines in type expression %r. Will not "
-                        "evaluate to avoid arbitrary code execution."
+    if expect is not None and expect not in {'semantic', 'primitive',
+                                             'visualization'}:
+        raise ValueError("`expect` got %r, must be 'semantic', 'primitive',"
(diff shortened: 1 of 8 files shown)


qiime/core/util.py
```diff
@@ -20,3 +22,36 @@ def decorator(func):
...
         return func
     return decorator
+
+
+def parse_type(string):
+    # Avoid circular imports
+    import qiime.sdk
+    from .type import Visualization, Properties
+
+    if len(string.split('\n')) != 1:
```
jairideout: Come to think of it, we're disallowing even a single `\n`. Is that desirable? If so, we should probably update the error message and just check `'\n' in string` which'll be faster.

qiime/core/util.py
```diff
@@ -20,3 +22,36 @@ def decorator(func):
...
+    locals_ = {k: v[1] for k, v in pm.semantic_types.items()}
+    locals_[Visualization.name] = Visualization
+    locals_["Properties"] = Properties
+
+    try:
+        return eval(string, {'__builtins__': {}}, locals_)
+    except NameError as e:
+        # http://stackoverflow.com/a/2270822/579416
+        name, = re.findall("name '(\w+)' is not defined", str(e))
+        raise UnknownTypeError("Name %r is not a defined QIIME type, a plugin"
```
jairideout: 👍

qiime/sdk/method.py
```diff
@@ -157,32 +157,13 @@ def _from_markdown(self, markdown_filepath):
...
-        if ';' in semantic_type_exp:
-            raise TypeError("Invalid type expression in artifact. Will not"
-                            " load to avoid arbitrary code execution.")
-
-        pm = qiime.sdk.PluginManager()
-        locals_ = {k: v[1] for k, v in pm.semantic_types.items()}
-        # Set up all of the types we know about in local scope of the eval
-        # so that complicated type expressions are evaluated.
-        type_ = eval(semantic_type_exp, {'__builtins__': {}}, locals_)
+        type_ = qiime.sdk.parse_type(semantic_type_exp)
```
jairideout: The previous code implicitly ensured the parsed type was a semantic type. Can you add a check here that `type_` meets that requirement?

jairideout: It would also be nice to remove most of the logic in `_parse_primitive_type` so that it just calls `parse_type` and verifies the type is a primitive. `_parse_primitive_type` is currently hardcoded to a few primitives and will break soon.

Thinking about this API more, perhaps it makes sense to have `parse_type` accept an optional parameter that specifies the allowed type expressions? e.g. a parameter that indicates primitive vs semantic vs visualization, and by default allows all defined types.

qiime/sdk/plugin_manager.py
```diff
@@ -11,7 +11,16 @@
 
 
 class PluginManager:
```
jairideout: Can you comment that this class is a singleton, and why it is a singleton?

jairideout:
@ebolyen this is really useful! Had a few comments/questions

qiime/sdk/plugin_manager.py
ebolyen: done

jairideout:
Beautiful!!!

PR #69, opened 2016-06-30
ENH: initial artifact API

ebolyen:
~~Not even remotely ready for merge. This was just a quick hack that I wanted to share.~~

Questions:
- ~~What should the importable name be?~~
- What happens if a plugin imports another plugin's generated API?
- ~~Should visualizations and methods be separated?~~
- ~~Should plugins be able to define a hierarchical package structure?~~

Example:
```ipynb
In [1]: from qiime.plugin.feature_table import 
presence_absence    relative_frequency  
rarefy              summarize           

In [1]: from qiime.plugin.feature_table import rarefy

In [2]: rarefy?
Type:           Method
String form:    <qiime.sdk.method.Method object at 0x7fb2c688cef0>
File:           ~/qiime2/qiime2/qiime/sdk/method.py
Docstring:      <no docstring>
Call signature: rarefy(table:biom.table.Table, depth:int) -> biom.table.Table
```

### DIFF
--- a/qiime/plugin.py
+++ b/qiime/plugin.py
@@ -7,6 +7,8 @@
 # ----------------------------------------------------------------------------
 
 import pkg_resources
+import sys
+import importlib.machinery
 
 import qiime.sdk
 import qiime.core.data_layout
@@ -18,6 +20,7 @@ from qiime.core.type import (SemanticType, is_semantic_type, Int, Str, Float,
 __all__ = ['Plugin', 'Int', 'Str', 'Float', 'Color', 'Metadata',
            'MetadataCategory', 'SemanticType', 'Properties', 'Range',
            'Choices', 'Arguments']
+__path__ = []
 
 
 class Plugin:
@@ -27,8 +30,8 @@ class Plugin:
         self.website = website
         self.package = package
 
-        self.methods = PluginMethods(self.package)
-        self.visualizers = PluginVisualizers()
+        self.methods = PluginMethods(self.name, self.package)
+        self.visualizers = PluginVisualizers(self.name)
 
         self.data_layouts = {}
         self.types = {}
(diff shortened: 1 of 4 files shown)


jairideout:
🚀 🌞 

> What should the importable name be?

I think what you have is good (`Method.id`) since we have a TODO to make `.id` a Python-friendly identifier. I like the import path you have too, though perhaps we should differentiate between "official" plugins that we know will be stable vs. third-party, which may not be stable.

> What happens if a plugin imports another plugin's generated API?

Ideally this would "just work" but maybe it's okay to have plugin developers import the "view" API directly from the plugins they rely on. I think that case will be much more common than a plugin relying on the "artifact" API of another plugin (at least until we support pipelines).

> Should visualizations and methods be separated?

I think it's a good idea to separate them because they have different APIs and intended uses. It'll be consistent with how they are separated at plugin registration.

> Should plugins be able to define a hierarchical package structure?

A future enhancement if it becomes necessary. For now it's simpler to recommend that plugins stay fairly limited in scope (pick one domain-specific topic and be good at that, ala Unix).

ebolyen:
> I think it's a good idea to separate them because they have different APIs and intended uses. It'll be consistent with how they are separated at plugin registration.

Would this imply that it would work something like:
```python
from qiime.plugin.feature_table.methods import rarefy
```
and that decision would impact any implementation of a hierarchical package structure.

While I agree in principle that separating them is good idea, I also think that might be a little obtuse to deal with. Also, do we *really* want plugin developers reusing names for visualizers and methods? That might confuse users (myself in particular) as they would need to keep track of "which" `foobar` they were dealing with at that moment (and we should hope that `Method::foobar` and `Visualizer::foobar` have the same signature!).

jairideout:
>  Also, do we really want plugin developers reusing names for visualizers and methods?

Probably not, we can guard against that at registration time to prevent confusion.

> and we should hope that Method::foobar and Visualizer::foobar have the same signature!).

They won't because one returns a `Visualization` while the other returns one or more `Artifacts`. This is solved by disallowing a method and visualization with the same name.

I still think it's a good idea to separate them, as we're already planning to do that in other interfaces. I've been a little frustrated that q2cli doesn't group methods and visualizations, you have to call `--help` on a method/viz name to see what it is. Since this is the lowest-level interface we provide, it seems like a good idea to have that distinction here too. The return types are fundamentally different, with different APIs and semantics (and even file extension, if you're saving).

ebolyen:
Updated example:
```pytb
In [1]: from qiime.plugin import dummy_plugin

In [2]: dummy_plugin
Out[2]: <module 'qiime.plugin.dummy_plugin' (generated QIIME API)>

In [3]: dummy_plugin.
dummy_plugin.methods      dummy_plugin.visualizers  

In [3]: dummy_plugin.methods
Out[3]: <module 'qiime.plugin.dummy_plugin.methods' (generated QIIME API)>

In [4]: dummy_plugin.methods.
dummy_plugin.methods.concatenate_ints           dummy_plugin.methods.split_ints
dummy_plugin.methods.concatenate_ints_markdown  dummy_plugin.methods.split_ints_markdown
dummy_plugin.methods.merge_mappings             

In [4]: dummy_plugin.methods.concatenate_ints
Out[4]: <method qiime.plugin.dummy_plugin.methods.concatenate_ints>

In [5]: dummy_plugin.methods.concatenate_ints.async
Out[5]: <function qiime.plugin.dummy_plugin.methods.concatenate_ints.async>

In [6]: dummy_plugin.methods.concatenate_ints.__call__
Out[6]: <function qiime.plugin.dummy_plugin.methods.concatenate_ints.__call__>

In [7]: dummy_plugin.visualizers.
dummy_plugin.visualizers.mapping_viz      dummy_plugin.visualizers.most_common_viz  

In [7]: dummy_plugin.visualizers.mapping_viz
Out[7]: <visualizer qiime.plugin.dummy_plugin.visualizers.mapping_viz>

In [8]: dummy_plugin.visualizers.mapping_viz.__call__
Out[8]: <function qiime.plugin.dummy_plugin.visualizers.mapping_viz.__call__>

In [9]: dummy_plugin.visualizers.mapping_viz.async
Out[9]: <function qiime.plugin.dummy_plugin.visualizers.mapping_viz.async>
```

@jairideout, @gregcaporaso does this look good to you guys? I can work on some tests if so.

gregcaporaso:
This looks pretty reasonable to me. Just confirming: we wouldn't normally see a difference between markdown methods/visualizers (e.g., ``dummy_plugin.methods.concatenate_ints`` and ``dummy_plugin.methods.concatenate_ints_markdown``), right? That's only showing up here because these are actually defined as different methods that do the same thing for testing purposes?

ebolyen:
@gregcaporaso yeah those should be handled identically so they are both 
```<method qiime.plugin.dummy_plugin.methods.[name here]>``` as far as this API is concerned.

**Comment by ebolyen** [comment]
To clarify, the `_markdown` suffix is just an artifact of this particular testing plugin.

**Comment by jairideout** [comment]
@ebolyen 👍 👍 👍 👍 👍 👍 👍 👍 👍 !!! Let me know when you're ready with tests and I'll review.

**Comment by jairideout** [comment]
@gregcaporaso `concatenate_ints` and `concatenate_ints_markdown` are two different methods that are functionally identical, one is defined as a Python function and the other in a Markdown file. I named one of the methods to indicate it came from a Markdown file because method IDs have to be unique. Sorry for the confusion on that! `split_ints` and `split_ints_markdown` follow the same reasoning.

**Comment by ebolyen** [comment]
@jairideout added tests, ready for review.

File: qiime/plugin.py [hunk 1]
```diff
@@ -77,26 +80,101 @@ def register_type_to_data_layout(self, semantic_type, name, version):
 
 # TODO refactor these classes, they are basically the same
 class PluginMethods(dict):
-    def __init__(self, package):
+    def __init__(self, plugin, package):
```
jairideout: Can you call this `plugin_name`, or pass the actual `Plugin` object? The current name is confusing

qiime/plugin.py
```diff
@@ -77,26 +80,101 @@ def register_type_to_data_layout(self, semantic_type, name, version):
...
         self._package = package
+        self._plugin = plugin
         super().__init__()
 
     def register_function(self, function, inputs, parameters, outputs, name,
                           description):
         method = qiime.sdk.Method.from_function(function, inputs, parameters,
-                                                outputs, name, description)
+                                                outputs, name, description,
+                                                plugin=self._plugin)
```
jairideout: Same comment for `plugin` parameter name

qiime/sdk/method.py
```diff
@@ -45,7 +45,7 @@ def __new__(cls):
 
     @classmethod
     def from_function(cls, function, inputs, parameters, outputs, name,
-                      description):
+                      description, plugin=None):
```
ebolyen: I'll update these and others to `plugin_name`

qiime/tests/test_artifact_api.py
```diff
@@ -0,0 +1,121 @@
...
+import sys
+import types
+import importlib
+
+import qiime.sdk
+
+
+class TestImports(unittest.TestCase):
+    def setUp(self):
+        self._sys_modules = sys.modules.copy()
```
jairideout: Can you call `get_dummy_plugin` to ensure QIIMETEST has been set?

ebolyen:
:tada: :balloon: :balloon: :tada: :fireworks: :100:

jairideout:
The future is here, courtesy of @ebolyen!!! 🚀 🚀 🚀 🌞

gregcaporaso:
image

PR #72, opened 2016-07-04
TST: add qiime.core.testing package with dummy plugin, add unit tests to framework

### LINKED ISSUE #43: unit test expansion
As part of #42, basic unit tests were added for pieces of the core framework. Expand on this to cover boundary cases and errors.

### LINKED ISSUE #44: centralize functionality that is reused across many tests
Specifically ``DummyType``, dummy workflows, etc. This is important for testing the basic functionality, as the actual types and workflows live in other repositories.

jairideout:
Adds a "dummy-plugin" to the framework for testing. The plugin includes several types, data layouts, methods, and visualizers. All plugin components are defined in the `qiime.core.testing` package and are imported by the unit tests.

Unit tests will now fail in a predictable way with an informative error message if the QIIMETEST environment variable isn't set.

Minor changes to archive format provenance (and `Provenance` object):

- method-reference -> executor-reference. The new name is more general as the reference can now be a Method or Visualizer.

- parameters -> parameter-references

`DataLayout` now lives under `qiime.core.data_layout` since it is a private class.

Fixes #43, fixes #44.


qiime/core/testing/util.py
```diff
@@ -0,0 +1,61 @@
...
+        raise RuntimeError(
+            "When running QIIME 2 unit tests, the QIIMETEST environment "
+            "variable must be defined so that plugins required by unit tests "
+            "are loaded. The value of the QIIMETEST environment variable can "
+            "be anything. Example command: QIIMETEST= nosetests")
+    return plugin_manager.plugins['dummy-plugin']
+
+
+# Taken from scikit-bio, see licenses directory for scikit-bio license.
+def get_data_path(fn, subfolder='data'):
```
ebolyen: Should we be using something like [this](http://setuptools.readthedocs.io/en/latest/pkg_resources.html#resourcemanager-api). It sounds like this is the "correct" way to actually do this. But I've never tried it before.

qiime/core/testing/data_layout.py
```diff
@@ -0,0 +1,131 @@
...
+def int_sequence_to_list(data_dir):
+    with open(os.path.join(data_dir, 'ints.txt'), 'r') as fh:
+        view = []
+        for line in fh:
+            view.append(int(line.rstrip('\n')))
+        return view
+
+
+def list_to_int_sequence(view, data_dir):
+    with open(os.path.join(data_dir, 'ints.txt'), 'w') as fh:
```
ebolyen: Not important, but food for thought. We should probably define an explicit `newline` type for our standard formats when writing, instead of relying on the user's system. Otherwise the data-layout is technically ambiguous in a pretty fundamental way.

qiime/core/testing/plugin.py
```diff
@@ -0,0 +1,163 @@
...
+    outputs=[
+        ('concatenated_ints', IntSequence1)
+    ],
+    name='Concatenate integers',
+    description='This method concatenates integers into a single sequence in '
+                'the order they are provided.'
+)
+
+# TODO update to use TypeMap so IntSequence1 | IntSequence2 are accepted, and
+# the return type is IntSequence1 or IntSequence2.
```
ebolyen: :+1: :+1: :+1:

qiime/core/testing/type.py
```diff
@@ -0,0 +1,15 @@
...
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import qiime.plugin
+
+
+IntSequence1 = qiime.plugin.SemanticType('IntSequence1')
+IntSequence2 = qiime.plugin.SemanticType('IntSequence2')
+Mapping = qiime.plugin.SemanticType('Mapping')
+FourInts = qiime.plugin.SemanticType('FourInts')
```
ebolyen: Is it worth having composite types here? It would be relevant for the sake of integration testing, but we could also just rely on the unit-tests of qiime.core.type

qiime/core/testing/util.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def get_dummy_plugin():
+    plugin_manager = qiime.sdk.PluginManager()
+    if 'dummy-plugin' not in plugin_manager.plugins:
+        raise RuntimeError(
+            "When running QIIME 2 unit tests, the QIIMETEST environment "
+            "variable must be defined so that plugins required by unit tests "
+            "are loaded. The value of the QIIMETEST environment variable can "
+            "be anything. Example command: QIIMETEST= nosetests")
```
ebolyen: Should we attach a value after `=`? Some users may not realize the whitespace is significant if they are typing it out.

qiime/plugin.py
```diff
@@ -9,51 +9,17 @@
...
                              Color, Metadata, MetadataCategory, Properties,
                              Range, Choices, Arguments)
 
-__all__ = ['load', 'Plugin', 'Int', 'Str', 'Float', 'Color', 'Metadata',
+__all__ = ['Plugin', 'Int', 'Str', 'Float', 'Color', 'Metadata',
            'MetadataCategory', 'SemanticType', 'Properties', 'Range',
            'Choices', 'Arguments']
 
 
-def load(plugin_name, plugin_entry_point_name=None):
```
ebolyen: I forgot this even existed!

qiime/sdk/plugin_manager.py
```diff
@@ -28,11 +28,11 @@ def _init(self):
...
-        plugins = [e.load() for e in pkg_resources.iter_entry_points(
-            group='qiime.plugin')]
-        if 'QIIMETEST' in os.environ:
-            import qiime.core.testing
-            plugins.append(qiime.core.testing.plugin)
+        plugins = []
+        for entry_point in pkg_resources.iter_entry_points(
+                group='qiime.plugin'):
+            if entry_point.name != 'dummy-plugin' or 'QIIMETEST' in os.environ:
+                plugins.append(entry_point.load())
```
ebolyen: :+1:

qiime/sdk/tests/test_method.py
```diff
@@ -0,0 +1,379 @@
...
+                    ('int2', inspect.Parameter('int2', kind))
+                ]
+                self.assertEqual(parameters, exp_parameters)
+
+                self.assertEqual(signature.return_annotation,
+                                 inspect.Signature.empty)
+
+    def test_callable_and_async_different_signature(self):
+        # Test that a different Method object has a different dynamic
+        # signature.
```
ebolyen: Can you provide the test dependency `test_callable_and_async_signature` in this comment and vice-versa? I was super confused for a few minutes how it was being tested. This test only works in the context of the other existing. Alternatively (to avoid any potential trickery by `setUp`) we could just test two different methods here (a little redundant, but then the test has no external dependencies for its claim)

qiime/core/testing/util.py
jairideout: I wanted to give the shortest possible command to avoid extra typing for devs, but you're right that this could cause confusion to end users wanting to run the tests. How about `QIIMETEST=on nosetests`?

ebolyen: `1` would be shorter, is generally understood as "on"/"true", and fits the general paradigm of UNIX tools.

qiime/sdk/tests/test_method.py
```diff
@@ -0,0 +1,379 @@
...
+                    provenance.executor_reference.startswith(method.id))
+                self.assertEqual(provenance.artifact_uuids, {
+                    'ints': artifact.uuid
+                })
+                self.assertEqual(provenance.parameter_references, {})
+
+                self.assertIsInstance(output_artifact.uuid, uuid.UUID)
+
+            # Output artifacts have the same provenance.
+            self.assertEqual(result[0].provenance, result[1].provenance)
```
ebolyen: :+1: Nice, this should be pretty robust regardless of implementation!

qiime/sdk/tests/test_visualization.py
```diff
@@ -0,0 +1,244 @@
...
+import qiime.core.type
+from qiime.sdk import Visualization, Provenance
+
+from qiime.core.testing.visualizer import mapping_viz
+from qiime.core.testing.util import get_data_path
+
+
+class TestVisualization(unittest.TestCase):
+    def setUp(self):
+        # TODO standardize temporary directories created by QIIME
```
ebolyen: :+1: it might even be worth creating a qiime directory in the associated tempdir to keep things tidy.

qiime/core/testing/type.py
jairideout: There isn't a whole lot of type-system integration tests yet (e.g. exercising `TypeMap`). Okay to add more type-related tests in the future? There's many many many more test cases to be added but I want to get this merged ASAP so we have a reasonable testbase to work from (and have master back to passing again).

qiime/sdk/visualizer.py
```diff
@@ -72,10 +72,11 @@ def from_function(cls, function, inputs, parameters, name, description):
             raise TypeError(
                 "Visualization function must have at least one argument")
 
-        if function.__annotations__['return'] is not None:
+        if ('return' in function.__annotations__ and
+                function.__annotations__['return'] is not None):
             raise TypeError(
-                "Visualization function cannot return anything. Write "
-                "visualization output to `output_dir`.")
+                "Visualization function %r cannot return anything. Write "
```
ebolyen: `Visualization` -> `Visualizer`?

qiime/sdk/tests/test_method.py
jairideout: I'll just rename both tests to indicate the type of signature they're exercising (sig with artifacts and parameters vs. a sig without parameters).

qiime/sdk/visualizer.py
jairideout: Good catch! I'll update the other error messages in this method too

ebolyen:
This is awesome @jairideout! Done reviewing, let me know when you are ready for merge!

qiime/core/testing/data_layout.py
jairideout: That's a good idea, right now the assumption is that readers and writers open files in universal-newline mode so `\n` can safely be used, but that assumption could easily be violated and things might not stay cross-platform compatible. Are there specific updates you want to see in this code now, or is this more of a general problem to solve post-alpha?

jairideout: Note that the current formats (here and in q2-types) should be cross-platform compatible b/c universal newline mode is enabled by default, so `\n` can be safely used when reading/writing.

ebolyen: Just a general problem

qiime/core/testing/util.py
jairideout: done, using `pkg_resources.resource_filename` totally works

jairideout:
Thanks @ebolyen!!!

ebolyen:
Merged https://github.com/qiime2/q2-types/pull/11

PR #76, opened 2016-07-05
adds ResultBase.extract and Visualization.get_indices

gregcaporaso:
This is ready for review. 

Here's how I envision using this:

```python
In [1]: from qiime.sdk import Visualization

In [2]: Visualization.get_index_paths('pd-compare.qzv')
Out[2]: {'html': 'pd-compare/data/index.html'}

In [3]: Visualization.extract('pd-compare.qzv', '.')
Out[3]: '.'

In [5]: index_paths = Visualization.get_index_paths('pd-compare.qzv')

In [6]: output_dir = Visualization.extract('pd-compare.qzv', '.')

In [9]: import os

In [12]: html_path = os.path.join(output_dir, index_paths['html'])

In [13]: os.path.exists(html_path)
Out[13]: True
```

### DIFF
--- a/qiime/core/archiver.py
+++ b/qiime/core/archiver.py
@@ -29,6 +29,12 @@ class Archiver:
     _data_dirname = 'data'
 
     @classmethod
+    def extract(cls, filepath, output_dir):
+        with zipfile.ZipFile(filepath, mode='r') as zf:
+            zf.extractall(output_dir)
+        return output_dir
+
+    @classmethod
     def load(cls, filepath):
         if not zipfile.is_zipfile(filepath):
             raise zipfile.BadZipFile(
@@ -211,6 +217,30 @@ class Archiver:
                                    self.provenance.artifact_uuids.items()},
                 'parameter-references': self.provenance.parameter_references
             }
+
+    def get_index_paths(self):
+        # TODO: this extracts the entire archive, even though that isn't
+        # necessary to get the index paths. Refactor to avoid this.
+        if not os.path.exists(self._data_dir):
+            self._extract_data()
+        result = {}
+        for path in os.listdir(self._data_dir):
+            if os.path.isfile(os.path.join(self._data_dir, path)):
+                filename = os.path.split(path)[1]
+                if filename.startswith('index.'):
+                    ext = os.path.splitext(filename)[1][1:]
+                    if ext in result:
+                        # TODO: this should [additionally] be handled
+                        # elsewhere, probably on population of self._data_dir.
+                        raise ValueError(
+                            "Multiple index files identified with %s "
+                            "extension (%s, %s). This is currently "
+                            "unsupported." %
+                            (ext, result[ext],
+                             os.path.join(self._data_dirname, path)))
+                    else:
+                        result[ext] = os.path.join(self._data_dirname, path)
+        return result
 
 
 _README_TEXT = """# QIIME 2 archive
--- a/qiime/core/result_base.py
+++ b/qiime/core/result_base.py
@@ -7,8 +7,6 @@
 # ----------------------------------------------------------------------------
 
 import abc
-import zipfile
-import tempfile
 
 import qiime.core.archiver
 
@@ -29,12 +27,8 @@ class ResultBase(metaclass=abc.ABCMeta):
         return result
 
     @classmethod
-    def extract(cls, filepath, output_dir=None):
-        if output_dir is None:
-            output_dir = tempfile.gettempdir()
-        with zipfile.ZipFile(filepath, mode='r') as zf:
-            zf.extractall(output_dir)
-        return output_dir
+    def extract(cls, filepath, output_dir):
+        return qiime.core.archiver.Archiver.extract(filepath, output_dir)
 
     @property
     def type(self):
--- a/qiime/core/result_base.py
+++ b/qiime/core/result_base.py
@@ -7,6 +7,8 @@
 # ----------------------------------------------------------------------------
 
 import abc
+import zipfile
+import tempfile
 
 import qiime.core.archiver
 
@@ -25,6 +27,14 @@ class ResultBase(metaclass=abc.ABCMeta):
         result._archiver = qiime.core.archiver.Archiver.load(filepath)
         result._assert_valid_type(result.type)
         return result
+
+    @classmethod
+    def extract(cls, filepath, output_dir=None):
+        if output_dir is None:
+            output_dir = tempfile.gettempdir()
+        with zipfile.ZipFile(filepath, mode='r') as zf:
+            zf.extractall(output_dir)
+        return output_dir
 
     @property
     def type(self):
--- a/qiime/sdk/tests/test_artifact.py
+++ b/qiime/sdk/tests/test_artifact.py
@@ -303,29 +303,9 @@ class TestArtifact(unittest.TestCase):
                                        self.provenance)
         artifact.save(fp)
 
-        Artifact.extract(fp)
-
-        contents = [
-            'artifact/VERSION',
-            'artifact/metadata.yaml',
-            'artifact/README.md',
-            'artifact/data/file1.txt',
-            'artifact/data/file2.txt',
-            'artifact/data/nested/file3.txt',
-            'artifact/data/nested/file4.txt']
-        for fp in contents:
-            expected_fp = os.path.join(tempfile.gettempdir(), fp)
-            self.assertTrue(os.path.exists(expected_fp),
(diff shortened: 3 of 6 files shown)


qiime/sdk/tests/test_artifact.py
```diff
@@ -297,6 +297,49 @@ def test_load_from_externally_created_zipfile(self):
             }
             self.assertEqual(fps, expected)
 
+    def test_extract(self):
+        fp = os.path.join(self.test_dir.name, 'artifact.qza')
+        artifact = Artifact._from_view([-1, 42, 0, 43], FourInts,
+                                       self.provenance)
+        artifact.save(fp)
+
+        Artifact.extract(fp)
```
ebolyen: Can you assert that the result is `fp`?

qiime/sdk/tests/test_artifact.py
```diff
@@ -297,6 +297,49 @@ def test_load_from_externally_created_zipfile(self):
...
+                            'File %s was not extracted.' % fp)
+
+    def test_extract_alt_output_dir(self):
+        fp = os.path.join(self.test_dir.name, 'artifact.qza')
+        artifact = Artifact._from_view([-1, 42, 0, 43], FourInts,
+                                       self.provenance)
+        artifact.save(fp)
+
+        output_dir = os.path.join(self.test_dir.name, 'artifact-extract-test')
+        Artifact.extract(fp, output_dir=output_dir)
```
ebolyen: Same here.

qiime/sdk/visualization.py
```diff
@@ -33,3 +35,7 @@ def _from_data_dir(cls, data_dir, provenance):
         # shutil.copytree doesn't allow the destination directory to exist.
         viz._archiver.save_data(data_dir, distutils.dir_util.copy_tree)
         return viz
+
+    def get_indices(self):
+        indices = glob.glob(os.path.join(self._archiver._data_dir, 'index.*'))
```
ebolyen: `_data_dir` may not exist (Artifacts/Visualizations) are lazy and won't self-extract until needed. Also I'm not a fan of accessing the private functionality of a instance member. The `Archiver` should probably provide some public method for getting the data-directory (allowing it to be extracted/populated if needed).

qiime/sdk/tests/test_visualization.py
```diff
@@ -240,6 +241,63 @@ def test_load_from_externally_created_zipfile(self):
...
+        visualization = Visualization._from_data_dir(self.data_dir,
+                                                     self.provenance)
+        actual = visualization.get_indices()
+        expected = {'data/index.html'}
+        self.assertEqual(actual, expected)
+
+    def test_get_indices_multiple(self):
+        most_common_viz(self.data_dir,
+                        collections.Counter(range(42)))
+        visualization = Visualization._from_data_dir(self.data_dir,
```
ebolyen: This particular path is causing `_data_dir` to exist on the `Archiver`, but I don't think that is generally true. @jairideout can you confirm?

qiime/sdk/visualization.py
```diff
@@ -33,3 +35,7 @@ def _from_data_dir(cls, data_dir, provenance):
         # shutil.copytree doesn't allow the destination directory to exist.
         viz._archiver.save_data(data_dir, distutils.dir_util.copy_tree)
         return viz
+
+    def get_indices(self):
```
gregcaporaso: I'm rethinking this API... Should this be a class method instead, and return tuples of extension and path (e.g., ``("tsv", "data/index.tsv")``)? That's more in line with how I'd use it for the cli. I'm open to suggestions here - I don't really like this as is.

qiime/sdk/tests/test_artifact.py
gregcaporaso: Yep, missed that, thx.

qiime/sdk/visualization.py
ebolyen: That would be a lot more useful I think.

qiime/sdk/visualization.py
```diff
@@ -33,3 +35,14 @@ def _from_data_dir(cls, data_dir, provenance):
         # shutil.copytree doesn't allow the destination directory to exist.
         viz._archiver.save_data(data_dir, distutils.dir_util.copy_tree)
         return viz
+
+    @classmethod
+    def get_index_paths(cls, fp):
```
ebolyen: It isn't necessarily true that `Visualization` will have a zipped form.

jairideout: I think this needs to be an instance method since `Visualization` can be constructed from a zipfile (`load`) or a directory (`_from_data_dir`). I think this method should exist on `Archiver` because it has access to its `self._data_dir`, which contains the index files and whatever else the visualizer produced. Note that `_data_dir` does not always exist, so you'll need to handle that with the following lines:

```python
if not os.path.exists(self._data_dir):
    self._extract_data()
```

ebolyen: :+1: I think the above lines should be exposed as a public method on the `Archiver` like `get_data_dir()` or similar.

qiime/core/result_base.py
```diff
@@ -26,6 +28,14 @@ def load(cls, filepath):
         result._assert_valid_type(result.type)
         return result
 
+    @classmethod
+    def extract(cls, filepath, output_dir=None):
```
jairideout: 👍

I'd prefer this method to exist on `Archiver` since it deals with zipfiles and `Archiver` will be responsible for handling multiple archive formats when they exist (and they may not always be zip-based). Can you move it there and create an `extract` method here that defers?

jairideout: I see now that `extract` is different from `Archiver._extract_data`; one extracts the entire archive while the other only extracts the `data` directory. I thought we'd be able to nix `_extract_data` in favor of this method, which was why I wanted `output_dir` to default to a temporary directory. Since we aren't using `extract` internally, I think `output_dir` should become a required parameter (unless there was another reason to keep the current default behavior).

qiime/core/archiver.py
```diff
@@ -212,6 +218,20 @@ def _formatted_provenance(self):
...
+        # necessary to get the index paths. Refactor to avoid this.
+        if not os.path.exists(self._data_dir):
+            self._extract_data()
+        result = {}
+        for path in os.listdir(self._data_dir):
+            if os.path.isfile(os.path.join(self._data_dir, path)):
+                filename = os.path.split(path)[1]
+                if filename.startswith('index.'):
+                    ext = os.path.splitext(filename)[1]
+                    result[ext[1:]] = os.path.join(self._data_dirname, path)
```
jairideout: There could be multiple files with the same extension, e.g. `foo.bar.png` and `foo.baz.png`. Can you store a `set` of files with a given extension?

jairideout: Probably worth adding a test case for that too

gregcaporaso: That's true, but what would an interface do with that (e.g., if the cli identifies two ``index*html`` files, which does it display)? Maybe we should only treat ``index.extension`` as the index associated with ``extension``? (This code would need to be adapted to do that.)

qiime/sdk/tests/test_visualization.py
```diff
@@ -240,6 +241,70 @@ def test_load_from_externally_created_zipfile(self):
...
+    def test_get_index_paths_single_from_data_dir(self):
+        visualization = Visualization._from_data_dir(self.data_dir,
+                                                     self.provenance)
+
+        actual = visualization.get_index_paths()
+        expected = {'html': 'data/index.html'}
+        self.assertEqual(actual, expected)
+
+    def test_get_index_paths_multiple_load(self):
+        most_common_viz(self.data_dir,
```
jairideout: `self.data_dir` contains the output of `mapping_viz` (see `setUp`), which has two files `index.html` and `css/style.css`. Can you write `most_common_viz` output to a different directory under `self.test_dir` in case `mapping_viz` creates other index files in the future?

qiime/core/archiver.py
jairideout: Chatted with @ebolyen and we're unsure if an interface would use multiple files with the same extension (it could be related to how we handle IPython reprs though). For now, let's have the method raise an error if it detects multiple files with the same extension, then we can figure out whether it needs to be supported in the future. Does that sound good to you?

jairideout:
@gregcaporaso done reviewing

qiime/core/archiver.py
gregcaporaso: That works.

jairideout:
Thanks @gregcaporaso!

PR #77, opened 2016-07-06
ENH/BUG: add Artifact/Visualization.peek for quick inspection of archives

### LINKED ISSUE #24: create a mechanism for inspecting Artifacts
Likely as a method of ``Artifact``, or some way to view properties of it from interfaces. We need to think about what this would be exactly, but we need some way to view provenance, view summaries of the underlying data, etc.

jairideout:
Solves bug in Artifact/Visualization.load where previously the archive filepath was extracted on demand, resulting in buggy behavior if the filepath was deleted or modified after loading but prior to extraction. This problem is not easily solvable in a cross-platform compatible way so a new `.peek()` classmethod is provided to quickly retrieve an archive's metadata (e.g. UUID, type, and provenance) without extracting. `.load()` now extracts immediately.

Fixes #24.

### DIFF
--- a/qiime/core/archiver.py
+++ b/qiime/core/archiver.py
@@ -6,8 +6,10 @@
 # The full license is in the file COPYING.txt, distributed with this software.
 # ----------------------------------------------------------------------------
 
+import collections
 import io
 import os
+import os.path
 import shutil
 import tempfile
 import uuid
@@ -16,6 +18,10 @@ import zipfile
 
 import qiime.sdk
 from .util import parse_type
+
+
+ArchiveMetadata = collections.namedtuple('ArchiveMetadata',
+                                         ['uuid', 'type', 'provenance'])
 
 
 # TODO use utf-8 encoding when reading/writing files
@@ -35,7 +41,7 @@ class Archiver:
         return output_dir
 
     @classmethod
-    def load(cls, filepath):
+    def peek(cls, filepath):
         if not zipfile.is_zipfile(filepath):
             raise zipfile.BadZipFile(
                 "%r is not a readable ZIP file, or the file does not exist" %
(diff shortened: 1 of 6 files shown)


qiime/core/archiver.py
```diff
@@ -130,39 +175,32 @@ def provenance(self):
     def orphan(self, pid):
         self._pid = pid
 
-    def load_data(self, loader):
-        if not os.path.exists(self._data_dir):
-            self._extract_data()
+    def get_index_paths(self):
```
jairideout: No changes were made this method except to move it into a better location and remove the guards checking if the archive had been extracted.

qiime/core/archiver.py
```diff
@@ -96,25 +107,59 @@ def _parse_provenance(cls, provenance):
                 parameter_references=provenance['parameter-references']
             )
 
-    def __init__(self, uuid, type, provenance, archive_filepath=None):
-        self._uuid = uuid
+    def __init__(self, uuid, type, provenance, archive_filepath=None,
+                 data_initializer=None):
```
jairideout: `data_initializer` is part of a refactoring to make `Archiver` easier to reason about. It is no longer a state-machine as `save_data` is gone (and those ugly `assert` statements) in favor of this new constructor parameter.

qiime/sdk/artifact.py
```diff
@@ -34,45 +35,44 @@ def _from_view(cls, view, type_, provenance):
         ----------
         view : Python object
             View to serialize.
-        type_ : qiime.plugin.Type or str
```
jairideout: We aren't using a `str` `type_` anymore, and this is a private method, so I removed for simplicity.

qiime/sdk/tests/test_artifact.py
```diff
@@ -297,6 +298,32 @@ def test_load_from_externally_created_zipfile(self):
             }
             self.assertEqual(fps, expected)
 
+    def test_load_with_archive_filepath_modified(self):
```
jairideout: I got way too excited when this test finally started passing.

jairideout:
@ebolyen this is ready for review. I made a few comments to help you as you review (sorry parts of these diffs are painful).

ebolyen:
This is awesome @jairideout!

jairideout:
Thanks @ebolyen!

PR #79, opened 2016-07-08
ENH/REF: add public qiime.sdk.Result class/factory for interacting with Artifacts and Visualizations in a generic way

jairideout:
`Result.load`, `Result.peek`, and `Result.extract` act as generic interfaces/factories for interacting with Artifacts and Visualizations.  `Result` prevents an Artifact from being interacted with as a Visualization, and vice-versa (in addition to other incompatible types we may have in the future).


qiime/sdk/result.py
```diff
@@ -0,0 +1,212 @@
...
+                   type(result).__name__))
+
+        result._archiver = archiver
+        return result
+
+    @classmethod
+    def extract(cls, filepath, output_dir):
+        # `peek` first to ensure the type is valid for the `cls` this was
+        # called on.
+        cls.peek(filepath)
```
ebolyen: The error message from `peek` will look a little weird here, but we can refactor that later.

ebolyen:
Thanks @jairideout!

gregcaporaso:
Thanks @jairideout and @ebolyen!

jairideout:
Thanks for reviewing Evan! Agree about the error message from extract,
maybe it can be rewritten to make sense in either context, or extract can
catch the error and raise a new one.

On Friday, July 8, 2016, Greg Caporaso <notifications@github.com> wrote:

> Thanks @jairideout <https://github.com/jairideout> and @ebolyen
> <https://github.com/ebolyen>!
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/qiime2/qiime2/pull/79#issuecomment-231433750>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABwvwKwVka6VoF7Clr2Joq5k22L3ihl6ks5qTpQxgaJpZM4JIQVu>
> .
>


-- 
Sent from my phone.

Issue #73, opened 2016-07-05
Visualizers fail on OSX through async calls, due to matplotlib main thread limitations


jakereps:
Testing out the summarize visualizer from the feature-table plugin through qiime-studio revealed an issue where matplotlib does not work outside of the main thread on OSX. The Visualizer fails with the following traceback:

``` bash
The process has forked and you cannot use this CoreFoundation functionality safely. You MUST exec().
Break on __THE_PROCESS_HAS_FORKED_AND_YOU_CANNOT_USE_THIS_COREFOUNDATION_FUNCTIONALITY___YOU_MUST_EXEC__() to debug.
The process has forked and you cannot use this CoreFoundation functionality safely. You MUST exec().
Break on __THE_PROCESS_HAS_FORKED_AND_YOU_CANNOT_USE_THIS_COREFOUNDATION_FUNCTIONALITY___YOU_MUST_EXEC__() to debug.
Traceback (most recent call last):
  File "/Users/Develop/Developer/work/qiime-studio/qiime_studio/api/jobs.py", line 111, in callback
    results = future.result()
  File "/Users/Develop/anaconda/envs/qiime_studio/lib/python3.5/concurrent/futures/_base.py", line 398, in result
    return self.__get_result()
  File "/Users/Develop/anaconda/envs/qiime_studio/lib/python3.5/concurrent/futures/_base.py", line 357, in __get_result
    raise self._exception
concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.
```

Found it was due to matplotlib through this SO exchange: http://stackoverflow.com/a/16303620
Sounds like this might need another workaround.

ebolyen:
Thanks @jakereps! I forgot about that issue, this is likely something we will  need to find a workaround for, for SciPy. @jairideout, @gregcaporaso are you familiar with this error?

jairideout:
I'm not familiar with this error. Does it still fail if you set the matplotlib backend to Agg (e.g., in your matplotlib config)? I wonder if it's just a GUI backend issue or more generally an issue with matplotlib. OS X ships with a matplotlib GUI backend by default.

gregcaporaso:
I'm also not familiar with this one.

ebolyen:
@gregcaporaso was able to summarize without running into this issue. So we'll need to figure out how to reproduce this...

jairideout:
[...]

PR #83, opened 2016-07-08
FIX: Notify users of invalid matplotlib backend

jakereps:
Temporary solution for #73

```bash
In [2]: from qiime.plugin.feature_table.visualizers import summarize

In [3]: from qiime.sdk import Artifact

In [4]: summarize.async(table=Artifact.load('feature-table-frequency.qza')).result()
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-4-a9c7a68dedce> in <module>()
----> 1 summarize.async(table=Artifact.load('feature-table-frequency.qza')).result()

<decorator-gen-140> in summarize(table)

/Users/Develop/Developer/work/qiime2/qiime/sdk/visualizer.py in async_wrapper(*args, **kwargs)
    164             if plt.rcParams['backend'] == 'MacOSX':
    165                 raise EnvironmentError(backend_error_template %
--> 166                                        plt.matplotlib_fname())
    167             args = args[1:]
    168             pool = concurrent.futures.ProcessPoolExecutor(max_workers=1)

OSError: 
Your current matplotlib backend (MacOSX) does not work with async calls.
A recommended backend is Agg, and can be changed by modifying your
matplotlibrc "backend" parameter, which can be found at: 

 /Users/Develop/anaconda/envs/qiime_studio/lib/python3.5/site-packages/matplotlib/mpl-data/matplotlibrc
```

### DIFF
--- a/qiime/sdk/visualizer.py
+++ b/qiime/sdk/visualizer.py
@@ -158,6 +158,15 @@ class Visualizer:
         # TODO `async` execution has some problems with garbage-collection in
         # the subprocess given certain inputs. Needs further investigation.
         def async_wrapper(*args, **kwargs):
+            # TODO handle this better in the future, but stop the massive error
+            # caused by MacOSX async runs for now.
+            try:
+                import matplotlib as plt
+                if plt.rcParams['backend'] == 'MacOSX':
+                    raise EnvironmentError(backend_error_template %
+                                           plt.matplotlib_fname())
+            except ImportError:
+                pass
             args = args[1:]
             pool = concurrent.futures.ProcessPoolExecutor(max_workers=1)
             future = pool.submit(self, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


qiime/sdk/visualizer.py
```diff
@@ -158,6 +158,12 @@ def _bind_executors(self):
         # TODO `async` execution has some problems with garbage-collection in
         # the subprocess given certain inputs. Needs further investigation.
         def async_wrapper(*args, **kwargs):
+            # TODO handle this better in the future, but stop the massive error
+            # caused by MacOSX async runs for now.
+            import matplotlib as plt
```
jairideout: Can you wrap this in a try/except for `ImportError` so that matplotlib is an optional dependency? Then matplotlib can be removed from .travis.yml and setup.py.

jairideout:
Thanks @jakereps! Just had one comment about matplotlib as an optional dependency.

jakereps:
Thanks @jairideout! Should be good to go, and that definitely makes more sense to keep it out of the requirements when it isn't actually used to do anything else.

jairideout:
Thanks @jakereps!

PR #85, opened 2016-07-11
ENH: Artifact import support (`Artifact.import_data`)

### LINKED ISSUE #31: create a mechanism for importing data into artifacts

jairideout:
Fixes #31.


ebolyen:
Thanks @jairideout!

PR #87, opened 2016-07-14
ENH: push Artifact/Visualization down to base package

ebolyen:
This will make things cleaner.


gregcaporaso:
Looks good, let's wait for the tests (even though we prob don't need them for this).

PR #95, opened 2016-07-18
TST: Check Archiver file encoding and decoding uses UTF-8.

thermokarst:
This is a quick rough-cut of some tests that check for UTF-8 during encoding and decoding.

- Set `errors='strict'` in existing readers (`_load_version`, `_load_metadata`). This will ensure that a `TypeError` is raised if there is an encoding mismatch. Not sure if we need to raise our own exception at this point, but feel free to comment.
- Manually encode the payloads for the existing writers (`_save_version`, `_save_readme`, `_save_metadata`)

Questions:
- I used a dummy class for testing the relevant Archiver methods discussed above: thoughts on that? Maybe it makes more sense to set up a real Archiver instance for these tests. Let me know.
- Do we want to set a package constant (or class, etc.) that specifies the encoding to use? Right now `'utf-8'` is sprinkled throughout as arguments.
- I didn't touch anything outside of the artifact metadata/info/version domain, let me know if there is something I missed.

Thanks!

### DIFF
--- a/qiime/core/archiver.py
+++ b/qiime/core/archiver.py
@@ -19,7 +19,6 @@ import qiime.sdk
 from .util import parse_type
 
 
-# TODO use utf-8 encoding when reading/writing files
 class Archiver:
     # This class will likely defer to one of many ArchiveFormats in the future.
     # There's only one supported archive format currently.
@@ -74,7 +73,7 @@ class Archiver:
         version_path = os.path.join(root_dir, cls._VERSION_FILENAME)
         with zf.open(version_path) as bytes_fh:
             with io.TextIOWrapper(bytes_fh, newline=None,
-                                  encoding='utf-8') as fh:
+                                  encoding='utf-8', errors='strict') as fh:
                 return fh.read().rstrip('\n')
 
     @classmethod
@@ -82,7 +81,7 @@ class Archiver:
         metadata_path = os.path.join(root_dir, cls._METADATA_FILENAME)
         with zf.open(metadata_path) as bytes_fh:
             with io.TextIOWrapper(bytes_fh, newline=None,
-                                  encoding='utf-8') as fh:
+                                  encoding='utf-8', errors='strict') as fh:
                 metadata = yaml.safe_load(fh)
 
         uuid_ = cls._parse_uuid(metadata['uuid'])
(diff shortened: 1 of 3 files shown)


qiime/core/archiver.py
```diff
@@ -211,12 +210,13 @@ def save(self, filepath):
                     zf.write(abspath, arcname=archive_path)
 
     def _save_version(self, zf, root_dir):
+        version = '%s\n' % self._VERSION
         zf.writestr(os.path.join(root_dir, self._VERSION_FILENAME),
-                    '%s\n' % self._VERSION)
+                    version.encode('utf-8'))
```
ebolyen: I suspect (but need to prove) that this will actually write `b'1'` to the `VERSION` file...

thermokarst: Hmm, my understanding in Python 3 is that `b''` represents a bytestring literal, which would just write out the octet sequence, but maybe I am totally off target here.

ebolyen: It does, but the weirdness is in that `str(bytes(b'a'))` -> `"b'a'"` (I guess to punish people who don't understand encoding?) and since the method is `writestr` I suspect it may be encoding already (I'm going to check this).

ebolyen: Looks like it works fine, so `zipfile` just has terrible API names (no surprise).

qiime/core/tests/test_archiver.py
```diff
@@ -0,0 +1,127 @@
...
+    def setUp(self):
+        prefix = "qiime2-test-temp-"
+        self.temp_dir = tempfile.TemporaryDirectory(prefix=prefix)
+        self.root_dir = self.temp_dir.name
+
+        # FakeArchiver is for testing some of the private methods on Archiver
+        # that don't require the full class.
+        class FakeArchiver(Archiver):
+            # Specifically not calling super() because we don't need to
+            # drag any of Archiver's baggage into this.
```
ebolyen: harsh!

qiime/core/tests/test_archiver.py
```diff
@@ -0,0 +1,127 @@
...
+        # non-UTF-8 encoded version file.
+        with zipfile.ZipFile(fp, mode="w",
+                             compression=zipfile.ZIP_DEFLATED,
+                             allowZip64=True) as zf:
+            zf.writestr(os.path.join(self.root_dir,
+                                     Archiver._VERSION_FILENAME),
+                        "version info".encode("utf-16"))
+
+            with self.assertRaises(UnicodeDecodeError):
+                Archiver._load_version(zf, self.root_dir)
```
ebolyen: :+1:

ebolyen:
This is awesome, thanks @thermokarst!

jairideout:
@ebolyen do the existing plugins need updating to use a specific encoding and/or newline character? The readers/writers are currently using universal-newline mode.

ebolyen:
@jairideout I think they do, it should just be `q2-types` as that is defining all of the data-layouts which is what we are trying to make more deterministic.

jairideout:
Sounds good, can you create an issue on q2-types so we don't forget? I'd create one but I'm not clear on what exactly needs to change in the data layout readers/writers (and file formats?).

It's starting to look like we need a richer API than supplying `data_dir` to readers/writers since this is really easy to mess up and not notice until it comes up in the wild (plus it's some boilerplate that plugin devs shouldn't need to write). What do you think?

ebolyen:
@jairideout Will do.

Re: Richer API; Perhaps, the downside is they would have to interact with file-objects at that point, so we would want this API to come with the ability to get just a file to make wrapping tools easier.

PR #96, opened 2016-07-18
ENH: Method, validate markdown

thermokarst:
Adds basic checks for length of `type_tuple`, raises an exception if value is not two items long.

Questions:
- Do we need to raise separate errors for each metadata section? Or, could this be collapsed down into exception handling in `_split_type_tuple()`?
- If we are keeping the handling as-is, I imagine the error text could use a tune-up. Thoughts?
- Instead of saving temp MD files in the tests, should I be building this up as a permanent bad example in `qiime.core.testing`?

Thanks!

### DIFF
--- a/qiime/sdk/tests/test_method.py
+++ b/qiime/sdk/tests/test_method.py
@@ -9,6 +9,7 @@
 import collections
 import concurrent.futures
 import inspect
+import pkg_resources
 import unittest
 import uuid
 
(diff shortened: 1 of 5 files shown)


qiime/sdk/tests/test_method.py
```diff
@@ -374,6 +376,93 @@ def test_async(self):
...
                              [10, 20, 0, 42, 43, 99, -22, 55, 1])
 
+    def test_markdown_input_section_validation(self):
+        temp_dir = tempfile.TemporaryDirectory(prefix="qiime2-test-temp-")
+        root_dir = temp_dir.name
+        fp = os.path.join(root_dir, "bad_markdown.md")
+
+        with open(fp, mode="w") as fh:
+            fh.write("""---
+name: Bad markdown
```
jairideout: Can you put each of the invalid markdown strings into files under `qiime/sdk/tests/data/`? You can then access those files in your tests using:

```python
fp = pkg_resources.resource_filename('qiime.sdk.tests', 'data/my-markdown-file.md')
```

See [this test case](https://github.com/qiime2/qiime2/blob/master/qiime/sdk/tests/test_artifact.py#L256-L257) for an example.

qiime/sdk/tests/test_method.py
```diff
@@ -374,6 +376,93 @@ def test_async(self):
...
+parameters:
+    - int1:
+        - Int
+        - int
+outputs:
+    - concatenated_ints:
+        - IntSequence1
+        - list
+---""")
+        with self.assertRaises(TypeError) as te:
```
jairideout: It'd be preferable to use

```python
self.assertRaisesRegex(TypeError, 'input section'):
    ...
```

so that the test case isn't too dependent on the exact error message that's raised, in case that changes in the future. I recommend using a similar regex for the other error cases (e.g. `'parameters section'`, `'outputs section'`).

jairideout:
Thanks @thermokarst!

> Do we need to raise separate errors for each metadata section? Or, could this be collapsed down into exception handling in _split_type_tuple()?

I like how you have separate error messages to help plugin devs narrow down where their markdown is mis-formatted. 

> If we are keeping the handling as-is, I imagine the error text could use a tune-up. Thoughts?

The error messages look good as-is since they'll be displayed to plugin devs and won't be seen by users.

> Instead of saving temp MD files in the tests, should I be building this up as a permanent bad example in qiime.core.testing?

I think it makes sense to have the invalid markdown as files under `tests/data/`. I commented inline with details about how to accomplish that. I think it'll make the test code more readable by avoiding large inline markdown strings. If other tests need those files in the future we can put them under `qiime.core.testing` but they seem to be pretty specific to `Method`'s unit tests.

jairideout:
@thermokarst done reviewing!

jairideout:
Thanks @thermokarst!

PR #98, opened 2016-07-19
ENH: Pretty-print metadata YAML

thermokarst:
Opening this as a placeholder for work right now. I based this initial cut off of #95, since both PRs introduce tests for `qiime.core`, and will rebase once #95 is merged.

This adds `default_flow_style` kwarg to the metadata YAML dump, which writes out standard YAML.

### DIFF
--- a/qiime/core/tests/test_archiver.py
+++ b/qiime/core/tests/test_archiver.py
@@ -24,8 +24,7 @@ class TestArchiver(unittest.TestCase):
         # FakeArchiver is for testing some of the private methods on Archiver
         # that don't require the full class.
         class FakeArchiver(Archiver):
-            # Specifically not calling super() because we don't need to
-            # drag any of Archiver's baggage into this.
+            # Explicitly not calling super()
             def __init__(self, *args, **kwargs):
                 self._temp_dir = tempfile.mkdtemp(prefix)
                 self._data_dir = os.path.join(self._temp_dir,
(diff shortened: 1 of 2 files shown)


thermokarst:
Questions:

- Do we want this output aligned? e.g.
```yaml
provenance: baz
type:       '''bar'''
uuid:       foo
```
- Seems like we should be using an OrderedDict in the call to `yaml.safe_dump` so that we can guarantee consistent output (I suspect the new test in this PR is fragile due to this). Looks like we would need to [implement a dumper](http://stackoverflow.com/a/21912744/313548) for serialization, but that isn't too crazy.

ebolyen:
Alignment isn't necessary I think, and an `OrderedDict` makes sense.

qiime/core/tests/test_archiver.py
```diff
@@ -125,3 +124,21 @@ def test_save_metadata_file_encoding(self):
...
+        fp = os.path.join(self.root_dir, "archive.qza")
+
+        with zipfile.ZipFile(fp, mode="w",
+                             compression=zipfile.ZIP_DEFLATED,
+                             allowZip64=True) as zf:
+            archive._save_metadata(zf, self.root_dir)
+
+            with zf.open(os.path.join(self.root_dir,
+                                      Archiver._METADATA_FILENAME)) as zip_fh:
+                with io.TextIOWrapper(zip_fh, newline=None,
```
ebolyen: This is some serious nesting! I don't see any good way to simplify this though.

qiime/core/tests/test_archiver.py
```diff
@@ -125,3 +124,21 @@ def test_save_metadata_file_encoding(self):
                 with io.TextIOWrapper(bytes_fh, newline=None,
                                       encoding='utf-8', errors='strict') as fh:
                     self.assertTrue(fh.read().rstrip('\n'))
+
+    def test_metadata_pprint_yaml(self):
+        archive = self.FakeArchiver()
+        fp = os.path.join(self.root_dir, "archive.qza")
```
ebolyen: This should be a temporary file of some kind, unless I'm missing something, this file won't get cleaned up after the tests are finished.

ebolyen: never mind, it's created in a temporary directory, we're fine.

qiime/core/tests/test_archiver.py
thermokarst: Agreed! I guess I could not use CM's, and then have a few extra lines for cleanup. I don't see that as being particularly better, though. Your call!

ebolyen: No, definitely use the context managers!

ebolyen:
Works great, and I've confirmed these changes are backwards compatible, thanks @thermokarst!

PR #100, opened 2016-07-19
DOC: Fix LICENSE holder and fix old links

jakereps:


jairideout:
@jakereps there's a couple test failures that had those links in them. Can you update those?

jakereps:
Is that something we'll want to change in the sample artifacts, or just let those be `/biocore/`?

jairideout:
I think leaving those as `/biocore/` is fine, it's really just a placeholder until we have provenance finished. I'm also fine with updating the unit tests to be less specific in their assertions of that string (they're doing exact matches right now but could be modified to check that the string `startswith` the expected value).

jairideout:
👍 👍 👍

PR #103, opened 2016-07-21
ENH: support for QIIME 1 feature metadata files in qiime.Metadata

jairideout:
qiime.Metadata and qiime.MetadataCategory reprs use underlying DataFrame repr instead of JSON. Added HTML repr to qiime.Metadata.

Pair-programmed with @jakereps.

### DIFF
--- a/qiime/metadata.py
+++ b/qiime/metadata.py
@@ -14,13 +14,15 @@ class Metadata:
         self._dataframe = dataframe
 
     def __repr__(self):
-        return self._dataframe.to_json()
+        return repr(self._dataframe)
+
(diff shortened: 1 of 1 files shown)


ebolyen:
I like this, just need to test with `studio`/`q2cli` in case they have janky uses of the `repr`

jairideout:
We tested the studio and q2cli using the emperor plugin which accepts metadata as input. Things appear to be non-janky!

ebolyen:
Thanks @jairideout and @jakereps!

ElDeveloper:
:tada:

PR #106, opened 2016-07-23
ENH: Add class-level extensions to Artifact, Visualization

### LINKED ISSUE #7: warn if recommended artifact file extension isn't used
`Artifact` and `Visualization` should warn when `load` or `save` is called with a file path that doesn't have the recommended file extension. This issue depends on #6.

### LINKED ISSUE #59: define standard file extensions so they can be imported by interfaces
The extensions ``qza`` and ``qzv`` should be discoverable via the sdk.

jakereps:
Allows interfaces to use `Artifact.extension` or `Visualization.extension` to access the currently supported/recommended file extensions. 

Fix a few tests that were found to be using `visualization.qza` instead of `.qzv` through the warnings.

Example Output:

```shell
>>> Artifact.load('/Users/Develop/Developer/work/q2-types/q2_types/tests/data/feature-table-frequency.qza').save('test')
/Users/Develop/Developer/work/qiime2/qiime/sdk/result.py:236: UserWarning: The most widely supported and recommended file extension for a QIIME 2 Artifact is '.qza'. The Artifact will still be created, but the provided file extension can not be guaranteed to work with all QIIME 2 interfaces.
```

Resolves #7
Resolves #59

### DIFF
--- a/qiime/core/util.py
+++ b/qiime/core/util.py
@@ -7,6 +7,8 @@
 # ----------------------------------------------------------------------------
 
 import re
+import contextlib
+import warnings
 
 
 def tuplize(x):
@@ -93,3 +95,16 @@ def parse_type(string, expect=None):
 # Makes it possible to programmatically detect when a type doesn't exist.
 class UnknownTypeError(TypeError):
     pass
+
+
+@contextlib.contextmanager
+def warning():
+    def _warnformat(msg, category, filename, lineno, file=None, line=None):
+        return '%s:%s: %s: %s\n' % (filename, lineno, category.__name__, msg)
+
+    default_warn_format = warnings.formatwarning
+    try:
+        warnings.formatwarning = _warnformat
+        yield warnings.warn
+    finally:
+        warnings.formatwarning = default_warn_format
--- a/qiime/sdk/result.py
+++ b/qiime/sdk/result.py
@@ -17,6 +17,7 @@ import uuid
 import qiime.sdk
 import qiime.core.archiver
 import qiime.core.type
+import qiime.core.util
 
 # Note: Result, Artifact, and Visualization classes are in this file to avoid
 # circular dependencies between Result and its subclasses. Result is tightly
@@ -118,6 +119,8 @@ class Result:
 
 
 class Artifact(Result):
+    extension = '.qza'
+
     @classmethod
     def _is_valid_type(cls, type_):
         if qiime.core.type.is_semantic_type(type_) and type_.is_concrete():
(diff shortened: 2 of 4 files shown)


qiime/sdk/tests/test_visualization.py
```diff
@@ -385,7 +385,7 @@ def get_abs_path(rel):
 
     def test_peek(self):
```
thermokarst: Maybe we should add another test that asserts that the warning is raised when saving with the `qza` extension?

jakereps:
Thanks @ebolyen and @thermokarst, this should be ready for review!

qiime/core/util.py
```diff
@@ -93,3 +95,16 @@ def parse_type(string, expect=None):
...
+
+@contextlib.contextmanager
+def warning():
+    def _warnformat(msg, category, filename, lineno, file=None, line=None):
+        return '%s:%s: %s: %s\n' % (filename, lineno, category.__name__, msg)
+
+    default_warn_format = warnings.formatwarning
+    try:
+        warnings.formatwarning = _warnformat
+        yield None
```
ebolyen: `yield warnings.warn` so that we can adjust all QIIME warnings easily.

qiime/sdk/result.py
```diff
@@ -223,8 +227,17 @@ def view(self, view_type):
         reader = data_layout.readers[view_type]
         return self._archiver.load_data(reader)
 
+    def save(self, filepath):
+        if not filepath.endswith(self.extension):
+            with qiime.core.util.warning():
```
ebolyen: this would become:
```python
with qiime.core.util.warning as warn:
    warn(...)
```

qiime/sdk/tests/test_result.py
```diff
@@ -219,6 +220,30 @@ def test_peek_visualization(self):
...
+    def test_save_artifact_warning(self):
+        artifact = Artifact._from_view([0, 0, 42, 1000], FourInts,
+                                       self.provenance)
+        fp = os.path.join(self.test_dir.name, 'artifact.zip')
+
+        with warnings.catch_warnings(record=True) as w:
+            artifact.save(fp)
+
+            self.assertEqual(len(w), 1)
+            self.assertTrue(issubclass(w[0].category, UserWarning))
```
ebolyen: This looks a little strange, is there not a way to use `assertIsInstance`?

jakereps: It doesn't like this one for some reason. 

```python
# using just w[0]
AssertionError: <warnings.WarningMessage object at 0x10ee1d128> is not an instance of <class 'UserWarning'>
# using w[0].category
AssertionError: <class 'UserWarning'> is not an instance of <class 'UserWarning'>
```

jakereps:
Thanks for the review @ebolyen, and @thermokarst! Should be good after tests!

PR #107, opened 2016-07-25
ENH: Add helper URLs to qiime.sdk

### LINKED ISSUE #80: various URLs should be importable from the framework
including:
* citation (QIIME 1 for now: http://www.ncbi.nlm.nih.gov/pubmed/20383131)
* help page: http://help.qiime.org
* conda channel: https://anaconda.org/qiime2

these should probably be globals under SDK.

jakereps:
Adds citation url, help page, and conda channel URLs to qiime.sdk

Resolves #80

### DIFF
--- a/qiime/sdk/__init__.py
+++ b/qiime/sdk/__init__.py
@@ -15,3 +15,8 @@ from ..core.util import parse_type
 
 __all__ = ['Result', 'Artifact', 'Visualization', 'Method', 'Visualizer',
            'PluginManager', 'Provenance', 'parse_type']
+
+# Various URLs
+CITATION = 'http://www.ncbi.nlm.nih.gov/pubmed/20383131'
+HELP_URL = 'http://2.qiime.org'
+CONDA_CHANNEL = 'https://anaconda.org/qiime2'


qiime/sdk/__init__.py
```diff
@@ -15,3 +15,8 @@
 
 __all__ = ['Result', 'Artifact', 'Visualization', 'Method', 'Visualizer',
            'PluginManager', 'Provenance', 'parse_type']
+
+# Various URLs
+citation = 'http://www.ncbi.nlm.nih.gov/pubmed/20383131'
+help = 'http://help.qiime.org'
```
ebolyen: This is a builtin that we don't want to override.

qiime/sdk/__init__.py
```diff
@@ -15,3 +15,8 @@
 
 __all__ = ['Result', 'Artifact', 'Visualization', 'Method', 'Visualizer',
            'PluginManager', 'Provenance', 'parse_type']
+
+# Various URLs
+citation = 'http://www.ncbi.nlm.nih.gov/pubmed/20383131'
```
ebolyen: I would make these all-caps so that they look like module-constants

qiime/sdk/__init__.py
```diff
@@ -15,3 +15,8 @@
...
+help = 'http://help.qiime.org'
+conda = 'https://anaconda.org/qiime2'
```
ebolyen: CONDA_CHANNEL

qiime/sdk/__init__.py
ebolyen: HELP_URL

ebolyen:
Thanks @jakereps!

PR #110, opened 2016-08-05
ENH: Adds qiime.plugin.plugin_init to initialize plugin package from …

Kleptobismol:
…cookiecutter template

Pair-programmed with @jairideout

### DIFF
--- a/qiime/plugin/template.py
+++ b/qiime/plugin/template.py
@@ -1,0 +1,39 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016--, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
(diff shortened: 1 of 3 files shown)


qiime/plugin/template.py
```diff
@@ -0,0 +1,36 @@
...
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+from cookiecutter.main import cookiecutter
+
+
+def plugin_init(output_dir='.'):
```
ebolyen: This will result in a subdirectory right? It doesn't template *inside* the current directory?

jairideout: It will create a subdirectory in `output_dir` based on the user's input

jairideout: We made the `output_dir` docs more explicit to clear up any confusion

PR #116, opened 2016-08-16
ENH: added Bool, addresses #115

wasade:
I'm still green with the codebase, so apologize if this is off-base. I believe it's to the expectation. Have support for `int`, where `0 is False` and any other number (positive or negative) is `True`:

```python
In [1]: from primitive import Bool

In [2]: Bool.decode("10")
Out[2]: True

In [3]: Bool.decode("0")
Out[3]: False

In [4]: Bool.decode("True")
Out[4]: True

In [5]: Bool.decode("False")
Out[5]: False
```

Note: Python's `bool` casting is unexpected hence the direct checks against `"0"` and `"False"`. See below, where the expectation is that `bool` would operate like `int` but it actually just appears to test for a non-empty string:

```python
In [1]: bool("True")
Out[1]: True

In [2]: bool("False")
Out[2]: True

In [3]: bool("")
Out[3]: False
```

### DIFF
--- a/qiime/core/type/__init__.py
+++ b/qiime/core/type/__init__.py
@@ -7,7 +7,7 @@
 # ----------------------------------------------------------------------------
 
 from .semantic import SemanticType, is_semantic_type, Properties
-from .primitive import (Str, Int, Float, Color, Dict, Set, Metadata,
+from .primitive import (Str, Int, Float, Color, Dict, Set, Metadata, Bool,
                         MetadataCategory, List, Range, Choices, Arguments,
                         is_primitive_type)
 from .grammar import is_qiime_type
@@ -16,7 +16,7 @@ from .signature import Signature
 
 __all__ = [
     'SemanticType', 'is_qiime_type', 'is_semantic_type', 'is_primitive_type',
-    'Str', 'Int', 'Float', 'Color', 'Dict', 'Set', 'List', 'Metadata',
+    'Str', 'Int', 'Float', 'Bool', 'Color', 'Dict', 'Set', 'List', 'Metadata',
     'MetadataCategory', 'Visualization', 'Signature', 'Properties', 'Range',
     'Choices', 'Arguments'
 ]
--- a/qiime/core/type/primitive.py
+++ b/qiime/core/type/primitive.py
@@ -349,6 +349,27 @@ class _Float(_Primitive):
 Float = _Float('Float')
 
 
+class _Bool(_Primitive):
+    _valid_predicates = {Arguments, }
+
+    def _is_element_(self, value):
+        return isinstance(value, bool)
+
+    def decode(self, string):
+        if string not in ('false', 'true'):
+            raise TypeError("%s is neither 'true' or 'false'" % string)
+
+        return string == 'true'
+
+    def encode(self, value):
+        if value:
+            return 'true'
+        else:
+            return 'false'
+
+Bool = _Bool('Bool')
+
+
 class _Color(type(Str)):
     def _is_element_(self, value):
         # Regex from: http://stackoverflow.com/a/1636354/579416
--- a/qiime/plugin/__init__.py
+++ b/qiime/plugin/__init__.py
@@ -13,9 +13,9 @@ from .template import plugin_init
 
 from qiime.core.type import (SemanticType, Int, Str, Float, Color, Metadata,
                              MetadataCategory, Properties, Range, Choices,
-                             Arguments)
+                             Arguments, Bool)
 
 
 __all__ = ['DataLayout', 'FileFormat', 'Plugin', 'SemanticType', 'Int', 'Str',
            'Float', 'Color', 'Metadata', 'MetadataCategory', 'Properties',
-           'Range', 'Choices', 'Arguments', 'plugin_init']
+           'Range', 'Choices', 'Arguments', 'plugin_init', 'Bool']


qiime/core/type/primitive.py
```diff
@@ -349,6 +349,30 @@ def encode(self, value):
 Float = _Float('Float')
 
 
+class _Bool(_Primitive):
+    _valid_predicates = {Choices, Arguments}
```
ebolyen: I don't think the `Choices` predicate makes sense. Anything but `Bool % Choices({True, False})` isn't worth parametrizing and that expression is identical to just `Bool`

qiime/core/type/primitive.py
```diff
@@ -349,6 +349,30 @@ def encode(self, value):
...
+
+    def _is_element(self, value):
```
ebolyen: missing trailing `_`

qiime/core/type/primitive.py
```diff
@@ -349,6 +349,30 @@ def encode(self, value):
...
+        if (isinstance(other, type(Int)) and
+                self.predicate <= other.predicate):
+            return True
+        return super()._is_subtype_(other)
+
+    def decode(self, string):
+        if string in ("False", "0"):
+            return False
+        else:
+            return True
```
ebolyen: This behavior probably makes sense, though I think `false` being an option would also make life easier. Otherwise `"false"` is true, which will trip up an interface in the future (the typical consumer of `decode`).

ebolyen:
`Bool` will need to be exposed at `qiime.core.types` and `qiime.plugins`

qiime/core/type/primitive.py
wasade: oh, I see, thanks!

BTW, are there corresponding docs for the `type` module?

ebolyen:
Thanks @wasade! This will be really useful, (we'll also need to update the interfaces so that they understand `Bool`, any volunteers?)

qiime/core/type/primitive.py
ebolyen: Thinking more on it, right now `encode` and `decode` are implicitly JSON types, so it might make more sense to just match that exclusively.

wasade:
Thanks! I think I covered the comments. I can take a look at the interfaces but I'm still totally green :)

ebolyen:
Thanks @wasade! `q2cli` is probably the easier to modify right now. Tests should pass on the next round.

ebolyen:
Hmm or not.

qiime/core/type/__init__.py
```diff
@@ -16,7 +16,7 @@
```
ebolyen: Missing the import here

wasade:
My bad :) pushed to quick

wasade:
@ebolyen, is the interface work blocking for this PR or can it be deferred? If review etc passes, I'll squash commits

qiime/core/type/primitive.py
```diff
@@ -349,6 +349,24 @@ def encode(self, value):
...
 
+class _Bool(_Primitive):
+    _valid_predicates = {Arguments,}
+
+    def _is_element_(self, value):
+        return isinstance(value, bool)
+
+    def decode(self, string):
+        if string not in ('false', 'true'):
+            raise TypeError("%s is neither 'true' or 'false'")
```
jairideout: This string interpolation is missing the interpolated variable `string`

qiime/core/type/primitive.py
```diff
@@ -349,6 +349,24 @@ def encode(self, value):
...
+
+        return string == 'true'
+
+    def encode(self, value):
+        return str(value)
```
jairideout: `encode` should return a value that can be decoded with `decode`, so this should return `'true'` or `'false'`

jairideout:
@wasade a couple comments from a quick pass through. I'll let @ebolyen do final review/merge

jairideout:
> is the interface work blocking for this PR or can it be deferred?

The interface work shouldn't block this. We may want to create issues on the relevant interface repos to add support for `Bool` just so we don't forget.

qiime/core/type/primitive.py
wasade: This isn't the case for `Int` or the other primitives though...?

jairideout: `str(42)` returns `'42'`, which is the correct JSON representation of an integer. `str(True)` returns `'True'`, which is not a valid JSON boolean type. Does that make sense? `encode`/`decode` really need to be renamed `to_json`/`from_json`.

wasade: Right, but Int.encode(1.234) -> is float

wasade: ...so I'm okay with doing the `encode` as described but I guess I'm still confused as to why the enforcement for `Bool` but non-enforcement for the other primitives? Is the goal that `encode` is valid JSON without assurance that it is valid w.r.t. the type? All of these are valid but incorrect:

```python
In [1]: from qiime.core.type.primitive import Set, Float, Int

In [2]: Set.encode('foo')
Out[2]: '["f", "o", "o"]'

In [3]: Float.encode('foo')
Out[3]: 'foo'

In [4]: Int.encode('foo')
Out[4]: 'foo'
```

ebolyen: I think you're reading way too much into the current behavior. Arguably everything you've mentioned is really more of a bug than intentional. I wouldn't consider anything the framework does right now as *canonical*, least of all anything in `qiime.core` and especially not `qiime.core.type` which is probably the most incomplete part of QIIME atm. There also hasn't been any use of `encode` which is why no one has noticed the above yet, so the current implementation is rather generous in it's assumption as to input. I don't know yet if that is necessarily a problem either, but it would be trivially solved by first asking if the value passed to `encode` was `in` the type (that functionality *should* be relatively robust, e.g. `'foo' in Float` -> `False`)

ebolyen: I'm not even certain `encode` should exist, it doesn't seem to be necessary. but we might find reasons to care when we start considering provenance more seriously. And in that context, the initial type-checking would have failed (via `in`) already if they weren't the right type, so there would be nothing to record)

wasade: Okay, thanks. Not sure how far to read into things yet :) but totally understand that much of this is still annealing. I'll go with @jairideout's suggestion right now. Okay with avoiding the `in` check for now given that `encode` may go away?

ebolyen: Yeah avoiding `in` is fine, if we were to do that we'd likely just apply that check to a base class generically.

ebolyen:
Looks good, thanks @wasade, can you collapse?

wasade:
@ebolyen squashed

ebolyen:
Thanks!

jairideout:
Thanks @wasade and @ebolyen! I created issues to support `Bool` in q2cli and qiime-studio:

https://github.com/qiime2/q2cli/issues/52

https://github.com/qiime2/qiime-studio/issues/70

PR #117, opened 2016-08-22
ENH: Add TempPath type, subclasses pathlib.Path

thermokarst:


qiime/core/path.py
```diff
@@ -0,0 +1,46 @@
...
+        super().__exit__(t, v, tb)
+
+    def _finalize(self):
+        if not self.exists():
+            return
+
+        if self.is_dir():
+            shutil.rmtree(str(self))
+        else:
+            self.unlink()
```
ebolyen: Is unlink idempotent? I assume it exists because `super().__exit__` uses it for something.

qiime/sdk/visualizer.py
```diff
@@ -219,14 +219,14 @@ def callable_wrapper(*args, **kwargs):
                 parameter_references)
 
             # TODO use user-configured temp dir
-            with tempfile.TemporaryDirectory('qiime2-temp-') as temp_dir:
-                ret_val = self._callable(output_dir=temp_dir, **view_args)
+            with TempPath(dir=True) as temp_dir:
+                ret_val = self._callable(output_dir=str(temp_dir), **view_args)
```
ebolyen: We might want to actually just export `temp_dir` directly, but I'm fine with this for the sake of moving forward. @jairideout thoughts on this?

qiime/core/path.py
thermokarst: No, `Path.unlink` is not idempotent. `Path.unlink` removes files/links at time of execution, and `Path.rmdir` removes empty directories. I decided to use `shutil.rmtree` to handle removing non-empty dirs.

qiime/sdk/visualizer.py
thermokarst: I agree on both points, curious to know what @jairideout thinks, too.

qiime/core/path.py
ebolyen: I see, the context manager adds an idea of opening and closing "paths" (not
files), so unlink isn't called anyways in the parent-classes' exit.

ebolyen:
Looks good, merging so we can move forward, thanks @thermokarst!

qiime/sdk/visualizer.py
jairideout: 👍 for making `temp_dir` a directory path (`str`) when used in a context manager (that will match `tempfile.TemporaryDirectory`). Is that what you guys are wanting input on?

jairideout: Didn't realize this was merged already, not a big deal if you guys want to keep the current behavior

ebolyen: The question actually refers to whether or not `visualizers` receive this object instead of a string for `output_dir`.

jairideout: Oh, no strong feelings there either :)

PR #125, opened 2016-09-05
ENH: added defaults to qiime.core.type.signature classes

BenKaehler:
I have added defaults to the PipelineSignature classes so that they can be exposed in q2cli. Happy to make any changes.

### DIFF
--- a/qiime/core/type/signature.py
+++ b/qiime/core/type/signature.py
@@ -24,7 +24,7 @@ from .visualization import Visualization
 # input semantic types, zero or more parameters, and produce one or more output
 # semantic types or Visualization types.
 class PipelineSignature:
-    def __init__(self, inputs, parameters, outputs):
+    def __init__(self, inputs, parameters, defaults, outputs):
         """
         Parameters
         ----------
@@ -32,18 +32,21 @@ class PipelineSignature:
             Parameter name to tuple of semantic type and view type.
         parameters : dict
             Parameter name to tuple of primitive type and view type.
+        defaults : dict
+            Parameter name to default argument.
         outputs : collections.OrderedDict
             Named output to tuple of QIIME type and view type.
-
         """
         # TODO ensure `inputs`, `parameters`, and `outputs` are OrderedDicts.
         self._assert_valid_inputs(inputs)
         self._assert_valid_parameters(parameters)
+        self._assert_valid_defaults(defaults, parameters)
         self._assert_valid_outputs(outputs)
 
         self.inputs = inputs
         self.parameters = parameters
         self.outputs = outputs
+        self.defaults = defaults
 
     def _assert_valid_inputs(self, inputs):
         if len(inputs) == 0:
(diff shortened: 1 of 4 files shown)


BenKaehler:
Also, thanks @ebolyen for initial advice. The mistakes are mine.

ebolyen:
Thanks @BenKaehler! We'll need to integrate this with construction of the pipelines (which happens somewhere between registration and Action construction). @jairideout just rewrote a lot of that, so he would know offhand where `inspect` is used so that the defaults can be provided.

jairideout:
I'll take a look at this soon, thanks @BenKaehler

jairideout:
@BenKaehler apologies for the delay in reviewing this. It's on my radar and I should have time in the next couple days.

BenKaehler:
@jairideout No worries. I imagine you're pretty busy.

qiime/core/type/signature.py
```diff
@@ -267,14 +282,20 @@ def function_to_signature(cls, function, inputs, parameters, outputs):
...
-        param_types[param_name] = (primitive_type, view_type)
+    defaults = {}
+    signature = inspect.signature(function)
+    for param_name, parameter in signature.parameters.items():
+        view_type = parameter.annotation
+        if param_name in inputs:
+            input_types[param_name] = inputs[param_name], view_type
+        elif param_name in parameters:
+            param_types[param_name] = parameters[param_name], view_type
+        elif param_name not in ('output_dir',):  # output_dir is special
```
jairideout: It'd be nice to not have to special-case here, since we don't know if this function is being called from `MethodSignature` or `VisualizerSignature` (`output_dir` only applies to visualizers). I think using the previous code (two loops) should work, something like:

```python
signature = inspect.signature(function)
input_types = {}
for param_name, semantic_type in inputs.items():
    parameter = signature.parameters[param_name]    

    view_type = parameter.annotation
    input_types[param_name] = (semantic_type, view_type)
    
    if parameter.default is not signature.empty:    
        defaults[param_name] = parameter.default

# Repeat this strategy, looping over `parameters`. Perhaps create a helper function since the code is the same.
```

qiime/sdk/tests/test_method.py
```diff
@@ -40,6 +40,9 @@ def setUp(self):
                 'int1': (qiime.plugin.Int, int),
                 'int2': (qiime.plugin.Int, int)
             },
+            defaults={
+                'int1': 42,
```
jairideout: Comparing signatures for equality ignores `defaults`. Can you add a comparison of `defaults` in `PipelineSignature.__eq__`? These tests will then start failing (and the visualizer tests below) since `'int1'` doesn't have a default value in the registered function (same issue for `'key_label'` below). You can set `defaults={}` for both cases to fix the tests.

qiime/core/type/signature.py
```diff
@@ -73,6 +76,17 @@ def _assert_valid_parameters(self, parameters):
...
                     "expression, not %r" % (param_name, primitive_type))
 
+    def _assert_valid_defaults(self, defaults, parameters):
+        # only parameters can have defaults and the type must match
+        for param_name, default in defaults.items():
+            if param_name not in parameters:
+                raise ValueError("Input %r must not have a default value" %
+                                 param_name)
+
+            if not isinstance(default, parameters[param_name][1]):
```
BenKaehler: Good comment, thanks. I wasn't aware that you could use a semantic type in that way.

BenKaehler:
Thanks @jairideout I've made some changes. Please let me know what you think.

jairideout:
Thanks @BenKaehler, 👍 👍 👍 !

gregcaporaso:
Thanks @BenKaehler!

ebolyen:
This is awesome @BenKaehler, thanks!!! :tada: :tada: :tada:

PR #131, opened 2016-09-07
ENH: Actions return Results object regardless of number of outputs

jairideout:
Now that we have a Results class it is more consistent and easier to reason about if a Results object is always returned from an Action, regardless of the number of outputs. This will also simplify logic in interfaces (e.g. q2cli and qiime-studio).

### DIFF
--- a/qiime/core/callable.py
+++ b/qiime/core/callable.py
@@ -249,15 +249,9 @@ class Callable(metaclass=abc.ABCMeta):
                     "defined in signature: %d != %d" %
                     (len(outputs_tuple), len(self.signature.outputs)))
 
-            # If there is a single output, don't wrap in a Results object to
-            # match how Python handles single return values. Otherwise, wrap in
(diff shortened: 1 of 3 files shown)


jairideout:
@ebolyen, just a ping on this, not sure if you saw I assigned it to you for review. It should be quick

thermokarst:
Thanks @jairideout, this looks good to me! Do we need to update the interfaces ([q2cli](https://github.com/qiime2/q2cli/blob/master/q2cli/commands.py#L153), [studio](https://github.com/qiime2/qiime-studio/blob/master/qiime_studio/api/jobs.py#L139)), too? If so, can you open issues to note that?

jairideout:
Thanks @thermokarst! I created issues for the interfaces:

https://github.com/qiime2/q2cli/issues/58
https://github.com/qiime2/qiime-studio/issues/76

PR #136, opened 2016-09-12
ENH: Archiver.save() filepath validation

jairideout:
Archiver.save() now validates filepaths to provide useful error messages for common mistakes. Added corresponding unit tests.

Also removed FakeArchiver from unit tests in favor of a real Archiver instance.


thermokarst:
RIP `FakeArchiver`. You will be missed.

PR #142, opened 2016-09-15
ENH: Archive version 0.3.0: use archive UUID as zipfile root directory

### LINKED ISSUE #140: make archive root dir the artifact's UUID
The unzipped directory will be the artifact's UUID. Ignore hidden files when loading and locating this directory. Validate this directory matches the UUID in metadata.yml.

jairideout:
Archive version 0.3.0 is not backward-compatible with older versions. The new format uses the archive's UUID as the root directory in the zipfile. Thus, extracting an archive will create a single directory with the archive's UUID as its name.

Previous archive formats determined the root directory name from the filepath the archive was being loaded from or saved to. This was fragile for two reasons: 1) if the file is renamed after saving, it became unloadable (!!!), and 2) if the file was extracted (e.g. using `unzip`), it could potentially mangle an existing directory that happened to have the same name. Now that a UUID is used for the extracted directory, collisions like this should never happen in practice.

Many thanks to @thermokarst for having the idea to use the archive's UUID as its root directory!

Other changes/notes:

- When loading an archive, dotfiles/dirs are ignored so that accidental addition of these files (e.g. user unzips, creates a .DS_Store, and rezips) won't make the archive unloadable.

- Archiver now manages UUID parsing, formatting, validation, and generation. UUID concerns were previously shared between Archiver, Artifact, and Visualization. UUIDs are now assured to be valid version 4 UUIDs, where other types of UUIDs were inadverently allowed before.  These changes to Archiver actually simplified code by removing constructor parameter `archive_filepath`, which was mutually exclusive with `data_initializer`. Now, Archiver requires a `data_initializer`.  If a UUID is provided to the constructor, it will be used (e.g. when loading an archive), otherwise a random version 4 UUID is generated by Archiver. These changes required modifications to Archiver's constructor signature (it's a private class though so shouldn't impact users).

Fixes #140.


qiime/core/archiver.py
```diff
@@ -66,12 +66,68 @@ def peek(cls, filepath):
...
+        if len(roots) == 0:
+            raise ValueError("Archive does not have a visible root directory.")
+        if len(roots) > 1:
+            raise ValueError(
+                "Archive has multiple root directories: %r" % roots)
+
+        root = roots.pop()
+        if not cls._is_uuid4(root):
+            raise ValueError(
+                "Archive root directory %r is not a valid version 4 UUID." %
```
jairideout: Add "name" to error message

qiime/core/archiver.py
```diff
@@ -66,12 +66,68 @@ def peek(cls, filepath):
 
     @classmethod
     def load(cls, filepath):
-        metadata = cls.peek(filepath)
-        return cls(*metadata, archive_filepath=filepath)
+        def extract_data(data_dir):
+            with zipfile.ZipFile(filepath, mode='r') as zf:
+                root_dir = cls._get_root_dir(zf)
+                prefix = os.path.join(root_dir, cls.DATA_DIRNAME, '')
```
jairideout: Add comment describing this path

jairideout: Create class-level variable for zip separator (`/`) since `zf.namelist()` will return `/` paths.

qiime/core/archiver.py
```diff
@@ -66,12 +66,68 @@ def peek(cls, filepath):
...
+
+                for file_ in zf.namelist():
+                    if file_.startswith(prefix) and file_ != prefix:
+                        zf.extract(file_, path=data_dir)
+
+            # This is `data_dir`/<archive uuid>/data/
+            extracted_data_dir = os.path.join(data_dir, prefix)
```
jairideout: `prefix` needs to use os-specific separator

qiime/core/archiver.py
```diff
@@ -66,12 +66,68 @@ def peek(cls, filepath):
...
+                   uuid_=uuid_)
 
     @classmethod
-    def _get_root_dir(cls, filepath):
-        return os.path.splitext(os.path.basename(filepath))[0]
+    def _get_root_dir(cls, zf):
+        roots = set()
+        for relpath in zf.namelist():
+            if not relpath.startswith('.'):
+                root = relpath.split(os.sep, maxsplit=1)[0]
```
jairideout: split on zipfile sep

qiime/core/testing/util.py
```diff
@@ -18,3 +21,58 @@ def get_dummy_plugin():
...
+            determining the root directory dynamically.
+        expected : set of str
+            Set of expected archive members extracted to `extract_dir`. Stored
+            as paths relative to `root_dir`.
+
+        """
+        expected = {os.path.join(extract_dir, root_dir, member)
+                    for member in expected}
+
+        for expected_path in expected:
```
jairideout: Make this a set comparison similar to above method

qiime/core/tests/test_archiver.py
```diff
@@ -167,23 +167,253 @@ def data_initializer(data_dir):
...
+            'VERSION',
+            'metadata.yaml',
+            'README.md',
+            'data/ints.txt'
+        }
+
+        self.assertArchiveMembers(fp, root_dir, expected)
+
+    def test_save_archive_members(self):
+        fp = os.path.join(self.temp_dir.name, 'archive.qza')
```
jairideout: update all .qza filepaths to .zip

qiime/core/tests/test_archiver.py
```diff
@@ -167,23 +167,253 @@ def data_initializer(data_dir):
...
+        archiver = Archiver.load(fp)
+
+        self.assertEqual(archiver.uuid, self.archiver.uuid)
+        self.assertEqual(archiver.type, IntSequence1)
+        self.assertEqual(archiver.format, 'IntSequenceDirectoryFormat')
+        self.assertIsNone(archiver.provenance)
+        self.assertEqual({e for (e, _) in archiver.get_data_paths()},
+                         {'ints.txt'})
+
+    def test_load_ignores_directory_members(self):
```
jairideout: Remove "external" .qza/.qzv files!

jairideout:
@ebolyen @thermokarst, this is ready for a final review. Thanks!

thermokarst:
This looks great! Thanks so much for wranglin' UUIDs all day!

jairideout:
Thanks @thermokarst!!

PR #155, opened 2016-09-27
FIX: Support semantic type registration to multiple formats in TestPluginBase harness

jakereps:
Found while writing tests for `q2_types.feature_table` subpackage.

It has one type registered to two formats, so with `assertSemanticTypeRegisteredToFormat` grabbing the first result it found it led to random occurences of passing and failing based on how they came out of the dict.

```python
plugin.register_semantic_type_to_format(
    FeatureTable[Frequency | RelativeFrequency | PresenceAbsence],
    artifact_format=BIOMV210DirFmt
)

plugin.register_semantic_type_to_format(
    FeatureTable[Frequency | RelativeFrequency | PresenceAbsence],
    artifact_format=BIOMV100DirFmt
)
```

### DIFF
--- a/qiime/plugin/testing.py
+++ b/qiime/plugin/testing.py
@@ -72,7 +72,8 @@ class TestPluginBase(unittest.TestCase):
     def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
         obs_format = None
         for type_format_record in self.plugin.type_formats:
-            if type_format_record.type_expression == semantic_type:
+            if (type_format_record.type_expression == semantic_type and
+                    exp_format == type_format_record.format):
                 obs_format = type_format_record.format
                 break
 


jairideout:
Thanks @jakereps, we're closing this since this type of assertion shouldn't be necessary as we don't plan on supporting multiple format registrations to a semantic type. I think we should just remove the duplicate registration of BIOM v1 in q2-types, which should solve the failure you were getting. We already have an issue open on the framework to explicitly disallow duplicate registrations, so it will be impossible for this case to happen in the future. Would you be willing to issue a PR to q2-types removing the duplicate BIOM registration?

jakereps:
Is it not actually a duplicate registration though? One is V100 and one is V210, so won't removing the V100 registration also remove compatibility with the JSON biom format and only leave the HDF5?

jairideout:
The registration associates a semantic type with an output format, so the FeatureTable types will always output data in BIOM v2, but support will remain for reading BIOM v1 files since those transformers still exist.

PR #158, opened 2016-09-27
ENH: adds support for None as default value

### LINKED ISSUE #144: support default parameter values of None
Discuss with @ebolyen about strategies. This is necessary before teaching workshops (e.g. adding explicit time axis support to Emperor plugin).

gregcaporaso:
fixes #144 

I tested this locally by modifying an existing method to have a default parameter of `None`.

### DIFF
--- a/qiime/core/type/signature.py
+++ b/qiime/core/type/signature.py
@@ -77,15 +77,16 @@ class PipelineSignature:
                     "expression, not %r" % (param_name, primitive_type))
 
     def _assert_valid_defaults(self, defaults, parameters):
-        # only parameters can have defaults and the type must match
+        # only parameters can have defaults and the type must match or be None
         for param_name, default in defaults.items():
             if param_name not in parameters:
                 raise ValueError("Input %r must not have a default value" %
                                  param_name)
 
-            if default not in parameters[param_name][0]:
+            if default is not None and \
+               default not in parameters[param_name][0]:
                 raise TypeError("Default value for parameter %r is not of "
-                                "semantic QIIME type %r" %
+                                "semantic QIIME type %r or None." %
                                 (param_name, parameters[param_name][0]))
 
     def _assert_valid_outputs(self, outputs):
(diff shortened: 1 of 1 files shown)


jairideout:
Do you even flake8, bro? 💔

gregcaporaso:
addressed @ebolyen's comment.

qiime/core/type/signature.py
```diff
@@ -113,12 +114,20 @@ def decode_parameters(self, **kwargs):
         return params
 
     def check_types(self, **kwargs):
-        for key, (type_, _) in itertools.chain(self.inputs.items(),
-                                               self.parameters.items()):
+        for key, (type_, _) in self.inputs.items():
+            if kwargs[key] is None:
```
wasade: is it assured that `key` will exist in `kwargs` or should this be a `get`?

jairideout: I think `key` will be in `kwargs` because the user would have either 1) had to supply a parameter, e.g. if it's a required parameter, or 2) they didn't supply the parameter but it has a default value. Note that the previous code assumed that `key` is in `kwargs` so a change probably isn't necessary in this PR.

qiime/core/type/signature.py
```diff
@@ -113,12 +114,20 @@ def decode_parameters(self, **kwargs):
...
+                raise TypeError("Argument to input %r is None, but only "
```
jairideout: Why is this check necessary? `None` isn't a subtype of a semantic type, just like `42` isn't. I don't think we need to special-case `None` here because it would fail below on line 121.

qiime/core/type/signature.py
```diff
@@ -113,12 +114,20 @@ def decode_parameters(self, **kwargs):
...
+                                "parameters can have values of None." % key)
             if kwargs[key] not in type_:
                 raise TypeError("Argument to parameter %r is not a subtype of"
                                 " %r." % (key, type_))
 
+        for key, (type_, _) in self.parameters.items():
+            if kwargs[key] not in type_:
+                if not (self.defaults[key] is None and kwargs[key] is None):
```
jairideout: I think `self.defaults[key]` could raise a `KeyError` if the parameter doesn't have a default.

qiime/core/type/signature.py
ebolyen: It isn't necessary, but I don't mind having it either.

qiime/core/type/signature.py
ebolyen: I'm not super familiar with how defaults work, I take it it isn't a 1:1 map on the signature but rather some subset of parameters?

qiime/core/type/signature.py
jairideout: Ok, I would prefer removing it to avoid special-casing `None` where it isn't necessary.

qiime/core/type/signature.py
jairideout: Yep it's a subset

gregcaporaso:
These comments should be addressed with these changes.

gregcaporaso:
Thanks!

PR #159, opened 2016-09-28
ENH: Add explicit format registration to Plugin objects

jakereps:
A "bug" was discovered while removing the duplicate BIOMv1 directory type registration from `q2-types`. The framework loses the ability to know this format still exists, as it was previously scraping transformer annotations, and semantic type to format registrations.

This PR adds explicit format registration to plugin objects, and makes them accessible through the `PluginManager`, a nice side effect is that this greatly reduces the runtime of `qiime.sdk.util.parse_format`.

### DIFF
--- a/qiime/core/testing/plugin.py
+++ b/qiime/core/testing/plugin.py
@@ -12,9 +12,12 @@ import qiime
 import qiime.plugin
 
 from .format import (
+    IntSequenceFormat,
+    MappingFormat,
+    SingleIntFormat,
     IntSequenceDirectoryFormat,
     MappingDirectoryFormat,
-    FourIntsDirectoryFormat,
+    FourIntsDirectoryFormat
 )
 
 from .type import IntSequence1, IntSequence2, Mapping, FourInts
@@ -37,6 +40,14 @@ dummy_plugin.register_semantic_type(IntSequence1)
 dummy_plugin.register_semantic_type(IntSequence2)
 dummy_plugin.register_semantic_type(Mapping)
 dummy_plugin.register_semantic_type(FourInts)
+
+# Register formats
+dummy_plugin.register_format(IntSequenceFormat)
+dummy_plugin.register_format(MappingFormat)
+dummy_plugin.register_format(SingleIntFormat)
+dummy_plugin.register_format(IntSequenceDirectoryFormat)
+dummy_plugin.register_format(MappingDirectoryFormat)
+dummy_plugin.register_format(FourIntsDirectoryFormat)
 
 
 dummy_plugin.register_semantic_type_to_format(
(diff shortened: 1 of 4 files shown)


qiime/core/testing/plugin.py
```diff
@@ -38,6 +41,14 @@
 dummy_plugin.register_semantic_type(Mapping)
 dummy_plugin.register_semantic_type(FourInts)
 
+# Register formats
+dummy_plugin.register_format(IntSequenceFormat)
```
jakereps: Can both `register_format` and `register_semantic_type` be combined into plurals?

```python
plugin.register_semantic_type(FeatureTable)
plugin.register_semantic_type(Frequency)
plugin.register_semantic_type(RelativeFrequency)
plugin.register_semantic_type(PresenceAbsence)
```

to

```python
plugin.register_semantic_types(FeatureTable, Frequency, RelativeFrequency,
                               PresenceAbsence)
```

jairideout:
Thanks @jakereps! @gregcaporaso, @ebolyen, and I chatted and think it would be great to have the plural forms (`register_semantic_types`, `register_formats`) to reduce boilerplate. The singular forms would be removed. It seems like a worthwhile change to make now since you're already going to update the plugins to use the new format registration API, so the plural semantic type registration updates could happen in those same PRs. Are you good with this plan?

qiime/plugin/plugin.py
```diff
@@ -65,6 +68,15 @@ def actions(self):
         actions.update(self.visualizers)
         return types.MappingProxyType(actions)
 
+    def register_format(self, format):
+        if not issubclass(format, FormatBase):
+            raise TypeError("%r is not a valid format." % format)
+        if format.__name__ in self.formats.keys():
```
jairideout: Minor, `.keys()` call is unnecessary

qiime/sdk/plugin_manager.py
```diff
@@ -65,6 +66,7 @@ def _integrate_plugin(self, plugin):
                                  % transformer_record)
             self.transformers[input][output] = transformer_record
 
+        self.formats.update(plugin.formats)
```
jairideout: Instead of doing an `update`, can you check for duplicate format registrations here? That seems more likely to happen than a duplicate registration within a plugin, though that check you have is great to keep too.

jairideout:
I forgot to mention in my review above that we want the plural registration forms instead of singular. @jakereps let me know when you're ready for another passthrough.

jakereps:
Thanks @jairideout! Everything should be addressed, and pluralization completed. Updated the `q2-types` PR to match the plural registration, and have a demux and classifier PR ready to go, will open them after this class!

jairideout:
Awesome, I really like the plural registrations 👍  Thanks @jakereps!

ebolyen:
:+1:

PR #165, opened 2016-10-07
REF: Refactored Archiver logic

ebolyen:
Archiver has become its own subpackage `archive`. There are now 3 distinct
components: `Archiver`, `_Archive` impl, and `ArchiveFormat`.

The `Archiver` is responsible for coordinating, it maintains a list of all
current archive formats and archive implementations as well as what the current
version of both of those are.

An `_Archive` implementation is responsible for the high level structure of an
archive (the uuid root dir and VERSION file). Its job is to recognize if a
given file is valid archive and if so mount/extract it. Additionally it
provides a `setup` which initializes an existing directory.

An `ArchiveFormat` specifies the rest of the layout of an archive, this
includes what the data directory is called, metadata, etc...

tldr; Archiver gets ZipArchive to convert an archive to the filesystem and then
gets ArchiveFormat to read the details.

Additionally it is now possible to `peek` at an archive even if you do not have
the correct plugins installed to `load` an archive.

Similarly you can `extract` an archive even if you do not have the latest
ArchiveFormat.


qiime/core/archive/archiver.py
```diff
@@ -0,0 +1,341 @@
...
+                "Archive root directory name %r is not a valid version 4 "
+                "UUID." % uuid)
+        return uuid
+
+    def _get_versions(self):
+        try:
+            with self.open(self.VERSION_FILE) as fh:
+                header, version_line, framework_version_line, eof = \
+                    fh.read().split('\n')
+            assert(header.strip() == 'QIIME 2')
```
ebolyen: switch to exception so `-O` flag works

qiime/core/archive/archiver.py
```diff
@@ -0,0 +1,341 @@
...
+    def mount(self, filepath):
+        # TODO: use FUSE/MacFUSE/Dokany bindings (many Python bindings are
+        # outdated, we may need to take up maintenance/fork)
+        root = self.extract(filepath)
+        return ArchiveRecord(root, root / self.VERSION_FILE,
+                             self.uuid, self.version, self.framework_version)
+
+    def extract(self, filepath):
+        with zipfile.ZipFile(str(self.path), mode='r') as zf:
+            zf.extractall(str(filepath))
```
ebolyen: make this safe

jairideout:
Thanks @ebolyen, nice job on the refactoring! 🎉

thermokarst:
🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 🌮 !!!!!!!!!

PR #170, opened 2016-10-19
ENH: Signature now matches order of function sig.

ebolyen:
Previously the `inputs` and `parameters` property of `Signature` objects
created by the `.from_function` alternate constructor did not respect the order
of the original function signature. Now these properties are always
`collections.OrderedDict` instances.


ebolyen:
This is a quick one, I need this functionality in order to retain the order in provenance.

jairideout:
Thanks @ebolyen!

PR #172, opened 2016-10-24
ENH: Add Result.export_data to sdk

jakereps:
Adds `Result.export_data` to `qiime.sdk` for exporting the data directories to the user's filesystem.

### DIFF
--- a/qiime/sdk/result.py
+++ b/qiime/sdk/result.py
@@ -119,6 +119,13 @@ class Result:
     def __ne__(self, other):
         return not (self == other)
 
+    def export_data(self, output_dir):
+        distutils.dir_util.copy_tree(
+            str(self._archiver.data_dir), str(output_dir))
+        # Return None for now, although future implementations that include
+        # format tranformations may return the invoked transformers
+        return None
+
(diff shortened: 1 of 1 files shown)


thermokarst:
@jakereps, after discussing with @gregcaporaso, @ebolyen, and @jairideout the lack of transformers to support the `format` argument , we think for now we should just drop that functionality (at least until we add in some of the missing transformers). How about something like this:

```python
def export_data(self, output_dir):
    from shutil import copytree
    copytree(str(self._archiver.data_dir), output_dir)
```

jakereps:
Sounds good, yeah it took hacking another transformer into `q2_types` to even test if it was working. Any preference on using the already imported `copy_tree`  from [here](https://github.com/qiime2/qiime2/pull/172/files#diff-86a19f23a3f1b6170b33fcc238c64704L247) over importing `shutil` as well?

thermokarst:
No preference here --- if it works then it is one less import!

jakereps:
Thanks @thermokarst and @jairideout! Should be good after tests 👍 !

PR #174, opened 2016-10-26
FIX: Directory Formats fail to validate from globbed hidden files

jakereps:
Trying to import using a `source_format` as a Directory Format is failing due to the glob grabbing hidden `.DS_Store` files and failing to validate.

``` bash
ValueError: Unrecognized file (InPath('/var/folders/t1/qdtgl9t964s0s_8yfsfx9d_c0
000gq/T/qiime2-archive-_4ak2iap/f00c26ac-4cea-4620-9f35-249024bd241c/data/.DS_St
ore')) for 'BIOMV100DirFmt'.
```

### DIFF
--- a/qiime/plugin/model/directory_format.py
+++ b/qiime/plugin/model/directory_format.py
@@ -162,7 +162,8 @@ class DirectoryFormat(FormatBase, metaclass=_DirectoryMeta):
         if not self.path.is_dir():
             raise ValueError("%r is not a directory." % self.path)
         collected_paths = {p: None for p in self.path.glob('**/*')
-                           if p.is_file()}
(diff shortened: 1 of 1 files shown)


qiime/plugin/model/directory_format.py
```diff
@@ -162,7 +162,8 @@ def validate(self):
         if not self.path.is_dir():
             raise ValueError("%r is not a directory." % self.path)
         collected_paths = {p: None for p in self.path.glob('**/*')
-                           if p.is_file()}
+                           if not p.name.startswith('.') and
+                           p.is_file()}
```
ebolyen: :+1:

PR #182, opened 2016-11-02
ENH: more informative error message

wasade:
Provides additional context, whereas the existing behavior was the default `KeyError` which did not specify the function being evaluated.

### DIFF
--- a/qiime/core/type/signature.py
+++ b/qiime/core/type/signature.py
@@ -315,8 +315,8 @@ def function_to_signature(cls, function, inputs, parameters, outputs):
     output_types = collections.OrderedDict()
     for output_name, semantic_type in outputs.items():
         view_type = output_view_types.get(output_name)
-
-        if view_type is None:
+        
+        if output_name != "visualization" and view_type is None:
             raise TypeError("Function %r does not describe %r in its return "
                             "annotation." % (function.__name__, output_name))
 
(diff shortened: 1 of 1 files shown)


wasade:
I'm a bit confused by the errors this triggers, is it due to a `TypeError` being raised instead of a `KeyError`?

jairideout:
Thanks for the fix! All QIIME 2 repos are frozen for the rest of today so we can cut the 2.0.6 train release. We'll have time over the next few days to review your PR.

wasade:
np, thanks!

thermokarst:
Hi @wasade, just a quick follow-up. Looks like there were some `flake8` issues in 17ddec2. If you are interested in continuing with this PR, we also recently made some changes to `PipelineSignature` that impact this PR. The utility function `function_to_signature` was [moved to](https://github.com/qiime2/qiime2/blob/master/qiime/core/type/signature.py#L53) `PipelineSignature._parse_annotations`. You should be able to just move your changes up into that method.

wasade:
Thanks. Yes, I'll resolve but might be another day.

thermokarst:
Hi @wasade - just chatted with @jairideout about this one. He did a pretty big refactor on this bit of the framework. Before proceeding with this PR, can you you let us know if your original problem is still present? There is a [better exception in place](https://github.com/qiime2/qiime2/blob/master/qiime/sdk/action.py#L267-L269) for methods that don't return the type declared in the annotation. If this **is** still a problem for you, can you provide a traceback and a MWE? Thanks!

wasade:
Will do.

thermokarst:
Hi @wasade --- I'm going to close this PR, @jairideout and I discussed, and we believe that recent changes to the framework (since the last release) address these concerns. If we are mistaken, feel free to open a new PR. Thanks!

PR #189, opened 2016-11-22
ENH/REF/MAINT: remove support for markdown methods

jairideout:
Removes support for registering methods defined in markdown files. This feature was only used in `q2-diversity core-metrics` method, and the idea represented by markdown methods will be replaced with pipeline support in a future release. Removing this feature also makes the codebase substantially simpler and easier to maintain.

User-facing changes:

- Removed `qiime.plugin.Plugin.methods.register_markdown`
- Methods are now type `qiime.sdk.Method` and visualizers are now type `qiime.sdk.Visualizer`. Both inherit from `qiime.sdk.Action`. Previously, methods and visualizers were type `qiime.sdk.Action` but `Action` is now an abstract base class.
- `Results`, the object returned by `Actions`, is now publicly importable at `qiime.sdk.Results`.
- python-frontmatter, jupyter, and ipymd are no longer dependencies.

Internal changes:

- Removed `qiime.core.callable.Callable` and its subclasses, merging their functionality with `qiime.sdk.Action/Method/Visualizer`.
- `qiime.sdk.Action` is now an ABC that new classes `Method` and `Visualizer` inherit from.
- Removed various `from_function` alternate constructors in favor of using a single constructor.


gregcaporaso:
🎆

PR #193, opened 2016-12-02
ENH: Add Transformer Test Helper to TestPluginBase

jakereps:
Discovered a pattern while making `q2_types` tests, and this helper cut over 40 lines of transformer test boilerplate from a single subpackage.

### DIFF
--- a/qiime/plugin/testing.py
+++ b/qiime/plugin/testing.py
@@ -91,23 +91,31 @@ class TestPluginBase(unittest.TestCase):
             "Expected semantic type %r to be registered to format %r, not %r."
             % (semantic_type, exp_format, obs_format))
 
-    def transform_format(self, source_format, target, filenames):
+    def transform_format(self, source_format, target, **kwargs):
         # Guard any non-QIIME2 Format sources from being tested
         self.assertTrue(issubclass(source_format, FormatBase))
 
         transformer = self.get_transformer(source_format, target)
 
-        for filename in filenames:
-            filepath = self.get_data_path(filename)
-            shutil.copy(filepath, self.temp_dir.name)
-        input = source_format(self.temp_dir.name, mode='r')
+        source_path = None
+        if 'filename' in kwargs:
+            source_path = kwargs['filename']
+        elif 'filenames' in kwargs:
+            source_path = self.temp_dir.name
+            for filename in kwargs['filenames']:
+                filepath = self.get_data_path(filename)
+                shutil.copy(filepath, source_path)
+        else:
+            self.fail("The helper method `transform_format` requires either "
+                      "a filename or a sequence of filenames supplied as their"
+                      " respective keyword arguments.")
+        input = source_format(source_path, mode='r')
 
         obs = transformer(input)
 
-        if issubclass(type(target), DirectoryFormat):
-            classes = (pathlib.Path, str, DirectoryFormat)
-            valid = (issubclass(type(obs), x) for x in classes).any()
-            self.assertTrue(valid)
+        if issubclass(target, DirectoryFormat):
+            self.assertTrue(type(obs) in (pathlib.Path, str) or
+                            issubclass(type(obs), DirectoryFormat))
         else:
             self.assertIsInstance(obs, target)
 
--- a/qiime/plugin/testing.py
+++ b/qiime/plugin/testing.py
@@ -14,7 +14,6 @@ import pathlib
 
 import qiime
 
-from qiime.plugin.model import DirectoryFormat
 from qiime.plugin.model.base import FormatBase
 
 
@@ -91,31 +90,32 @@ class TestPluginBase(unittest.TestCase):
             "Expected semantic type %r to be registered to format %r, not %r."
             % (semantic_type, exp_format, obs_format))
 
-    def transform_format(self, source_format, target, **kwargs):
+    def transform_format(self, source_format, target, filename=None,
+                         filenames=None):
         # Guard any non-QIIME2 Format sources from being tested
         self.assertTrue(issubclass(source_format, FormatBase))
 
-        transformer = self.get_transformer(source_format, target)
+        # Guard against invalid filename(s) usage
+        if filename is not None and filenames is not None:
+            raise ValueError("Cannot use both `filename` and `filenames` at "
+                             "the same time.")
 
+        # Handle format initialization
         source_path = None
-        if 'filename' in kwargs:
-            source_path = kwargs['filename']
-        elif 'filenames' in kwargs:
+        if filename:
+            source_path = self.get_data_path(filename)
+        elif filenames:
             source_path = self.temp_dir.name
-            for filename in kwargs['filenames']:
+            for filename in filenames:
                 filepath = self.get_data_path(filename)
                 shutil.copy(filepath, source_path)
-        else:
-            self.fail("The helper method `transform_format` requires either "
-                      "a filename or a sequence of filenames supplied as their"
-                      " respective keyword arguments.")
         input = source_format(source_path, mode='r')
 
+        transformer = self.get_transformer(source_format, target)
         obs = transformer(input)
 
-        if issubclass(target, DirectoryFormat):
-            self.assertTrue(type(obs) in (pathlib.Path, str) or
-                            issubclass(type(obs), DirectoryFormat))
+        if issubclass(target, FormatBase):
+            self.assertIsInstance(obs, (type(pathlib.Path()), str, target))
         else:
             self.assertIsInstance(obs, target)
 
--- a/qiime/plugin/testing.py
+++ b/qiime/plugin/testing.py
@@ -9,6 +9,8 @@
 import pkg_resources
 import tempfile
 import unittest
+import shutil
+
 import qiime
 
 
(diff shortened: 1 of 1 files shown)


qiime/plugin/testing.py
```diff
@@ -84,3 +86,17 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
             obs_format, exp_format,
             "Expected semantic type %r to be registered to format %r, not %r."
             % (semantic_type, exp_format, obs_format))
+
+    def transformer_helper(self, source, target, filenames):
```
thermokarst: `transform_format(self, source_format, target, filenames)`

qiime/plugin/testing.py
```diff
@@ -84,3 +86,17 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
...
+        transformer = self.get_transformer(source, target)
```
thermokarst: First off, check that `source_format` is the real deal.

qiime/plugin/testing.py
```diff
@@ -84,3 +86,17 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
...
+
+        for filename in filenames:
+            filepath = self.get_data_path(filename)
+            shutil.copy(filepath, self.temp_dir.name)
+        input = source(self.temp_dir.name, mode='r')
+
+        obs = transformer(input)
+
+        self.assertIsInstance(obs, target)
```
thermokarst: if `target` is a (`pathlib.Path`, `str`, or a subclass of `DirectoryFormat`), check that `obs` is an instance of `pathlib.Path`, `str`, or a subclass of `DirectoryFormat`. If not, just `assertIsInstance(obs, target)`.

qiime/plugin/testing.py
```diff
@@ -84,3 +90,25 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
             obs_format, exp_format,
             "Expected semantic type %r to be registered to format %r, not %r."
             % (semantic_type, exp_format, obs_format))
+
+    def transform_format(self, source_format, target, filenames):
```
thermokarst: Since we are supporting both `FileFormat` and `DirectoryFormat`, I think we need to support providing one `filename` or a list of `filenames`. How about we make those `kwargs`, and put an exclusivity guard in (one or the other). Motivation for this is described in a few of the following comments.

qiime/plugin/testing.py
```diff
@@ -84,3 +90,25 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
...
+        # Guard any non-QIIME2 Format sources from being tested
+        self.assertTrue(issubclass(source_format, FormatBase))
+
+        transformer = self.get_transformer(source_format, target)
+
+        for filename in filenames:
```
thermokarst: This existing loop is applicable for `DirectoryFormat`, but needs to be updated to accommodate `FileFormat`

qiime/plugin/testing.py
```diff
@@ -84,3 +90,25 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
...
+            filepath = self.get_data_path(filename)
+            shutil.copy(filepath, self.temp_dir.name)
+        input = source_format(self.temp_dir.name, mode='r')
```
thermokarst: `self.temp_dir.name` works for `DirectoryFormat`, but we need to provide the path to the actual file for `FileFormat`.

@ebolyen: I had a memory that creating a `FileFormat` with the path to the file's parent directory worked, but I must have been mistaken. We have this behavior for `SingleFileDirectoryFormat` (specify either the surrounding dir or the single file itself) --- does it make sense to support this type of construction for `FileFormat`, too?

qiime/plugin/testing.py
```diff
@@ -84,3 +90,25 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
...
+
+        obs = transformer(input)
+
+        if issubclass(type(target), DirectoryFormat):
+            classes = (pathlib.Path, str, DirectoryFormat)
+            valid = (issubclass(type(obs), x) for x in classes).any()
```
thermokarst: This throws an `AttributeError`, should it be

```python
valid = any(issubclass(type(obs), x) for x in classes)
```
?

qiime/plugin/testing.py
```diff
@@ -84,3 +90,33 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
...
+                shutil.copy(filepath, source_path)
+        else:
+            self.fail("The helper method `transform_format` requires either "
+                      "a filename or a sequence of filenames supplied as their"
+                      " respective keyword arguments.")
+        input = source_format(source_path, mode='r')
+
+        obs = transformer(input)
+
+        if issubclass(target, DirectoryFormat):
```
thermokarst: We should probably test for `FormatBase` here, instead of `DirectoryFormat`, since we can be transforming to to `FileFormat`, which has the same `pathlib.Path`, `str`, etc properties as `DirectoryFormat`

qiime/plugin/testing.py
```diff
@@ -84,3 +90,33 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
             obs_format, exp_format,
             "Expected semantic type %r to be registered to format %r, not %r."
             % (semantic_type, exp_format, obs_format))
+
+    def transform_format(self, source_format, target, **kwargs):
```
thermokarst: Can you replace `**kwargs` with `filename=None, filenames=None` here? That way the function signature is explicit about what we are up to here.

qiime/plugin/testing.py
```diff
@@ -84,3 +90,33 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
...
+        source_path = None
+        if 'filename' in kwargs:
+            source_path = kwargs['filename']
+        elif 'filenames' in kwargs:
+            source_path = self.temp_dir.name
+            for filename in kwargs['filenames']:
+                filepath = self.get_data_path(filename)
+                shutil.copy(filepath, source_path)
+        else:
+            self.fail("The helper method `transform_format` requires either "
```
thermokarst: I'm not sure if failing is what we want to do here, since this is a developer error, not a test failure. With the signature changes requested above you could do something like:

```python
if filename is not None and filenames is not None:
    raise ValueError("Cannot use both `filename` and `filenames` at the same time.")
```

qiime/plugin/testing.py
```diff
@@ -84,3 +90,33 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
...
+                      "a filename or a sequence of filenames supplied as their"
+                      " respective keyword arguments.")
+        input = source_format(source_path, mode='r')
+
+        obs = transformer(input)
+
+        if issubclass(target, DirectoryFormat):
+            self.assertTrue(type(obs) in (pathlib.Path, str) or
+                            issubclass(type(obs), DirectoryFormat))
```
thermokarst: Besides a transformer returning an actual `pathlib.Path`, it can also return subclasses of `pathlib.Path`

qiime/plugin/testing.py
jakereps: Now that makes more sense, I got thrown off by review having the `kwargs` being in code highlighting over just making them "keywords". I'll get these all tweaked today, my head was apparently out of it for a few of these.

thermokarst: Oops! That was a typo on my part, totally meant keywords! Sorry! 🌮 🌮 🌮

qiime/plugin/testing.py
jakereps: Ah, got it. I misheard that it was only `DirectoryFormat` that had the wildcard returns.

jakereps:
Thanks @thermokarst! Tested on a few `DistanceMatrix` format transformer tests and it passed.

qiime/plugin/testing.py
```diff
@@ -84,3 +89,34 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
             obs_format, exp_format,
             "Expected semantic type %r to be registered to format %r, not %r."
             % (semantic_type, exp_format, obs_format))
+
+    def transform_format(self, source_format, target, filename=None,
+                         filenames=None):
+        # Guard any non-QIIME2 Format sources from being tested
+        self.assertTrue(issubclass(source_format, FormatBase))
```
thermokarst: Sorry, missed this earlier: for the same reasons why we added the `ValueError` below, I think i makes sense to raise a `TypeError` here, instead of asserting the type. If a developer tries to use the wrong format with this helper, it shouldn't be a test failure, it should be a flat-out error. We could say something like  `source_format must be a subclass of FormatBase`, instead of listing out the three possible subclasses.

qiime/plugin/testing.py
```diff
@@ -84,3 +89,34 @@ def assertSemanticTypeRegisteredToFormat(self, semantic_type, exp_format):
...
+            for filename in filenames:
+                filepath = self.get_data_path(filename)
+                shutil.copy(filepath, source_path)
+        input = source_format(source_path, mode='r')
+
+        transformer = self.get_transformer(source_format, target)
+        obs = transformer(input)
+
+        if issubclass(target, FormatBase):
+            self.assertIsInstance(obs, (type(pathlib.Path()), str, target))
```
thermokarst: :heart:

jakereps: ⌨️ Credit: @ebolyen

jakereps:
🎉 !

PR #194, opened 2016-12-07
ENH: new Signature API

ebolyen:
The `.defaults` property has been replaced by storing a `ParameterSpec` for
every public parameter. `inputs`, `paramters`, and `outputs` are still present
but now map a parameter name to a `ParameterSpec`.

The `ParameterSpec` has three properties: `qiime_type`, `view_type`, and
`default`. Additionally it provides `has_<property>` helper methods to
determine if any of the prior properties exist in a given `ParameterSpec`
instance.


ebolyen:
This is going to break *everything*. Working on corresponding interfaces now.

ebolyen:
q2cli: https://github.com/qiime2/q2cli/pull/106

jairideout:
flake8 failures

qiime/core/type/signature.py
```diff
@@ -9,18 +9,74 @@
...
+
+
+class _NoValue:
+    def __repr__(self):
+        return "NOVALUE"
+
+
+class ParameterSpec(ImmutableBase):
+    def __init__(self, **kwargs):
+        self.NOVALUE = _NoValue()
```
jairideout: Any reason this isn't a module-level sentinel?

qiime/core/type/signature.py
```diff
@@ -9,18 +9,74 @@
...
+        self.qiime_type = kwargs.pop('qiime_type', self.NOVALUE)
```
jairideout: Instead of using kwargs, is it possible to make the constructor explicitly accept these three parameters, which all have a default value of the `NOVALUE` sentinel?

ebolyen: yeah if I make it non-instance this works out better

qiime/core/type/signature.py
```diff
@@ -9,18 +9,74 @@
...
 from .visualization import Visualization
+from ..util import ImmutableBase
+
+
+class _NoValue:
+    def __repr__(self):
+        return "NOVALUE"
+
+
+class ParameterSpec(ImmutableBase):
```
ebolyen: Didn't know about the "default" ability, but it also seems a little unintuitive. I'd rather leave this a class I think.

jairideout: Sounds good!

qiime/core/type/signature.py
```diff
@@ -9,18 +9,74 @@
...
+        return self.default is not self.NOVALUE
+
+    def __repr__(self):
+        return ("ParameterSpec(qiime_type=%r, view_type=%r, default=%r)"
+                % (self.qiime_type, self.view_type, self.default))
+
+    def __eq__(self, other):
+        if self.has_qiime_type() != other.has_qiime_type():
+            return False
+        if self.has_qiime_type() and not self.qiime_type.equals(
```
jairideout: Can this `__eq__` be simplified to so that `==` is used to compare each property? I think this is possible once a module-level sentinel is used, but I'm not sure what the difference is between `__eq__` and `equals` on the QIIME types.

```python
return (
    self.qiime_type == other.qiime_type and
    self.view_type == other.view_type and
    self.default == other.default
)
```

qiime/core/type/signature.py
```diff
@@ -9,18 +9,74 @@
...
+
+    def __ne__(self, other):
+        return not (self == other)
 
 
 # Note: Pipeline doesn't exist yet but it is expected to accept one or more
 # input semantic types, zero or more parameters, and produce one or more output
 # semantic types or Visualization types.
 class PipelineSignature:
+    provided_args = ('ctx',)
```
jairideout: Rename to `builtin_args`, check that builtins come before `inputs` in `_parse_signature`

jairideout:
👍, thanks @ebolyen!!

PR #196, opened 2016-12-12
ENH: Add help_text support to action signatures

jakereps:
Allow optional help text strings to be associated with action parameters.

```python
In [1]: from qiime.plugins.test_plugin.methods import blocking_test

In [2]: blocking_test.signature
Out[2]: 
inputs:
    table: ParameterSpec(qiime_type=FeatureTable[Frequency], view_type=<class 'biom.table.Table'>, default=NOVALUE, help_text=NOVALUE)
parameters:
    timeout: ParameterSpec(qiime_type=Int, view_type=<class 'int'>, default=NOVALUE, help_text='The amount of time to wait before continuing execution of the method.')
outputs:
    rarefied_table: ParameterSpec(qiime_type=FeatureTable[Frequency], view_type=<class 'biom.table.Table'>, default=NOVALUE, help_text=NOVALUE)


```

### DIFF
--- a/qiime/core/type/signature.py
+++ b/qiime/core/type/signature.py
@@ -148,9 +148,7 @@ class PipelineSignature:
             default = ParameterSpec.NOVALUE
             if parameter.default is not parameter.empty:
                 default = parameter.default
-            help_text = help_texts.get(name, ParameterSpec.NOVALUE)
-            if help_text is not ParameterSpec.NOVALUE:
-                help_texts.pop(name)
+            help_text = help_texts.pop(name, ParameterSpec.NOVALUE)
 
             if name in inputs:
                 if in_parameter_section:
@@ -184,24 +182,20 @@ class PipelineSignature:
                                 (len(outputs), len(output_views)))
 
             for (name, qiime_type), view_type in zip(outputs, output_views):
-                help_text = help_texts.get(name, ParameterSpec.NOVALUE)
-                if help_text is not ParameterSpec.NOVALUE:
-                    help_texts.pop(name)
+                help_text = help_texts.pop(name, ParameterSpec.NOVALUE)
                 annotated_outputs[name] = ParameterSpec(qiime_type=qiime_type,
                                                         view_type=view_type,
                                                         help_text=help_text)
         else:
             for name, qiime_type in outputs:
-                help_text = help_texts.get(name, ParameterSpec.NOVALUE)
-                if help_text is not ParameterSpec.NOVALUE:
-                    help_texts.pop(name)
+                help_text = help_texts.pop(name, ParameterSpec.NOVALUE)
                 annotated_outputs[name] = ParameterSpec(qiime_type=qiime_type,
                                                         help_text=help_text)
 
         # we should have popped the help_texts empty by this point
         if help_texts:
-            raise TypeError("Callable does not have parameter(s) found in "
-                            "help_texts: %r" % list(help_texts))
+            raise TypeError("Callable does not have parameter(s)/output(s) "
+                            "found in help_texts: %r" % list(help_texts))
 
         return annotated_inputs, annotated_parameters, annotated_outputs
 
--- a/qiime/core/type/signature.py
+++ b/qiime/core/type/signature.py
@@ -213,9 +213,9 @@ class PipelineSignature:
         if input_descriptions or parameter_descriptions or output_descriptions:
             raise TypeError(
                 "Callable does not have parameter(s)/output(s) found in "
-                " descriptions: %r" % list({**input_descriptions,
-                                            **parameter_descriptions,
-                                            **output_descriptions}))
+                "descriptions: %r" % [*input_descriptions,
+                                      *parameter_descriptions,
+                                      *output_descriptions])
 
         return annotated_inputs, annotated_parameters, annotated_outputs
 
--- a/qiime/core/type/signature.py
+++ b/qiime/core/type/signature.py
@@ -26,10 +26,12 @@ class _NoValue:
 class ParameterSpec(ImmutableBase):
     NOVALUE = _NoValue()
 
-    def __init__(self, qiime_type=NOVALUE, view_type=NOVALUE, default=NOVALUE):
+    def __init__(self, qiime_type=NOVALUE, view_type=NOVALUE, default=NOVALUE,
+                 help_text=NOVALUE):
         self.qiime_type = qiime_type
         self.view_type = view_type
         self.default = default
+        self.help_text = help_text
 
         self._freeze_()
 
@@ -42,14 +44,19 @@ class ParameterSpec(ImmutableBase):
     def has_default(self):
         return self.default is not self.NOVALUE
 
+    def has_help_text(self):
+        return self.help_text is not self.NOVALUE
+
     def __repr__(self):
-        return ("ParameterSpec(qiime_type=%r, view_type=%r, default=%r)"
-                % (self.qiime_type, self.view_type, self.default))
+        return ("ParameterSpec(qiime_type=%r, view_type=%r, default=%r, "
+                "help_text=%r)" % (self.qiime_type, self.view_type,
+                                   self.default, self.help_text))
 
     def __eq__(self, other):
         return (self.qiime_type == other.qiime_type and
                 self.view_type == other.view_type and
-                self.default == other.default)
+                self.default == other.default and
+                self.help_text == other.help_text)
 
     def __ne__(self, other):
         return not (self == other)
@@ -61,7 +68,7 @@ class ParameterSpec(ImmutableBase):
 class PipelineSignature:
     builtin_args = ('ctx',)
 
-    def __init__(self, callable, inputs, parameters, outputs):
+    def __init__(self, callable, inputs, parameters, outputs, help_text=None):
         """
 
         Parameters
@@ -75,10 +82,12 @@ class PipelineSignature:
         outputs : list of tuple
             Each tuple contains the name of the output (str) and its QIIME
             type.
+        help_text : dict, optional
+            Parameter name to help text.
 
         """
         inputs, parameters, outputs = self._parse_signature(
-            callable, inputs, parameters, outputs)
+            callable, inputs, parameters, outputs, help_text)
 
         self._assert_valid_inputs(inputs)
         self._assert_valid_parameters(parameters)
@@ -94,7 +103,8 @@ class PipelineSignature:
             (name, spec.default) for name, spec in self.parameters.items()
             if spec.has_default()])
 
-    def _parse_signature(self, callable, inputs, parameters, outputs):
+    def _parse_signature(self, callable, inputs, parameters, outputs,
+                         help_text):
         # Copy so we can "exhaust" the collections and check for missing params
         inputs = copy.copy(inputs)
         parameters = copy.copy(parameters)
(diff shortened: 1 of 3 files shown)


qiime/core/type/signature.py
```diff
@@ -133,12 +146,12 @@ def _parse_signature(self, callable, inputs, parameters, outputs):
                                     " parameters in callable signature.")
                 annotated_inputs[name] = ParameterSpec(
                     qiime_type=inputs.pop(name), view_type=view_type,
-                    default=default)
+                    default=default, help_text=help_text.get(name))
```
ebolyen: This should be `help_text=help_text.get(name, Signature.NOVALUE)` to be consistent with other elements. (Currently will be set to `None` if not present).

ebolyen: This is the case in a couple places. You might be able to set a `help_text` variable where the `default` and `view_type` are set, but you'll need another name for the `help_text` dict.

qiime/core/type/signature.py
```diff
@@ -94,7 +103,8 @@ def defaults(self):
             (name, spec.default) for name, spec in self.parameters.items()
             if spec.has_default()])
 
-    def _parse_signature(self, callable, inputs, parameters, outputs):
+    def _parse_signature(self, callable, inputs, parameters, outputs,
+                         help_text):
         # Copy so we can "exhaust" the collections and check for missing params
         inputs = copy.copy(inputs)
         parameters = copy.copy(parameters)
```
ebolyen: It would probably be good to `copy.copy(help_text)` so that we can `pop` and see if there's anything left over at the end (these represent parameters that were forgotten/misspelled).

qiime/core/type/signature.py
jakereps: Ah, I didn't realize `None` overrode the default in the `ParameterSpec` itself.

ebolyen: Yup! That's why we use a sentinel value at all, because for this particular API `None` *is* a real value, so we have to once more disambiguate between "nothing" and "no really, nothing"

jakereps:
Thanks @ebolyen, your comments should be addressed!

qiime/core/type/signature.py
```diff
@@ -137,6 +148,9 @@ def _parse_signature(self, callable, inputs, parameters, outputs,
             default = ParameterSpec.NOVALUE
             if parameter.default is not parameter.empty:
                 default = parameter.default
+            help_text = help_texts.get(name, ParameterSpec.NOVALUE)
+            if help_text is not ParameterSpec.NOVALUE:
+                help_texts.pop(name)
```
ebolyen: This is better written as:
```python
help_text = help_texts.pop(name, ParameterSpec.NOVALUE)
```

qiime/core/type/signature.py
```diff
@@ -170,13 +184,24 @@ def _parse_signature(self, callable, inputs, parameters, outputs,
...
-                    qiime_type=qiime_type, help_text=help_text.get(name))
+                help_text = help_texts.get(name, ParameterSpec.NOVALUE)
+                if help_text is not ParameterSpec.NOVALUE:
+                    help_texts.pop(name)
+                annotated_outputs[name] = ParameterSpec(qiime_type=qiime_type,
+                                                        help_text=help_text)
+
+        # we should have popped the help_texts empty by this point
+        if help_texts:
+            raise TypeError("Callable does not have parameter(s) found in "
```
ebolyen: `paramter(s)/output(s)`

qiime/plugin/plugin.py
```diff
@@ -188,19 +188,19 @@ class PluginMethods(PluginActions):
     # TODO is `register` a better name now that functions are the only accepted
     # source (i.e. markdown support is gone)?
     def register_function(self, function, inputs, parameters, outputs, name,
-                          description, help_text=None):
+                          description, help_texts=None):
```
ebolyen: I liked the `help_text` form better. @jairideout what do you think?

qiime/core/type/signature.py
jakereps: TIL `pop` has a fallback.

qiime/plugin/plugin.py
jairideout: Can the parameter help text be stored on a per-parameter basis in the `inputs`/`parameters`/`outputs` dictionaries? With `help_texts`/`help_text` it isn't clear what the help text applies to.

jakereps: When chatting offline last Friday @ebolyen and I opted to go with a separate parameter as inlining seemed strange with requiring nested dicts. It is kind of unclear with no context though.

```python
plugin.methods.register_function(
    ...
    parameters={'timeout': Int},
    ...
    help_texts={'timeout': 'The amount of time to wait before continuing '
                           'execution of the method.',
                'table': 'The table to give the function.'}
)
```

```python
plugin.methods.register_function(
    ...
    parameters={'timeout': {'type': Int,
                            'help_text': 'The amount of time to wait before '
                                         'continuing execution of the '
                                         'method.'}},
    ...
)
```

jakereps: Any thoughts on scraping doc strings of the function instead?


```
from numpydoc.docscrape import FunctionDoc
doc = FunctionDoc(function)
print(doc["Parameters"])

>>> [('a', 'int, default: 5', ['Does something cool']), ('b', 'str', ['Wow'])]
```

Once we make the callable this info is gone, but as it is being pulled in during registration it is all there.

ebolyen: Numpydoc is definitely an option. I think I'd rather delay this (keeping it as is, just a dict) and spend the effort thinking of a better registration API altogether.

jairideout: That sounds good, I agree coming up with a better plugin registration API is a better use of time, and this solution is backwards-compatible. Can we rename to something a little more descriptive, such as `parameter_help_text` or `parameter_descriptions`?

ebolyen: I think those names might be confusing, they aren't just for "qiime parameters", but also inputs and outputs. what about `annotation_help_text` or something?

jairideout: `annotation_help_text` is confusing to me because the annotations (in the context of QIIME 2) are Python types, but the function inputs and outputs are being documented here, not their annotations. Maybe `signature_descriptions`, `signature_docs`, or `signature_help_text` is clearer? I don't really like any of those but can't think of a better name.

Another option is to have three description dictionaries: `input_descriptions`, `parameter_descriptions`, `output_descriptions`. Then it's pretty clear what these descriptions apply to.

(Note: I'm favoring "description" over "help text" because we've used `description` in `register_function` to describe the method/visualizer).

jakereps:
Thanks @ebolyen, @jairideout, and @thermokarst! Everything we discussed offline should be ready for review!

qiime/core/type/signature.py
```diff
@@ -157,11 +197,25 @@ def _parse_signature(self, callable, inputs, parameters, outputs):
...
+                annotated_outputs[name] = ParameterSpec(
+                    qiime_type=qiime_type, description=description)
+
+        # we should have popped the descriptions empty by this point
+        if input_descriptions or parameter_descriptions or output_descriptions:
+            raise TypeError(
+                "Callable does not have parameter(s)/output(s) found in "
+                " descriptions: %r" % list({**input_descriptions,
+                                            **parameter_descriptions,
+                                            **output_descriptions}))
```
ebolyen: :+1: so cool!

ebolyen: although, `list(itertools.chain(input_descriptions, parameter_descriptions, output_descriptions))` would probably make more sense, since the key-value relationship isn't important to preserve.

ebolyen: Even better: `[*input_descriptions, *parameter_descriptions, *output_descriptions]`

jakereps: Yeah, that looks good! I'll tweak it to just spread the keys, because I also noticed an extra space in the error text.

jakereps:
Thanks @ebolyen, got it tweaked! Should be good to go!

qiime/core/type/signature.py
```diff
@@ -94,10 +110,30 @@ def defaults(self):
             (name, spec.default) for name, spec in self.parameters.items()
             if spec.has_default()])
 
-    def _parse_signature(self, callable, inputs, parameters, outputs):
+    @property
+    def descriptions(self):
+        params = {**self.inputs, **self.parameters, **self.outputs}
```
jairideout: I think you lose order here when converting to a set, so the `OrderedDict` will have arbitrary order. I'm not sure how important order is here though. @ebolyen, do we need this method, or is it enough that each `ParameterSpec` has a `description`? I think the `defaults` property (above) was added for backwards-compatibility, so we may not need a corresponding property for descriptions.

jakereps: This was a helper being utilized [here](https://github.com/qiime2/q2cli/pull/108/files#diff-8b6223dfee196a9888f17c22ddabcf94R269) in `q2cli` in the same manner as `defaults`.

jairideout: Cool, @jakereps if it's useful to q2cli let's keep the method. It should probably be updated to either 1) maintain order and return an OrderedDict, or 2) return a regular dict

jakereps: This maintains that same issue, as `description` is being stored as a `NOVALUE` in the signature spec if it doesn't exist. Which raises a `TypeError: NOVALUE is not JSON serializable` on caching if you just add a `description` attribute to the action signature state. So it'd require something like:

```python
for group in 'inputs', 'outputs':
    for name, spec in getattr(action.signature, group).items():
        data = {
            'name': name,
            'repr': repr(spec.qiime_type),
        }
        if isinstance(spec.description, str):
            data['description'] = spec.description

        state['signature'][group].append(data)
```

Because we don't have access to the `ParameterSpec.NOVALUE` to compare it to, so we would have to see if it is a string every time. Unless that's fine? Having this helper that gathers all the descriptions seemed cleaner allowing, as it allows `description = descriptions.get(name, q2cli.handlers.NoDefault)` to be used on handler generation.

jairideout: I'm good with keeping this helper method. Can you let me know what you think about the ordering issue I brought up?

jakereps: I'm not sure on the ordering. Would the only issue be if someone named an output the same as the others? Because being a python function, the inputs and parameters should all be restricted to needing different names, but if our function registration allows outputs to be named the same as an input I guess it could cause issues. Otherwise all the descriptions are just being grabbed by the name of the input/parameter/output, so I don't know for sure if the ordering would affect it, as it's not just being iterated over? I don't know though.

ebolyen: In the `NOVALUE` case, it would make sense for the CLI to store an empty string. But I'm fine with leaving as is. As far as ordering, that shouldn't matter to the CLI (as JSON doesn't have order anyways), and I don't see a tangible benefit for this helper (it doesn't need to be an authoritative source on order, and we probably don't want it to be used as such).

jairideout: You're converting the ordered keys from three OrderedDicts into a set, which is unordered. Right now it's misleading because an OrderedDict with an arbitrary order is returned. We should pick one of the options I listed, I don't have a strong opinion on either.

jairideout: I'm also good with nixing this helper method in favor of storing an empty string in the CLI cache (no strong feelings).

jakereps: Works for me, I'll go with the empty string.

jairideout: Sounds good!

jakereps:
Thanks @ebolyen and @jairideout! Everything should be addressed, and the `q2cli` PR also updated to work with the new changes!

jairideout:
Awesome, thanks @jakereps!

jakereps:
🎉

PR #209, opened 2017-01-25
BUG: Normalize matplotlib backend check

jakereps:
Stumbled upon this issue while working on something else. Matplotlib backend `MacOSX` also works as `macosx`, but this check is checking case-sensitive parameters. 

Modify and compare it against a lowercase string instead.


jairideout:
Thanks @jakereps! 🌮 🌚

Issue #81, opened 2016-07-08
plugins should have a description that can be used by interfaces


gregcaporaso:
It'd be nice to have short descriptions that could be used as help text. For example, here:

![screenshot 2016-07-08 14 16 03](https://cloud.githubusercontent.com/assets/192372/16701741/9c9b47c4-4516-11e6-8a95-f59a2a40fa4c.png)

Related to qiime2/q2cli#16

jairideout:
It would be useful to have short and long descriptions, similar to actions, so that the CLI and other interfaces can choose how much detail to display. How about adding two new keyword arguments (to preserve backwards-compatibility): `short_description=None` and `description=None` ? `short_description` would be a single short sentence like what's displayed in your screenshot, and `description` is longer text that is displayed when passing `--help` to a plugin.

gregcaporaso:
I like it.

maxvonhippel:
So just a brief update on progress, I've set up my dev environment and read around 40% of the code from the 5 repositories.  I've added `short_description` and `description` to plugin.py with `=None` in the params so (as you suggested) we keep backward compatibility.  I'll do some tests and make sure nothing's broken but I mean I really see no reason why this would break anything.

It seems like the logical next step is to incorporate `description` and `short_description` into the code in appropriate places.  I can start by throwing in an extra argument to print the `short_description` tab'd across from the plugin name in the output from `--help` per @gregcaporaso's suggestion.  Are there any other particular places you can think of where these two values would definitely be worth printing?  Should I just read the code some more and see if I find anything?  Or should I just add the `short_description` print statement to the `--help` output as suggested, test to make sure nothing's broken, PR and call it a day?

jairideout:
Thanks for working on this @maxvonhippel! I'd start by submitting a PR to this repo (the QIIME 2 framework) that includes the new `short_description` and `description` API and some unit tests. I can take the lead on reviewing and merging that PR.

After that, it'd be great to hook up to q2cli so that the short description is displayed when running `qiime --help` (like in @gregcaporaso's screenshot), and the long description is displayed when running `qiime <plugin-name> --help`. To test that it's working, I recommend submitting a PR to q2-feature-table to give the plugin a description, which should become visible from the command line. Let me know if you need some description text for that plugin.

[...]

PR #222, opened 2017-02-17
plugins should have a description that can be used by interfaces

maxvonhippel:
* Added `description` and `short_description` params to plugin class
* Each has default value of None so we retain backwards-compatibility
* Added default values for each in the style of the default values for the other optional params in that class 
* Added basic unit tests for each in the style of the unit tests I saw for plugins
* Added a link to the Markdown README for improved readability

This is in response to [issue #81](https://github.com/qiime2/qiime2/issues/81#issuecomment-279822099)

### DIFF
--- a/qiime2/plugin/plugin.py
+++ b/qiime2/plugin/plugin.py
@@ -36,9 +36,9 @@ class Plugin:
         ]
         return dumper.represent_dict(items)
 
-    def __init__(self, name, version, website, package, 
-        description=None, short_description=None, 
-        citation_text=None, user_support_text=None):
+    def __init__(self, name, version, website, package,
+                 description=None, short_description=None,
+                 citation_text=None, user_support_text=None):
         self.name = name
         self.version = version
         self.website = website
--- a/qiime2/plugin/plugin.py
+++ b/qiime2/plugin/plugin.py
@@ -37,8 +37,8 @@ class Plugin:
         return dumper.represent_dict(items)
 
     def __init__(self, name, version, website, package,
-                 description=None, short_description=None,
-                 citation_text=None, user_support_text=None):
+                 citation_text=None, user_support_text=None,
+                 short_description=None, description=None):
         self.name = name
         self.version = version
         self.website = website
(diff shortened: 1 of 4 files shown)


qiime2/plugin/plugin.py
```diff
@@ -36,8 +36,9 @@ def yaml_representer(dumper, data):
         ]
         return dumper.represent_dict(items)
 
-    def __init__(self, name, version, website, package, citation_text=None,
-                 user_support_text=None):
+    def __init__(self, name, version, website, package,
+                 description=None, short_description=None,
```
jairideout: Can you put `short_description` and `description` as the last two keyword arguments in the signature? That'll be more backwards-compatible in case plugin developers are only using positional arguments (hopefully not, but you never know).

Also, can you swap the relative order of these new parameters so that `short_description` comes before `description`? That order makes more sense to me -- write the summary first and then the expanded description.

qiime2/plugin/plugin.py
```diff
@@ -53,6 +54,16 @@ def __init__(self, name, version, website, package, citation_text=None,
                                       % self.website)
         else:
             self.user_support_text = user_support_text
+        if description is None:
+            self.description = ('No description available. '
```
jairideout: 👍

jairideout:
A general comment: you made your changes on the master branch of your forked repo. This is generally discouraged with git. In the future I recommend creating a feature branch to make your changes in, and then submitting the PR from that branch instead of master. For rationale and details about how to do this, see [this guide](https://gist.github.com/Chaser324/ce0505fbed06b947d962). Don't worry about it for this PR though.

qiime2/plugin/plugin.py
maxvonhippel: Yep, makes sense!  And done :)

maxvonhippel:
Yep, understood RE PRs in branches, will do in the future.  Thanks!

jairideout:
Awesome, thanks for the quick updates @maxvonhippel!

PR #244, opened 2017-03-23
Revert "MAINT: versioneer"

ebolyen:
Temporary until we are ready to go through the entire organization

Reverts qiime2/qiime2#243


ebolyen:
Doesn't work when PRs are squashed

PR #263, opened 2017-05-17
BUG: fix SingleFileDirectoryFormat transformation

ebolyen:
When importing it is sometimes the case that the user is importing a directory
and both the supplied view_type and the registered view_type are
SingleFileDirectoryFormats with only a single transformer between their
respective file-formats. This should be fine, but the implicit transformers of:

XDirFmt -> X => Y -> YDirFmt

were not working (where -> is an implicit transformer and => is a registered
one).

### DIFF
--- a/qiime2/core/transform.py
+++ b/qiime2/core/transform.py
@@ -171,10 +171,10 @@ class SingleFileDirectoryFormatType(FormatType):
             return None
 
         if wrap_input:
-            transformer = self._wrap_input(transformer)
+            transformer = in_._wrap_input(transformer)
 
         if wrap_output:
-            transformer = self._wrap_output(transformer)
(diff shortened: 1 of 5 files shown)


qiime2/core/transform.py
```diff
@@ -171,10 +171,10 @@ def _wrap_transformer(self, in_, out_, wrap_input=False,
             return None
 
         if wrap_input:
-            transformer = self._wrap_input(transformer)
+            transformer = in_._wrap_input(transformer)
 
         if wrap_output:
-            transformer = self._wrap_output(transformer)
+            transformer = out_._wrap_output(transformer)
```
ebolyen: This is the only *real* change. `in_` and `self` are the same object technically, I just renamed for symmetry.

PR #266, opened 2017-05-22
ENH: Validate Metadata

### LINKED ISSUE #220: mapping file column with / causes error
Hey people, 
I get an error when the mapping file column contains "/" (which is not an acceptable character in the qiime mapping file format, but users may not validate with Keemei). Alpha correlation (and other alpha commands) fail in the manner attached.

screen shot 2017-02-14 at 11 37 19 am
  
Changing the "/" to a "_ " does solve the issue. It would be nice if it the error message was clearer about what the problem is. 
Thanks,
Arron

thermokarst:
Introduces several validation rules for ensuring Metadata is
well-behaved.

Fixes #220

Pair programmed with @maxvonhippel

### DIFF
--- a/qiime2/metadata.py
+++ b/qiime2/metadata.py
@@ -14,6 +14,38 @@ import pandas as pd
 
 class Metadata:
     def __init__(self, dataframe):
+        # `/` and `\0` aren't permitted because they are invalid filename
+        # characters on *nix filesystems. The remaining values aren't permitted
+        # because they *could* be misinterpreted by a shell (e.g. `*`, `|`).
+        illegal_chars = ['/', '\0', '\\', '*', '<', '>', '?', '|', '$']
+        illegal_chars = set(illegal_chars)
+        chars_for_msg = ", ".join("%r" % i for i in illegal_chars)
+
+        categories = dataframe.columns.values.tolist()
+        indices = dataframe.index.tolist()
+
+        if dataframe.columns.dtype_str != 'object':
+            msg = "Non-string Metadata category labels detected"
+            raise ValueError(invalid_metadata_template % msg)
+
+        if any(illegal_chars & set(category) for category in categories):
+            msg = "Invalid characters (e.g. %s) detected in metadata " \
+                  "category labels" % chars_for_msg
+            raise ValueError(invalid_metadata_template % msg)
+
+        if dataframe.index.dtype_str != 'object':
+            msg = "Non-string Metadata index values detected"
+            raise ValueError(invalid_metadata_template % msg)
+
+        if any(illegal_chars & set(index) for index in indices):
+            msg = "Invalid characters (e.g. %s) detected in index (e.g. " \
+                  "`SampleID`) values" % chars_for_msg
+            raise ValueError(invalid_metadata_template % msg)
+
+        if len(indices) != len(set(indices)):
+            msg = "Non-unique Metadata index values detected"
+            raise ValueError(invalid_metadata_template % msg)
+
         self._dataframe = dataframe
         self._artifact = None
 
@@ -61,11 +93,8 @@ class Metadata:
                 "Metadata file %s doesn't exist or isn't accessible (e.g., "
                 "due to incompatible file permissions)." % path)
         except (pd.io.common.CParserError, KeyError):
-            raise ValueError(
-                'Metadata file format is invalid for file %s. Currently only '
-                'QIIME 1 sample/feature metadata mapping files are officially '
-                'supported. Sample metadata mapping files can be validated '
-                'using Keemei: http://keemei.qiime.org' % path)
+            msg = 'Metadata file format is invalid for file %s' % path
+            raise ValueError(invalid_metadata_template % msg)
         return cls(df)
 
     def get_category(self, *names):
@@ -198,3 +227,9 @@ class MetadataCategory:
     def _add_artifact(self, artifact):
         """ Adds the artifact to self."""
         self._artifact = artifact
+
+
+invalid_metadata_template = "%s. There may be more errors present in this " \
+    "metadata. Currently only QIIME 1 sample/feature metadata mapping files " \
+    "are officially supported. Sample metadata files can be validated using " \
+    "Keemei: http://keemei.qiime.org."
--- a/qiime2/tests/test_metadata.py
+++ b/qiime2/tests/test_metadata.py
@@ -17,6 +17,59 @@ import qiime2
 from qiime2.core.testing.util import get_dummy_plugin
 
 
+class TestMetadata(unittest.TestCase):
+    def setUp(self):
+        self.illegal_chars = ['/', '\0', '\\', '*', '<', '>', '?', '|', '$']
+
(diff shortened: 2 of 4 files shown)


qiime2/metadata.py
```diff
@@ -14,6 +14,38 @@
 
 class Metadata:
     def __init__(self, dataframe):
+        # `/` and `\0` aren't permitted because they are invalid filename
+        # characters on *nix filesystems. The remaining values aren't permitted
+        # because they *could* be misinterpreted by a shell (e.g. `*`, `|`).
+        illegal_chars = ['/', '\0', '\\', '*', '<', '>', '?', '|', '$']
+        illegal_chars = set(illegal_chars)
```
ebolyen: Unless you are hitting length limits, I would collapse these two lines.

qiime2/metadata.py
```diff
@@ -14,6 +14,38 @@
...
+        chars_for_msg = ", ".join("%r" % i for i in illegal_chars)
```
ebolyen: Actually it might be better to leave the list intact so that the message isn't stochastic.

qiime2/metadata.py
```diff
@@ -14,6 +14,38 @@
...
+
+        categories = dataframe.columns.values.tolist()
+        indices = dataframe.index.tolist()
+
+        if dataframe.columns.dtype_str != 'object':
```
ebolyen: til about `dtype_str`, but what does it do? I can't find any docs on it. I assume it lists `str` types as `object`?

qiime2/metadata.py
```diff
@@ -14,6 +14,38 @@
...
+            msg = "Non-string Metadata category labels detected"
+            raise ValueError(invalid_metadata_template % msg)
+
+        if any(illegal_chars & set(category) for category in categories):
+            msg = "Invalid characters (e.g. %s) detected in metadata " \
+                  "category labels" % chars_for_msg
```
ebolyen: I'm ok with this (and the conditional is very pretty), but it may be worth going to the extra mile to figure out what columns are wrong and why.

qiime2/metadata.py
```diff
@@ -14,6 +14,38 @@
...
+            raise ValueError(invalid_metadata_template % msg)
+
+        if any(illegal_chars & set(index) for index in indices):
+            msg = "Invalid characters (e.g. %s) detected in index (e.g. " \
+                  "`SampleID`) values" % chars_for_msg
+            raise ValueError(invalid_metadata_template % msg)
+
+        if len(indices) != len(set(indices)):
+            msg = "Non-unique Metadata index values detected"
+            raise ValueError(invalid_metadata_template % msg)
```
ebolyen: Could we factor this into a validation helper that can be parametrized on axis? It seems like the uniqueness condition would be valuable for categories as well.

qiime2/metadata.py
```diff
@@ -61,11 +93,8 @@ def load(cls, path):
                 "Metadata file %s doesn't exist or isn't accessible (e.g., "
                 "due to incompatible file permissions)." % path)
         except (pd.io.common.CParserError, KeyError):
-            raise ValueError(
-                'Metadata file format is invalid for file %s. Currently only '
-                'QIIME 1 sample/feature metadata mapping files are officially '
-                'supported. Sample metadata mapping files can be validated '
-                'using Keemei: http://keemei.qiime.org' % path)
+            msg = 'Metadata file format is invalid for file %s' % path
+            raise ValueError(invalid_metadata_template % msg)
```
ebolyen: :+1: :+1: :+1:

qiime2/tests/test_metadata.py
```diff
@@ -17,6 +17,59 @@
...
+        with self.assertRaisesRegex(ValueError, 'Non-string.*index values'):
+            qiime2.Metadata(pd.DataFrame({'foo': ['a', 'b', 'c']}))
+
+    def test_duplicate_indices(self):
+        index = pd.Index(['b', 'b', 'b'], dtype=object)
+        df = pd.DataFrame({'foo': [1, 2, 3]}, index=index)
+
+        with self.assertRaisesRegex(ValueError, 'Non-unique.*index values'):
+            qiime2.Metadata(df)
+
```
ebolyen: Can we add a test for calling `.load` on a blob (e.g. `.qza`?)

qiime2/metadata.py
jairideout: It can even be a `set` literal:

```python
illegal_chars = {'/', '\0', '\\', ...}
```

qiime2/metadata.py
jairideout: :+1: for non-stochastic error message -- keeping the list around works, or sorting the set items before printing

qiime2/metadata.py
jairideout: :+1: , I'm in favor of describing what category or sample ID the error occurred in

qiime2/metadata.py
jairideout: :+1: for factoring, and :+1: for unique category checking (thanks, pandas...)

qiime2/tests/test_metadata.py
jairideout: :+1: , just a note that this test should be added below to `TestMetadataLoad`

thermokarst: We started to write that test, but it looked like we would need to either tweak the `Mapping` format, or define new transformers for another `dummy` plugin type. Thoughts?

ebolyen: I don't think that should be necessary. Any old artifact would work, but we want the `.save`'ed path (not the object), because it won't be a zip file until then.

qiime2/metadata.py
thermokarst: That was new to me too! Would you prefer to see a `str(dataframe.columns.dtype)` here?

qiime2/tests/test_metadata.py
ebolyen: To clarify, I expect this test to assert failure from `.load` (it should touch the `pd.io.common.CParserError` path).

qiime2/metadata.py
ebolyen: oh I see! What you have is fine then, but I think the dtype could also be `str` and that would be valid as well. This probably would only happen when someone initializes the object from the constructor (instead of `.load` or `.from_artifact`).

qiime2/tests/test_metadata.py
thermokarst: Oops --- I think I misunderstood this question, please ignore me!

thermokarst: Just to clarify what I misunderstood --- I missed the `.load` part of your request and just saw the `qza`, in the context of testing the index values. That was the test case I was referring to above in my comment about starting to write the test. Thanks for clarifying!

thermokarst:
Thanks @ebolyen & @jairideout! @maxvonhippel and I really appreciate your suggestions! Speaking of suggestions, I think I have addressed everything - ready for the next round!

PR #267, opened 2017-05-25
ENH: Adds `importable_formats` property to Plugin Manager

Kleptobismol:
Adds new property, ``importable_formats``, to Plugin Manager class
``importable_formats`` is a dictionary containing:
``name : FormatRecord``
Pair-programmed with @maxvonhippel

### DIFF
--- a/qiime2/core/testing/format.py
+++ b/qiime2/core/testing/format.py
@@ -99,6 +99,7 @@ class FourIntsDirectoryFormat(model.DirectoryFormat):
         else:
             return 'file%d.txt' % num
 
+
 class DummyFormat(TextFileFormat):
     """
     Useless format which maps to nothing, used for testing the plugin
--- a/qiime2/core/testing/format.py
+++ b/qiime2/core/testing/format.py
@@ -98,3 +98,10 @@ class FourIntsDirectoryFormat(model.DirectoryFormat):
             return 'nested/file%d.txt' % num
         else:
             return 'file%d.txt' % num
+
+class DummyFormat(TextFileFormat):
+    """
+    Useless format which maps to nothing, used for testing the plugin
+    manager.
+
+    """
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -106,7 +106,7 @@ class PluginManager:
                     from_type = transform.ModelType.from_view_type(record.format)
                     to_type = transform.ModelType.from_view_type(type_format.format)
                     if from_type.has_transformation(to_type):
-                        _importable_formats[name] = record
+                        _importable_formats[name] = record.format
 
         return _importable_formats
 
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -48,7 +48,7 @@ class PluginManager:
     def _init(self):
         self.plugins = {}
         self.semantic_types = {}
-        self.importable_formats = {}
+        self._importable_formats = {}
         self.transformers = collections.defaultdict(dict)
         self.formats = {}
         self.type_formats = []
@@ -80,7 +80,7 @@ class PluginManager:
             self.transformers[input][output] = transformer_record
 
         for name, record in plugin.formats.items():
-            self.importable_formats[name] = record
+            self._importable_formats[name] = record
             if name in self.formats:
                 raise NameError(
                     "Duplicate format registration (%r) defined in plugins: %r"
@@ -93,6 +93,13 @@ class PluginManager:
     # TODO: Should plugin loading be transactional? i.e. if there's
     # something wrong, the entire plugin fails to load any piece, like a
     # databases rollback/commit
+
+    @property
+    def importable_formats(self):
+        """Return dict of formats that are importable
+
+        """
+        return self._importable_formats
 
     @property
     def importable_types(self):
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -48,6 +48,7 @@ class PluginManager:
     def _init(self):
         self.plugins = {}
         self.semantic_types = {}
+        self.importable_formats = {}
         self.transformers = collections.defaultdict(dict)
         self.formats = {}
         self.type_formats = []
@@ -79,6 +80,7 @@ class PluginManager:
             self.transformers[input][output] = transformer_record
 
         for name, record in plugin.formats.items():
+            self.importable_formats[name] = record
             if name in self.formats:
                 raise NameError(
                     "Duplicate format registration (%r) defined in plugins: %r"
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -79,8 +79,15 @@ class PluginManager:
                                  % transformer_record)
             self.transformers[input][output] = transformer_record
 
+        #### ???? ----------------------------- 
         for name, record in plugin.formats.items():
-            self._importable_formats[name] = record
+            for type_format in self.type_formats:
+                record.has_transformation(type_format):
+                    self._importable_formats[name] = record
+                    break
+
+        #### ???? -----------------------------
+
             if name in self.formats:
                 raise NameError(
                     "Duplicate format registration (%r) defined in plugins: %r"
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -96,15 +96,20 @@ class PluginManager:
 
     @property
     def importable_formats(self):
-        """Return dict of formats that are importable
+        """Return dict of formats that are importable.
+
+        A format is importable if an installed plugin can transform it into
+        one of the canonical semantic types.
 
         """
         _importable_formats = {}
         for plugin in self.plugins.values():
             for name, record in plugin.formats.items():
                 for type_format in self.type_formats:
-                    from_type = transform.ModelType.from_view_type(record.format)
-                    to_type = transform.ModelType.from_view_type(type_format.format)
+                    from_type = transform.ModelType.from_view_type(
+                        record.format)
+                    to_type = transform.ModelType.from_view_type(
+                        type_format.format)
                     if from_type.has_transformation(to_type):
                         _importable_formats[name] = record.format
 
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -48,7 +48,6 @@ class PluginManager:
     def _init(self):
         self.plugins = {}
         self.semantic_types = {}
-        self._importable_formats = {}
         self.transformers = collections.defaultdict(dict)
         self.formats = {}
         self.type_formats = []
@@ -79,15 +78,7 @@ class PluginManager:
                                  % transformer_record)
             self.transformers[input][output] = transformer_record
 
-        to_type = transform.ModelType.from_view_type(qiime2.Metadata)
         for name, record in plugin.formats.items():
-
-            for type_format in self.type_formats:
-                print('foo')
-                #if from_type.has_transformation(to_type):
-                #    print(blarg)
-                #    self._importable_formats[name] = record
-                #    break
 
             if name in self.formats:
                 raise NameError(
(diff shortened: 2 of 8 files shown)


maxvonhippel:
I think we could probably implement an option where the user could specify a plugin and see the `importable_formats` applicable to that specific plugin.  Is this desirable, and if so, how should we format the command line args?  e.g,

````BASH
qiime tools importable_formats [plugin name]
```` ?

**Comment by jairideout** [comment]
That's a neat idea! I don't think we have an immediate use-case for restricting by plugin yet, and it'd be a bit awkward to hook up to q2cli since we want a `--show-importable-formats` option added to `qiime tools import` (instead of a separate `importable_formats` subcommand).

The `importable_formats` dictionary's values are `FormatRecords`, so you can see which plugin the format was registered in with `.plugin`. I'm in favor of waiting on this "plugin restriction" feature until there's a need for it, and there's a pretty easy workaround to filter by plugin if devs need this functionality in the meantime.

**Comment by maxvonhippel** [comment]
Following conversation with @jairideout , pseudo-code for algorithm to determine subset of formats which actually map to semantic types to return:

````
output = some list or something
for format in formats:
    for canonical type in canonical semantic types:
        if some transformation exists from that format to that canonical semantic type:
            add the format to the output list thingy or whatever

return output
````

File: qiime2/sdk/tests/test_plugin_manager.py [hunk 1]
```diff
@@ -60,6 +60,42 @@ def test_importable_types(self):
 
     # TODO: add tests for type/directory/transformer registrations
 
+    def test_importable_formats(self):
```
jairideout: Can you add a new format to the dummy plugin that doesn't have a transformer mapping it to a canonical format? That format shouldn't show up in the set of importable formats.

qiime2/sdk/plugin_manager.py
```diff
@@ -48,6 +48,7 @@ def __new__(cls):
     def _init(self):
         self.plugins = {}
         self.semantic_types = {}
+        self.importable_formats = {}
```
jairideout: Can you make this a private attribute, `_importable_formats`, and make it accessible via a read-only `importable_formats` property? I realize there are many public attributes already in `PluginManager` but moving forward I'd like to start making this class safer for SDK developers. For example, it'd be bad if an SDK user imported the `PluginManager` and started mutating the set of importable types after instantiation.

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +60,42 @@ def test_importable_types(self):
...
+        _format = self.pm.importable_formats
+        exp = {'IntSequenceV2DirectoryFormat':
+               FormatRecord(format='qiime2.plugin.model.directory_format.'
```
jairideout: Can you have these format records store the actual format type objects instead of strings? I think the tests will start failing once the assertion below is updated to check values.

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +60,42 @@ def test_importable_types(self):
...
+                            plugin=self.plugin),
+               'MappingFormat':
+               FormatRecord(format='qiime2.core.testing.format.MappingFormat',
+                            plugin=self.plugin),
+               'SingleIntFormat':
+               FormatRecord(format='qiime2.core.testing.format.'
+                                   'SingleIntFormat',
+                            plugin=self.plugin)}
+
+        self.assertTrue(set(exp).issubset(set(_format)))
```
jairideout: Can you change this assertion to `self.assertEqual(obs, exp)`? That way the dictionary keys and values will be compared instead of just the keys.

qiime2/sdk/tests/test_plugin_manager.py
maxvonhippel: Good idea!  Will do.

qiime2/sdk/tests/test_plugin_manager.py
maxvonhippel: Yeah fair point about the keys, this isn't a valid test as is.

qiime2/sdk/tests/test_plugin_manager.py
maxvonhippel: Our worry is that then it will have to be updated if new canonical formats are added.   Is this a risk?  Or desirable?

qiime2/sdk/plugin_manager.py
maxvonhippel: Sure can do. Should it still be a dictionary as it currently is in that case?

qiime2/sdk/tests/test_plugin_manager.py
Kleptobismol: done

qiime2/sdk/plugin_manager.py
Kleptobismol: done

qiime2/sdk/tests/test_plugin_manager.py
jairideout: That's a good point! I'm okay with having this test updated if new formats are added, at least until it becomes too annoying to do so in the future. A lot of the existing unit tests are coded this way and it hasn't been much of an issue to update them when the dummy plugin changes (it's also a nice safety net that forces developers to think about whether their dummy plugin changes make sense in the context of these unit tests).

qiime2/sdk/plugin_manager.py
jairideout: Yeah let's keep it the same dictionary structure 👍

qiime2/sdk/tests/test_plugin_manager.py
Kleptobismol: done

Kleptobismol:
ready for review @jairideout or @ebolyen

qiime2/sdk/tests/test_plugin_manager.py
maxvonhippel: (Taken care of)

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +66,27 @@ def test_importable_types(self):
...
+            'IntSequenceV2DirectoryFormat': IntSequenceV2DirectoryFormat,
+            'IntSequenceFormatV2': IntSequenceFormatV2,
+            'FourIntsDirectoryFormat': FourIntsDirectoryFormat,
+            'IntSequenceFormat': IntSequenceFormat
+        }
+
+        self.assertEqual(obs, exp)
+
+    def test_importable_formats_with_dummy(self):
+        exp = "DummyFormat"
```
jairideout: `exp` is an odd name here because it isn't really the "expected" result (it's just a format name). Can you remove this variable and just use the format name in the assertion below? `obs` can also be removed and the assertion just becomes `self.assertFalse('DummyFormat' in self.pm.importable_formats)`

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +66,27 @@ def test_importable_types(self):
...
+        obs = self.pm.importable_formats.keys()
+        self.assertFalse(exp in obs)
+
+        obs = [item[0] for item in self.plugin.formats.items()]
```
jairideout: It'd be better to use `self.pm.formats` (i.e. the `PluginManager`) in case there are other "dummy" plugins defined in the framework for testing.

The test case can be simplified to:

```python
self.assertTrue('DummyFormat' in self.pm.formats)
```

qiime2/core/testing/format.py
```diff
@@ -98,3 +98,11 @@ def single_ints_path_maker(self, num):
             return 'nested/file%d.txt' % num
         else:
             return 'file%d.txt' % num
+
+
+class DummyFormat(TextFileFormat):
```
maxvonhippel: Good idea, will do.

qiime2/sdk/plugin_manager.py
```diff
@@ -79,20 +79,43 @@ def _integrate_plugin(self, plugin):
...
 
     @property
+    def importable_formats(self):
+        """Return dict of formats that are importable.
+
+        A format is importable if an installed plugin can transform it into
+        one of the canonical semantic types.
+
+        """
+        _importable_formats = {}
```
maxvonhippel: I'll change it, it's almost as much work to write a todo as it is to just do it right :) thanks

qiime2/sdk/plugin_manager.py
```diff
@@ -79,20 +79,43 @@ def _integrate_plugin(self, plugin):
...
+        for plugin in self.plugins.values():
+            for name, record in plugin.formats.items():
+                for type_format in self.type_formats:
+                    from_type = transform.ModelType.from_view_type(
+                        record.format)
+                    to_type = transform.ModelType.from_view_type(
+                        type_format.format)
+                    if from_type.has_transformation(to_type):
+                        _importable_formats[name] = record.format
```
maxvonhippel: I was a bit concerned about the scenario in which multiple plugins use the same format.  Is this a concern?  If so we might have duplicate formats (which I guess could be OK, I'm not totally sure).

jairideout: It shouldn't be an issue because the `FormatRecord` tracks which plugin the format is *defined* in, not which plugin(s) use the format. Also, `PluginManager` disallows duplicate formats from being registered in `_integrate_plugins()`.

qiime2/core/testing/format.py
maxvonhippel: Done.

qiime2/core/testing/format.py
```diff
@@ -98,3 +98,11 @@ def single_ints_path_maker(self, num):
...
+    """
+    Useless format which maps to nothing, used for testing the plugin
```
maxvonhippel: Good idea on both fronts, and done

qiime2/sdk/plugin_manager.py
```diff
@@ -79,20 +79,43 @@ def _integrate_plugin(self, plugin):
...
     # TODO: Should plugin loading be transactional? i.e. if there's
     # something wrong, the entire plugin fails to load any piece, like a
     # databases rollback/commit
 
     @property
+    def importable_formats(self):
+        """Return dict of formats that are importable.
+
+        A format is importable if an installed plugin can transform it into
+        one of the canonical semantic types.
```
maxvonhippel: Yeah that's a better wording, thanks.  And done.

qiime2/sdk/plugin_manager.py
maxvonhippel: Done, and when you review this look at the `_post_init()` function to see my precise implementation as well as my slight modification of `importable_types`.

qiime2/sdk/plugin_manager.py
```diff
@@ -79,20 +79,43 @@ def _integrate_plugin(self, plugin):
...
+
+        """
+        _importable_formats = {}
+        for plugin in self.plugins.values():
```
maxvonhippel: Good point, and done.  Also I worked in `importable_formats` to save us another loop.

qiime2/sdk/plugin_manager.py
maxvonhippel: Right!  Good catch, and done!

maxvonhippel: Actually, I'm not going to do this for now because I think the benefit I get from leaving it in in order to work in `importable_types` probably outweighs the cost.  Once I push I'd like to know if you have any ideas RE how to achieve a similar affect using `while` and `pop()` or some other such mechanism.  Thanks!

qiime2/sdk/tests/test_plugin_manager.py
maxvonhippel: Done.

qiime2/sdk/tests/test_plugin_manager.py
maxvonhippel: Good point, done, thanks

maxvonhippel:
[Moving PR review and discussion here](https://github.com/qiime2/qiime2/pull/275).  This PR can be closed.  Thanks!

PR #275, opened 2017-06-02
ENH: Adds `importable_formats` property to Plugin Manager

maxvonhippel:
@Kleptobismol and I pair-programmed the first PR for this feature, which can be found [at this link](https://github.com/qiime2/qiime2/pull/267).  I responded to @jairideout 's code review on that PR, and am moving that PR and subsequent review and discussion here given that @Kleptobismol is out of the office (and I didn't plan ahead and set myself up with write access to his branch).  Thus that PR can be closed and subsequent review should happen here.

While I primarily followed @jairideout 's advice, I did do a few things differently, which are worth pointing out.

1. I had some trouble dealing with the [Plugin Manager's Transformer's](https://github.com/qiime2/qiime2/blob/master/qiime2/sdk/plugin_manager.py#L51) -> [Plugin Manager](https://github.com/qiime2/qiime2/blob/master/qiime2/core/transform.py#L45).  (Each of those links separately points to the relevant code, for clarification.)  I asked @thermokarst for help, and he came up with the idea of a `_post_init()` function to solve this problem.  His solution works, but there are certainly other ways to address this.  I like Matt's solution but am definitely open to thoughts on that regard.

````PYTHON
def __new__(cls):
        if cls.__instance is None:
            self = super().__new__(cls)
            self._init()
            cls.__instance = self
            self._post_init()
        return cls.__instance

def _init(self):
    # ...

def _post_init(self):
   # ...
````

2. I incorporated the population of the `_importable_formats` and `_importable_types` sets into a single set of loops in my `_post_init()` function.  Unfortunately, I have not been able to figure out how to skip redundant steps in my `_importable_formats` recursion.  What I'd really like to do is copy `self.formats.values()` to some new list or something, and then `pop()` `record`s when they are added, so that each time I iterate I am only looping over the stuff that hasn't yet been added to `self._importable_formats`.  I gave this a go and ran into a couple problems.  

    Most importantly, I started getting non-deterministic errors - which pretty much just suggests I don't fully understand how recursion over something you are actively modifying really works.  Maybe I need a `while` loop?  

    And secondly, I became concerned with the actual mechanisms of `pop()` in Python.  Would all that popping ultimately waste more time/memory than it saved, because the popping code still has to iterate over `O(n)`, and so with whatever extra iota of computation is involved in doing that work maybe you come out a net zero or even net negative?  I have no idea.  We're not really looking at a ton of items so it might not matter that much (ultimately we might favor readability over super crazy linker-loader-hardcore-whatever-ninja-optimization), but I don't claim to know the answer here and would appreciate input.

````PYTHON
for type_format in self.type_formats:
            for type in type_format.type_expression:
                self._importable_types.add(type)
            for record in self.formats.values():
                from_type = transform.ModelType.from_view_type(
                    record.format)
                to_type = transform.ModelType.from_view_type(
                    type_format.format)
                if from_type.has_transformation(to_type):
                    self._importable_formats.add(record)
                    # once this happens everything after in the loop is redundant
                    # but a break doesn't work because it's the outer, rather than inner,
                    # loop which we would want to escape
                    # and switching the loop order so I can use break would strip the advantages
                    # currently being accrued by self._importable_types ... :0 :0
````

3. I am now returning entire `FormatRecord`s, which means the [q2cli PR about this feature](https://github.com/qiime2/q2cli/pull/138) needs some re-working to accommodate this change in SDK, and this should be taken into account when working on the related [Q2Studio issue](https://github.com/qiime2/q2studio/issues/115).

### DIFF
--- a/qiime2/core/testing/format.py
+++ b/qiime2/core/testing/format.py
@@ -99,6 +99,7 @@ class FourIntsDirectoryFormat(model.DirectoryFormat):
         else:
             return 'file%d.txt' % num
 
+
 class DummyFormat(TextFileFormat):
     """
     Useless format which maps to nothing, used for testing the plugin
--- a/qiime2/core/testing/format.py
+++ b/qiime2/core/testing/format.py
@@ -108,9 +108,5 @@ class UnimportableFormat(TextFileFormat):
     """
 
 
-class UnimportableDirectoryFormat(model.DirectoryFormat):
-    """
-    Useless format which maps to nothing, used for testing the plugin
-    manager.
-
-    """
+UnimportableDirectoryFormat = model.SingleFileDirectoryFormat(
+    'UnimportableDirectoryFormat', 'ints.txt', UnimportableFormat)
--- a/qiime2/core/testing/format.py
+++ b/qiime2/core/testing/format.py
@@ -98,3 +98,10 @@ class FourIntsDirectoryFormat(model.DirectoryFormat):
             return 'nested/file%d.txt' % num
         else:
             return 'file%d.txt' % num
+
+class DummyFormat(TextFileFormat):
+    """
+    Useless format which maps to nothing, used for testing the plugin
+    manager.
+
+    """
--- a/qiime2/core/testing/format.py
+++ b/qiime2/core/testing/format.py
@@ -100,7 +100,15 @@ class FourIntsDirectoryFormat(model.DirectoryFormat):
             return 'file%d.txt' % num
 
 
-class DummyFormat(TextFileFormat):
+class UnimportableFormat(TextFileFormat):
+    """
+    Useless format which maps to nothing, used for testing the plugin
+    manager.
+
+    """
+
+
+class UnimportableDirectoryFormat(model.DirectoryFormat):
     """
     Useless format which maps to nothing, used for testing the plugin
     manager.
--- a/qiime2/core/testing/format.py
+++ b/qiime2/core/testing/format.py
@@ -102,9 +102,7 @@ class FourIntsDirectoryFormat(model.DirectoryFormat):
 
 class UnimportableFormat(TextFileFormat):
     """
-    Useless format which maps to nothing, used for testing the plugin
-    manager.
-
+    Unimportable format used for testing.
     """
 
 
--- a/qiime2/core/testing/transformer.py
+++ b/qiime2/core/testing/transformer.py
@@ -16,6 +16,8 @@ from .format import (
     IntSequenceFormatV2,
     SingleIntFormat,
     MappingFormat,
+    UnimportableFormat,
+    UnimportableDirectoryFormat
 )
 from .plugin import dummy_plugin
 
@@ -147,3 +149,13 @@ def _1(data: list) -> FourIntsDirectoryFormat:
     for i, int_ in enumerate(data, 1):
         df.single_ints.write_data(int_, int, num=i)
     return df
+
+
+@dummy_plugin.register_transformer
+def _4(ff: UnimportableFormat) -> int:
+    return 1
+
+
+@dummy_plugin.register_transformer
+def _6(df: UnimportableDirectoryFormat) -> int:
+    return 2
--- a/qiime2/core/testing/transformer.py
+++ b/qiime2/core/testing/transformer.py
@@ -16,8 +16,7 @@ from .format import (
     IntSequenceFormatV2,
     SingleIntFormat,
     MappingFormat,
-    UnimportableFormat,
-    UnimportableDirectoryFormat
+    UnimportableFormat
 )
 from .plugin import dummy_plugin
 
@@ -154,8 +153,3 @@ def _1(data: list) -> FourIntsDirectoryFormat:
 @dummy_plugin.register_transformer
 def _4(ff: UnimportableFormat) -> int:
     return 1
-
-
-@dummy_plugin.register_transformer
-def _6(df: UnimportableDirectoryFormat) -> int:
-    return 2
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -43,6 +43,7 @@ class PluginManager:
             self = super().__new__(cls)
             self._init()
             cls.__instance = self
+            self._post_init()
         return cls.__instance
 
     def _init(self):
@@ -51,6 +52,7 @@ class PluginManager:
         self.transformers = collections.defaultdict(dict)
         self.formats = {}
         self.type_formats = []
+        self._importable_formats = {}
 
         # These are all dependent loops, each requires the loop above it to
         # be completed.
@@ -60,6 +62,16 @@ class PluginManager:
 
         for plugin in self.plugins.values():
             self._integrate_plugin(plugin)
+
+    def _post_init(self):
+        for name, record in self.formats.items():
+            for type_format in self.type_formats:
+                from_type = transform.ModelType.from_view_type(
+                    record.format)
+                to_type = transform.ModelType.from_view_type(
+                    type_format.format)
+                if from_type.has_transformation(to_type):
+                    self._importable_formats[name] = record.format
 
     def _integrate_plugin(self, plugin):
         for type_name, type_record in plugin.types.items():
@@ -102,18 +114,8 @@ class PluginManager:
         one of the canonical semantic types.
 
         """
-        _importable_formats = {}
-        for plugin in self.plugins.values():
-            for name, record in plugin.formats.items():
-                for type_format in self.type_formats:
-                    from_type = transform.ModelType.from_view_type(
-                        record.format)
-                    to_type = transform.ModelType.from_view_type(
-                        type_format.format)
-                    if from_type.has_transformation(to_type):
-                        _importable_formats[name] = record.format
 
-        return _importable_formats
+        return self._importable_formats
 
     @property
     def importable_types(self):
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -52,7 +52,7 @@ class PluginManager:
         self.transformers = collections.defaultdict(dict)
         self.formats = {}
         self.type_formats = []
-        self._importable_formats = {}
+        self._importable_formats = set()
         self._importable_types = set()
 
         # These are all dependent loops, each requires the loop above it to
@@ -67,13 +67,13 @@ class PluginManager:
     # This hook is for setting up additional properties that *require* the PluginManager be instantiated
     def _post_init(self):
         for type_format in self.type_formats:
-            for name, record in self.formats.items():
+            for record in self.formats.values():
                 from_type = transform.ModelType.from_view_type(
                     record.format)
                 to_type = transform.ModelType.from_view_type(
                     type_format.format)
                 if from_type.has_transformation(to_type):
-                    self._importable_formats[name] = record.format
+                    self._importable_formats.add(record)
             for type in type_format.type_expression:
                 self._importable_types.add(type)
 
@@ -112,7 +112,7 @@ class PluginManager:
 
     @property
     def importable_formats(self):
-        """Return dict of formats that are importable.
+        """Return set of formats that are importable.
 
         A format is importable if an installed plugin can transform it into
         one of the canonical semantic types.
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -106,7 +106,7 @@ class PluginManager:
                     from_type = transform.ModelType.from_view_type(record.format)
                     to_type = transform.ModelType.from_view_type(type_format.format)
                     if from_type.has_transformation(to_type):
-                        _importable_formats[name] = record
+                        _importable_formats[name] = record.format
 
         return _importable_formats
 
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -53,6 +53,7 @@ class PluginManager:
         self.formats = {}
         self.type_formats = []
         self._importable_formats = {}
+        self._importable_types = set()
 
         # These are all dependent loops, each requires the loop above it to
         # be completed.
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

**Comment by maxvonhippel** [comment]
@thermokarst , @ebolyen and I just had a brief conversation about the optimization problem I discuss in my comment above.  Here's the TL;DR:

1. `Q2Studio` only asks for `importable`**_anything** once, so memoization doesn't help there.
2. `Q2Cli` creates a new instance each time it asks, so there's no advantage to memoization of `importable`**_anything** there.
3. Most of the time users (and robots and code thingies and such) which employ `qiime2` directly will only ask for `importable`**_anything** once, unless they fall into categories 1 or 2, in which case, as already explained, memoization is **still** not useful.
4. Figuring out the answer to `importable`**_anything** is neither easy nor fast, so why force this burden onto the user upon `init` of `PluginManager` if most users will never ask and those who do ask can wait upon asking anyway, and there's no reason to memoize anything, as expanded upon in points 1, 2, and 3.

@jairideout , you previously encouraged @Kleptobismol and I to stick the calculation of `importable_types` and `importable_formats` in the initialization.  What do you think of the points above?  Should we move those two calculations back to their respective property functions?  Or is there some advantage to keeping the code where it is now which @thermokarst and @ebolyen and I missed?

Thanks everyone!

**Comment by jairideout** [comment]
> @jairideout , you previously encouraged @Kleptobismol and I to stick the calculation of importable_types and importable_formats in the initialization. What do you think of the points above? 

Those points all make sense, thanks @maxvonhippel @ebolyen @thermokarst! I suggested moving that code into the constructor because it sounded easy to do, but it looks like there are circular dependencies on `PluginManager` that complicate things. I'm totally fine with leaving `importable_types` and `importable_formats` lazy, where they are recomputed each time the property is invoked.

Let me know when this is ready for a final pass, and thanks guys for your work on it!

**Comment by maxvonhippel** [comment]
Cool, thanks @jairideout .  I moved the code out to be lazy for now.  I think we can wait until we've reworked the `PluginManager` as part of building transitivity into transformers before possibly moving the code back into the initialization.  I'll push now, and then I believe we will be ready for a final pass ( @Kleptobismol ?).  Thanks!

Also, the `q2cli` PR is edited to accommodate this latest push, so once this is accepted that PR should be ready to be looked at.

File: qiime2/core/testing/format.py [hunk 1]
```diff
@@ -98,3 +98,19 @@ def single_ints_path_maker(self, num):
...
+
+class UnimportableFormat(TextFileFormat):
+    """
+    Useless format which maps to nothing, used for testing the plugin
+    manager.
+
+    """
+
+
+class UnimportableDirectoryFormat(model.DirectoryFormat):
```
jairideout: Can you make this a single-file directory format using the `model.SingleFileDirectoryFormat` factory? There's examples of that in this file, e.g. [this line](https://github.com/qiime2/qiime2/pull/275/files#diff-e363c313e0bcaae16f7f16b373d2ea2bL69)

qiime2/core/testing/format.py
```diff
@@ -98,3 +98,19 @@ def single_ints_path_maker(self, num):
             return 'nested/file%d.txt' % num
         else:
             return 'file%d.txt' % num
+
+
+class UnimportableFormat(TextFileFormat):
+    """
+    Useless format which maps to nothing, used for testing the plugin
```
jairideout: This comment isn't entirely accurate since there's a transformer mapping it to *something* -- I'd be fine with just removing the docstring now that the format name is self-explanatory.

qiime2/core/testing/transformer.py
```diff
@@ -147,3 +149,13 @@ def _1(data: list) -> FourIntsDirectoryFormat:
...
     return df
+
+
+@dummy_plugin.register_transformer
+def _4(ff: UnimportableFormat) -> int:
+    return 1
+
+
+@dummy_plugin.register_transformer
+def _6(df: UnimportableDirectoryFormat) -> int:
```
jairideout: Once `UnimportableDirectoryFormat` is removed in favor of using the single-file directory factory this transformer can be removed.

qiime2/sdk/plugin_manager.py
```diff
@@ -93,18 +93,38 @@ def _integrate_plugin(self, plugin):
...
         A concrete semantic type is importable if it has an associated
         directory format.
 
         """
-        types = set()
-        for record in self.type_formats:
-            for type in record.type_expression:
-                types.add(type)
-        return types
+        _importable_types = set()
```
jairideout: Just a stylistic note that it's not necessary to prefix local variables with an underscore since they're local (*private*) to the function.

qiime2/sdk/plugin_manager.py
```diff
@@ -93,18 +93,38 @@ def _integrate_plugin(self, plugin):
     # databases rollback/commit
 
     @property
+    def importable_formats(self):
+        """Return set of formats that are importable.
```
jairideout: The return type we're looking for here is a `dict` mapping format names (strings) to `FormatRecord` objects (it's the same structure that `PluginManager.formats` uses). It looks like a `set` of `FormatRecord` objects are being returned currently.

qiime2/core/testing/format.py
maxvonhippel: Good point, will do!

qiime2/core/testing/format.py
maxvonhippel: Absolutely.

qiime2/sdk/plugin_manager.py
maxvonhippel: What's the advantage of returning it as a dict?  In its current form you can pretty easily get the name using `thing_in_the_set.format.__name__`, so it seems to me like putting that in a dict results in redundancy.  Am I missing something?  Or is it just more elegant in a dict?

maxvonhippel: (Following conversation I am changing it a dict for user convenience and API transparency)

qiime2/core/testing/format.py
```diff
@@ -98,3 +98,15 @@ def single_ints_path_maker(self, num):
...
```
jairideout: ping on this comment

qiime2/sdk/plugin_manager.py
jairideout: Can you update this sentence to: *Return formats that are importable.*

qiime2/sdk/plugin_manager.py
```diff
@@ -93,18 +93,38 @@ def _integrate_plugin(self, plugin):
...
+
+        A format is importable in a QIIME 2 deployment if it can be transformed
+        into at least one of the canonical semantic type formats.
+
+        """
+        importable_formats = {}
+        for name, record in self.formats.items():
+            for type_format in self.type_formats:
+                from_type = transform.ModelType.from_view_type(
```
jairideout: Sorry I didn't notice this before: can you move the `from_type` instantiation out of the inner loop since it doesn't change in this loop? That may save a little time.

qiime2/core/testing/format.py
maxvonhippel: Deleting the comment entirely causes `flake8` issues, so I cut the comment down considerably in size and made it more precise.

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +68,36 @@ def test_importable_types(self):
...
+            'IntSequenceDirectoryFormat': IntSequenceDirectoryFormat,
+            'MappingDirectoryFormat': MappingDirectoryFormat,
+            'IntSequenceV2DirectoryFormat': IntSequenceV2DirectoryFormat,
+            'IntSequenceFormatV2': IntSequenceFormatV2,
+            'FourIntsDirectoryFormat': FourIntsDirectoryFormat,
+            'IntSequenceFormat': IntSequenceFormat
+        }
+        self.assertEqual(set(obs), set(exp))
+
+    def test_importable_formats_with_dummy(self):
```
jairideout: I think the test name is outdated from when `DummyFormat` existed -- maybe update to something like `test_importable_formats_excludes_unimportables`?

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +68,36 @@ def test_importable_types(self):
 
     # TODO: add tests for type/directory/transformer registrations
 
+    def test_importable_formats(self):
+        obs = self.pm.importable_formats
+        exp = {
+            'IntSequenceDirectoryFormat': IntSequenceDirectoryFormat,
```
jairideout: These values aren't quite right, they should be `FormatRecord` objects. The assertion below is only testing the keys (since you're turning `dict` into `set`, which only inserts the dict's keys into the set).

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +68,36 @@ def test_importable_types(self):
...
+            'MappingDirectoryFormat': MappingDirectoryFormat,
+            'IntSequenceV2DirectoryFormat': IntSequenceV2DirectoryFormat,
+            'IntSequenceFormatV2': IntSequenceFormatV2,
+            'FourIntsDirectoryFormat': FourIntsDirectoryFormat,
+            'IntSequenceFormat': IntSequenceFormat
+        }
+        self.assertEqual(set(obs), set(exp))
```
jairideout: Update to use `self.assertEqual(obs, exp)` once comment above is addressed.

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +68,36 @@ def test_importable_types(self):
...
+
+    def test_importable_formats_with_dummy(self):
+        obs = self.pm.importable_formats
+        self.assertFalse('UnimportableFormat' in obs.keys())
```
jairideout: Minor: when checking if a key exists you can just use `in` directly, e.g. `'UnimportableFormat' in obs`. Even shorter, you can use `self.assertNotIn`:

```python
self.assertNotIn('UnimportableFormat', obs)
self.assertNotIn('UnimportableDirectoryFormat', obs)
```

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +68,36 @@ def test_importable_types(self):
...
+        self.assertFalse(UnimportableFormat in obs.values())
```
jairideout: This will never evaluate `True` because `.values()` contains `FormatRecords` instead of the formats themselves. I'm okay with removing this assertion (and the corresponding directory format assertion below) and just check the keys like you're doing.

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +68,36 @@ def test_importable_types(self):
...
+        self.assertFalse('UnimportableDirectoryFormat' in obs.keys())
+        self.assertFalse(UnimportableDirectoryFormat in obs.values())
+
+        obs = self.pm.formats.items()
+        exp = {
```
jairideout: It's probably enough to just assert that the format names exist in `.formats` keys:

```python
self.assertIn('UnimportableFormat', obs)
self.assertIn('UnimportableDirectoryFormat', obs)
```

qiime2/core/testing/format.py
jairideout: looks like you addressed this one during review, thanks!

qiime2/sdk/plugin_manager.py
maxvonhippel: Right!  Good catch.

qiime2/sdk/plugin_manager.py
maxvonhippel: Yeah, good idea!

qiime2/sdk/tests/test_plugin_manager.py
maxvonhippel: Oh ok, I'll work on rectifying that test.

maxvonhippel: (Done)

qiime2/sdk/tests/test_plugin_manager.py
maxvonhippel: Got it, done, thanks!

qiime2/sdk/tests/test_plugin_manager.py
maxvonhippel: Very cool!  I was unaware of `assertNotIn`.

qiime2/sdk/tests/test_plugin_manager.py
maxvonhippel: Ok, fair enough, thanks.

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +66,39 @@ def test_importable_types(self):
...
+                             plugin=self.plugin),
+            'IntSequenceFormat':
+                FormatRecord(format=IntSequenceFormat,
+                             plugin=self.plugin)
+        }
+        self.assertEqual(obs, exp)
+
+    def test_importable_formats_with_dummy(self):
+        obs = self.pm.importable_formats
+        self.assertNotIn('UnimportableFormat', obs.keys())
```
jairideout: calling `.keys()` isn't necessary on this line or the one below -- `in` will operate on a `dict's` keys implicitly

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +66,39 @@ def test_importable_types(self):
...
+        self.assertNotIn('UnimportableDirectoryFormat', obs.keys())
+
+        obs = self.pm.formats.keys()
```
jairideout: You don't need to call `.keys()` here either:

```python
obs = self.pm.formats
self.assertIn('UnimportableFormat', obs)
self.assertIn('UnimportableDirectoryFormat', obs)
```

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -60,6 +66,39 @@ def test_importable_types(self):
...
+            'FourIntsDirectoryFormat':
+                FormatRecord(format=FourIntsDirectoryFormat,
+                             plugin=self.plugin),
+            'IntSequenceFormat':
+                FormatRecord(format=IntSequenceFormat,
+                             plugin=self.plugin)
+        }
+        self.assertEqual(obs, exp)
+
+    def test_importable_formats_with_dummy(self):
```
jairideout: ping on [renaming the test](https://github.com/qiime2/qiime2/pull/275#discussion_r120426185)

maxvonhippel: Oh, sorry I missed that before!  Thanks.

maxvonhippel:
Thanks!

PR #283, opened 2017-06-09
ENH: Adds basic docstring to Actions

### LINKED ISSUE #278: add basic docstring to actions
When `?` is called on actions (methods or visualizers), at the moment no docstring is displayed. Add a basic docstring displaying short and long descriptions.

Kleptobismol:
At the moment, when `?` is called on any QIIME 2 action, this is the help text that Jupyter produces (my examples use ``feature_table.rarefy``):
screen shot 2017-06-08 at 4 54 15 pm
This is extremely general, and not very helpful to the user. The only arguably useful information it provides is the arguments the function requires. 

After my changes, this is the help text that Jupyter produces:
screen shot 2017-06-08 at 4 57 59 pm
As you can see, now there is more information regarding the Method. The user sees that it is a QIIME 2 Method object, and is given some useful help text. 

fixes #278 
@jairideout, @ebolyen, @thermokarst, very simple, ready for review.

### DIFF
--- a/qiime2/sdk/action.py
+++ b/qiime2/sdk/action.py
@@ -11,6 +11,7 @@ import concurrent.futures
 import inspect
 import os.path
 import tempfile
+import textwrap
 
 import decorator
 
@@ -40,6 +41,8 @@ def _async_action(action, args, kwargs):
 
 
 class Action(metaclass=abc.ABCMeta):
+    """ QIIME 2 Action """
+
     type = 'action'
 
     __call__ = LateBindingAttribute('_dynamic_call')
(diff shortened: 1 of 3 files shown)


qiime2/sdk/action.py
```diff
@@ -254,6 +257,8 @@ def _rewrite_wrapper_signature(self, wrapper):
     def _set_wrapper_properties(self, wrapper, name):
         wrapper.__name__ = wrapper.__qualname__ = name
         wrapper.__module__ = self.package
+        wrapper.__doc__ = "{}\n\n{}".format(
+            self.name, '\n'.join(textwrap.wrap(self.description, width=79)))
```
jairideout: This `\n` joining works great, but can you use [`textwrap.fill`](https://docs.python.org/3/library/textwrap.html#textwrap.fill) instead? That function will do the `wrap` and `\n` joining for you, so you can get this same functionality in one `textwrap.fill` call.

Kleptobismol: done

PR #285, opened 2017-06-12
ENH: Adds annotations to Actions

Kleptobismol:
As of now, when ``?`` is called on a QIIME 2 Method or Visualizer, this is the general format of the help text produced by Jupyter:
screen shot 2017-06-12 at 1 21 06 pm

This help text is helpful, although it is missing a key component that is now tied to all Methods and Visualizers, due to some new Python functionality, that component being _annotations_. This PR adds annotations (including return annotations) and testing to Actions.

Calling ``?`` with the PR in effect:
screen shot 2017-06-12 at 1 21 40 pm

### DIFF
--- a/qiime2/sdk/action.py
+++ b/qiime2/sdk/action.py
@@ -259,11 +259,30 @@ class Action(metaclass=abc.ABCMeta):
         wrapper.__module__ = self.package
         wrapper.__doc__ = "{}\n\n{}".format(
             self.name, textwrap.fill(self.description, width=79))
-        del wrapper.__annotations__
+        wrapper.__annotations__ = self._build_annotations()
         # This is necessary so that `inspect` doesn't display the wrapped
         # function's annotations (the annotations apply to the "view API" and
         # not the "artifact API").
         del wrapper.__wrapped__
+
+    def _build_annotations(self):
+        annotations = {}
+        sig = self.signature
+
+        for inpt in sig.inputs:
+            annotations[inpt] = sig.inputs[inpt].qiime_type
+
+        for param in sig.parameters:
+            annotations[param] = sig.parameters[param].qiime_type
+
+        output = []
+        for out in sig.outputs:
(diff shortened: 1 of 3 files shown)


Kleptobismol:
@jairideout ready for review

qiime2/sdk/tests/test_visualizer.py
```diff
@@ -15,7 +15,8 @@
 
 import qiime2.plugin
 import qiime2.core.type
-from qiime2.core.type import VisualizerSignature
+from qiime2.core.type import VisualizerSignature, Str
+from qiime2.core.type.visualization import Visualization as VisualizationType
```
ebolyen: :+1:

jairideout:
> I think we need a test that the __doc__ property is set, we probably don't need an exact match but just checking that a keyword is present would work.

There's tests for `__doc__` in `test_method.py` and `test_visualizer.py` that were included in #283 (I might be misunderstanding what you're asking for though).

ebolyen:
@jairideout thanks, I totally missed that part.

qiime2/sdk/action.py
```diff
@@ -259,12 +259,31 @@ def _set_wrapper_properties(self, wrapper, name):
...
+        for inpt in sig.inputs:
+            annotations[inpt] = sig.inputs[inpt].qiime_type
+
+        for param in sig.parameters:
+            annotations[param] = sig.parameters[param].qiime_type
+
+        output = []
+        for out in sig.outputs:
+            output.append(sig.outputs[out].qiime_type)
+        output = tuple(output)
```
ebolyen: If it a single value, it will not be wrapped in a tuple (for normal Python functions). However, we always return a `Results` object so this actually works perfectly, since you do need to unwrap the result (which is basically a tuple).
:+1:

ebolyen: `.async` does get a bit weird, but I think the annotation is more useful this way anyhow.

PR #286, opened 2017-06-13
BUG: fix optional support in decode_parameters

ebolyen:
Discovered while updated q2studio.

This should *probably* have unit-tests

### DIFF
--- a/qiime2/core/type/signature.py
+++ b/qiime2/core/type/signature.py
@@ -268,7 +268,9 @@ class PipelineSignature:
     def decode_parameters(self, **kwargs):
         params = {}
         for key, spec in self.parameters.items():
-            if spec.has_default() and spec.default is kwargs[key] is None:
+            if (spec.has_default() and
+                    spec.default is None and
+                    kwargs[key] is None):
                 params[key] = None
             else:
                 params[key] = spec.qiime_type.decode(kwargs[key])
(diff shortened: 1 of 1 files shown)


qiime2/core/type/signature.py
```diff
@@ -268,7 +268,10 @@ def _assert_valid_outputs(self, outputs):
     def decode_parameters(self, **kwargs):
         params = {}
         for key, spec in self.parameters.items():
-            params[key] = spec.qiime_type.decode(kwargs[key])
+            if spec.has_default() and spec.default is kwargs[key] is None:
```
jairideout: Woah, I've never seen this form of conditional before (`<something> is <something> is <something>`). I played around with it locally and it seems to work... maybe it's clearer to make this a more "standard" conditional to help with readability?

```python
if spec.has_default() and spec.default is None and kwargs[key] is None:
    ...
```

thermokarst: @ebolyen, I am with @jairideout on that one: can you update to his proposed form?

ebolyen: Sure, the method below works the same way FYI

ebolyen:
For reference, here are [the docs](https://docs.paython.org/3/reference/expressions.html#comparisons) for why that chaining of `is` worked.

PR #289, opened 2017-06-24
ENH: support optional artifacts

### LINKED ISSUE #235: support optional artifacts

jairideout:
Methods and visualizers now support optional artifacts. If an artifact is optional the only supported default value is `None`.

With this change, signature inputs and parameters are now allowed to be declared in any order (previously inputs were required to be followed by parameters, and they were not allowed to be mixed).

Removed `Signature.defaults` property in favor of using the `ParameterSpec` objects available within `Signature`.

Fixes #235.

### DIFF
--- a/qiime2/core/archive/provenance.py
+++ b/qiime2/core/archive/provenance.py
@@ -285,7 +285,7 @@ class ActionProvenanceCapture(ProvenanceCapture):
 
     def add_input(self, name, artifact):
         if artifact is None:
-            self.inputs[name] = artifact
+            self.inputs[name] = None
         else:
             ancestral_provenance = self.add_ancestor(artifact)
             if ancestral_provenance is NotImplemented:
--- a/qiime2/core/archive/provenance.py
+++ b/qiime2/core/archive/provenance.py
@@ -284,11 +284,14 @@ class ActionProvenanceCapture(ProvenanceCapture):
         self.parameters[name] = handler(parameter)
 
     def add_input(self, name, artifact):
-        ancestral_provenance = self.add_ancestor(artifact)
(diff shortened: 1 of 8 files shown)


qiime2/core/archive/provenance.py
```diff
@@ -284,11 +284,14 @@ def add_parameter(self, name, type_expr, parameter):
         self.parameters[name] = handler(parameter)
 
     def add_input(self, name, artifact):
-        ancestral_provenance = self.add_ancestor(artifact)
-        if ancestral_provenance is NotImplemented:
-            self.inputs[name] = NoProvenance(artifact.uuid)
+        if artifact is None:
+            self.inputs[name] = artifact
```
BenKaehler: `self.inputs[name] = None` might be slightly clearer.

qiime2/sdk/action.py
```diff
@@ -196,8 +190,11 @@ def callable_wrapper(*args, **kwargs):
             view_args = parameters.copy()
             for name, spec in self.signature.inputs.items():
                 recorder = provenance.transformation_recorder(name)
-                view_args[name] = artifacts[name]._view(spec.view_type,
-                                                        recorder)
+                artifact = artifacts[name]
+                if artifact is None:
+                    view_args[name] = artifact
```
BenKaehler: `view_args[name] = None` might be slightly clearer.

jairideout:
Thanks for reviewing @BenKaehler and @ebolyen! I made the suggested changes in two separate commits. Whoever ends up merging this, please "Squash and Merge" and use the current PR description (above) as the commit message.

PR #293, opened 2017-06-27
ENH: Adds full Numpydoc-compliant docstring to Actions

### LINKED ISSUE #280: add full numpydoc-compliant docstring to actions
Including `Parameters` and `Returns` sections. Depends on #278.

Kleptobismol:
Previously, calling `?` on a QIIME 2 Action (`merge_mappings`) yielded the following call docstring:

screen shot 2017-06-27 at 3 03 55 pm

There could be much more information in this docstring, so I have added functionality which creates a full Numpydoc-compliant docstring for any QIIME 2 Method. Calling `?` on the Action now yields:

screen shot 2017-06-27 at 3 03 11 pm

@jairideout
fixes #280

### DIFF
--- a/qiime2/core/testing/method.py
+++ b/qiime2/core/testing/method.py
@@ -45,8 +45,8 @@ def no_input_method() -> dict:
     return {'foo': 42}
 
 
-def long_description_method(name: 'hello world', age: int) -> dict:
-    return {'foo': 42}
+def long_description_method(mapping1:dict, name: str, age: int) -> dict:
+    return {name: age}
 
 
 def identity_with_metadata(ints: list, metadata: qiime2.Metadata) -> list:
--- a/qiime2/core/testing/method.py
+++ b/qiime2/core/testing/method.py
@@ -45,7 +45,7 @@ def no_input_method() -> dict:
     return {'foo': 42}
 
 
-def long_description_method(mapping1:dict, name: str, age: int) -> dict:
+def long_description_method(mapping1: dict, name: str, age: int) -> dict:
     return {name: age}
 
 
--- a/qiime2/core/testing/method.py
+++ b/qiime2/core/testing/method.py
@@ -45,6 +45,10 @@ def no_input_method() -> dict:
     return {'foo': 42}
 
 
+def long_description_method(name: 'hello world', age: int) -> dict:
+    return {'foo': 42}
+
+
 def identity_with_metadata(ints: list, metadata: qiime2.Metadata) -> list:
     return ints
 
--- a/qiime2/core/testing/plugin.py
+++ b/qiime2/core/testing/plugin.py
@@ -31,7 +31,7 @@ from .method import (concatenate_ints, split_ints, merge_mappings,
                      identity_with_optional_metadata,
                      identity_with_optional_metadata_category,
                      params_only_method, no_input_method,
-                     long_description_method)
+                     optional_artifacts_method, long_description_method)
 from .visualizer import (most_common_viz, mapping_viz, params_only_viz,
                          no_input_viz)
 
--- a/qiime2/core/testing/plugin.py
+++ b/qiime2/core/testing/plugin.py
@@ -179,7 +179,7 @@ dummy_plugin.methods.register_function(
     ],
     output_descriptions={
         'out': ("This is a very long description. If asked about its length,"
-                 " I would have to say it is greater than 79 characters.")
+                " I would have to say it is greater than 79 characters.")
     },
     name="Long Description",
     description=("This is a very long description. If asked about its length,"
--- a/qiime2/core/testing/plugin.py
+++ b/qiime2/core/testing/plugin.py
@@ -125,10 +125,15 @@ dummy_plugin.methods.register_function(
         'mapping1': Mapping,
         'mapping2': Mapping
     },
+    input_descriptions={
+        'mapping1': 'Mapping object to be merged'
+    },
     parameters={},
     outputs=[
         ('merged_mapping', Mapping)
     ],
+    output_descriptions={
+        'merged_mapping': 'Resulting merged Mapping object'},
     name='Merge mappings',
     description='This method merges two mappings into a single new mapping. '
                 'If a key is shared between mappings and the values differ, '
--- a/qiime2/core/testing/plugin.py
+++ b/qiime2/core/testing/plugin.py
@@ -30,7 +30,8 @@ from .method import (concatenate_ints, split_ints, merge_mappings,
                      identity_with_metadata, identity_with_metadata_category,
                      identity_with_optional_metadata,
                      identity_with_optional_metadata_category,
-                     params_only_method, no_input_method)
+                     params_only_method, no_input_method,
+                     long_description_method)
 from .visualizer import (most_common_viz, mapping_viz, params_only_viz,
                          no_input_viz)
 
@@ -153,6 +154,24 @@ dummy_plugin.methods.register_function(
     ],
     name='Identity',
     description='This method does nothing, but takes metadata'
+)
+
+dummy_plugin.methods.register_function(
+    function=long_description_method,
+    inputs={},
+    parameters={
+        'name': qiime2.plugin.Str,
+        'age': qiime2.plugin.Int
+    },
+    parameter_descriptions={
+        'name': """This is a very long description. If asked about its length,\
+ I would have to say it is greater than 79 characters."""
+    },
+    outputs=[
+        ('out', Mapping)
+    ],
+    name="Long Description",
+    description="This method does nothing, but includes a long annotation"
 )
 
 dummy_plugin.methods.register_function(
--- a/qiime2/core/testing/plugin.py
+++ b/qiime2/core/testing/plugin.py
@@ -158,20 +158,32 @@ dummy_plugin.methods.register_function(
 
 dummy_plugin.methods.register_function(
     function=long_description_method,
-    inputs={},
+    inputs={
+        'mapping1': Mapping
+    },
+    input_descriptions={
+        'mapping1': ("This is a very long description. If asked about its "
+                     "length, I would have to say it is greater than 79 "
+                     "characters.")
+    },
     parameters={
         'name': qiime2.plugin.Str,
         'age': qiime2.plugin.Int
     },
     parameter_descriptions={
-        'name': """This is a very long description. If asked about its length,\
- I would have to say it is greater than 79 characters."""
+        'name': ("This is a very long description. If asked about its length,"
+                 " I would have to say it is greater than 79 characters.")
     },
     outputs=[
         ('out', Mapping)
     ],
+    output_descriptions={
+        'out': ("This is a very long description. If asked about its length,"
+                 " I would have to say it is greater than 79 characters.")
+    },
     name="Long Description",
-    description="This method does nothing, but includes a long annotation"
+    description=("This is a very long description. If asked about its length,"
+                 " I would have to say it is greater than 79 characters.")
 )
 
 dummy_plugin.methods.register_function(
--- a/qiime2/sdk/action.py
+++ b/qiime2/sdk/action.py
@@ -11,6 +11,7 @@ import concurrent.futures
 import inspect
 import os.path
 import tempfile
+import textwrap
 
 import decorator
 
@@ -283,45 +284,54 @@ class Action(metaclass=abc.ABCMeta):
         return annotations
 
     def _build_numpydoc(self):
-        numpydoc = ""
-        numpydoc += "{}\n\n".format(self.name)
-        numpydoc += "{}\n\n".format(self.description)
+        numpydoc = []
+        numpydoc.append(textwrap.fill(self.name, width=79))
+        numpydoc.append(textwrap.fill(self.description, width=79))
 
         sig = self.signature
         params = True if len(sig.inputs) != 0 else False
 
-        numpydoc += "{}".format(self._build_section(
-            "Parameters", sig.inputs))
-        numpydoc += "{}".format(self._build_section(
-            "Parameters", sig.parameters, parameters=params))
-        numpydoc += ("\n" if len(sig.parameters) != 0 or
-                     len(sig.inputs) != 0 else "")
-        numpydoc += "{}\n".format(self._build_section(
-            "Returns", sig.outputs))
+        inputs = self._build_section("Parameters", sig.inputs)
+        parameters = self._build_section("Parameters", sig.parameters,
+                                         parameters=params)
+        returns = self._build_section("Returns", sig.outputs)
 
-        return numpydoc
+        if len(inputs) != 0 and len(parameters) != 0:
+            parameters = '\n'.join([inputs, parameters])
+            inputs = None
+
+        if inputs is not None:
+            for section in [inputs, parameters, returns]:
+                if len(section) != 0:
+                    numpydoc.append(section)
+        else:
+            numpydoc.append(parameters)
+            numpydoc.append(returns)
+
+        return '\n\n'.join(numpydoc)
 
     def _build_section(self, header, iterable, parameters=False):
-        section = ""
+        section = []
 
         if len(iterable) != 0:
             if not parameters:
-                section += "{}\n".format(header)
-                section += "{}\n".format(''.join(['-' for char in header]))
+                section.append(header)
+                section.append('-'*len(header))
             for item in iterable:
                 item_obj = iterable[item]
-                section += "{}: {}".format(item, item_obj.qiime_type)
-                if type(item_obj.default).__name__ != "_NoValue":
-                    section += ", optional\n"
-                else:
-                    section += "\n"
-                if type(item_obj.description).__name__ == "_NoValue":
-                    description = "No description available"
-                else:
-                    description = item_obj.description
-                section += "    {}\n".format(description)
+                info = {'item': item, 'type': item_obj.qiime_type}
+                section.append("{item} : {type}".format(**info))
+                if item_obj.has_default():
+                    section[-1] += ", optional"
+                if str(item_obj.description) != "NOVALUE":
+                    section.append("    {}".format(item_obj.description))
 
-        return section
+        for idx in range(len(section)):
+            section[idx] = textwrap.fill(section[idx], width=79)
+        if len(section) != 0:
+            return '\n'.join(section).strip()
+        else:
+            return ""
 
     def _is_subprocess(self):
         return self._pid != os.getpid()
--- a/qiime2/sdk/action.py
+++ b/qiime2/sdk/action.py
@@ -288,11 +288,12 @@ class Action(metaclass=abc.ABCMeta):
         numpydoc += "{}\n\n".format(self.description)
 
         sig = self.signature
+        params = True if len(sig.inputs) != 0 else False
 
         numpydoc += "{}".format(self._build_section(
             "Parameters", sig.inputs))
         numpydoc += "{}".format(self._build_section(
-            "Parameters", sig.parameters, parameters=True))
+            "Parameters", sig.parameters, parameters=params))
         numpydoc += ("\n" if len(sig.parameters) != 0 or
                      len(sig.inputs) != 0 else "")
         numpydoc += "{}\n".format(self._build_section(
--- a/qiime2/sdk/action.py
+++ b/qiime2/sdk/action.py
@@ -278,8 +278,8 @@ class Action(metaclass=abc.ABCMeta):
 
     def _build_numpydoc(self):
         numpydoc = []
-        numpydoc.append(textwrap.fill(self.name, width=79))
-        numpydoc.append(textwrap.fill(self.description, width=79))
+        numpydoc.append(textwrap.fill(self.name, width=75))
+        numpydoc.append(textwrap.fill(self.description, width=75))
 
         sig = self.signature
         params = collections.OrderedDict()
@@ -302,18 +302,16 @@ class Action(metaclass=abc.ABCMeta):
             section.append(header)
             section.append('-'*len(header))
             for key, value in iterable.items():
-                section.append(
+                variable_line = (
                     "{item} : {type}".format(item=key, type=value.qiime_type))
                 if value.has_default():
-                    section[-1] += ", optional"
+                    variable_line += ", optional"
+                section.append(variable_line)
                 if value.has_description():
                     section.append(textwrap.indent(textwrap.fill(
-                        str(value.description), width=79), '    '))
+                        str(value.description), width=71), '    '))
 
-        if section:
-            return '\n'.join(section).strip()
-        else:
-            return ""
+        return '\n'.join(section).strip()
 
     def _is_subprocess(self):
         return self._pid != os.getpid()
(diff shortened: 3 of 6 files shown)


Kleptobismol:
@ebolyen, that last testcase has been added, ready for review

thermokarst:
@Kleptobismol: can you please take a look at the [failing Travis](https://travis-ci.org/qiime2/qiime2/builds/248863641) builds?

Kleptobismol:
in the process of fixing @thermokarst

Kleptobismol:
There is something very weird occurring with this PR, which only occurs when I test the docs of the method params_only_method.

thermokarst:
@Kleptobismol, I am not super familiar with this code, but it looks like this isn't working because your aren't checking the docstring on the wrapped method, just the vanilla import. Take a look at [this line](https://github.com/qiime2/qiime2/pull/293/files#diff-852ef215d2c7f81c5de661dd98f06b14L431), where you are pulling in a visualizer through the plugin API.

```python
params_only_method = self.plugin.methods['params_only_method']
params_only = params_only_method.__call__.__doc__
self.assertEqual(exp_params_only, params_only)
```

I am not sure if this is the correct way to go about this, just an idea.

Kleptobismol:
@ebolyen ready for review

qiime2/sdk/tests/test_method.py
```diff
@@ -444,14 +450,102 @@ def test_async_with_multiple_outputs(self):
...
+
+exp_merge_calldoc = """\
+Merge mappings
+
+This method merges two mappings into a single new mapping. If a key is \
+shared between mappings and the values differ, an error will be raised.
+
+Parameters
+----------
+mapping1: Mapping
```
ebolyen: numpydoc requires a space between the parameter name and type:
```
mapping1 : Mapping
```

qiime2/sdk/tests/test_method.py
```diff
@@ -444,14 +450,102 @@ def test_async_with_multiple_outputs(self):
...
+    Mapping object to be merged
+mapping2: Mapping
+    No description available
```
ebolyen: Descriptions aren't required in numpydoc, so it's probably better to just omit them if not set by the plugin.

qiime2/sdk/action.py
```diff
@@ -284,6 +282,47 @@ def _build_annotations(self):
...
+        section = ""
+
+        if len(iterable) != 0:
+            if not parameters:
+                section += "{}\n".format(header)
+                section += "{}\n".format(''.join(['-' for char in header]))
+            for item in iterable:
+                item_obj = iterable[item]
+                section += "{}: {}".format(item, item_obj.qiime_type)
+                if type(item_obj.default).__name__ != "_NoValue":
```
ebolyen: You can should use `.has_default()` instead.  Knowledge about the specific sentinel is really an implementation detail of the ParameterSpec object.

qiime2/sdk/action.py
```diff
@@ -284,6 +282,47 @@ def _build_annotations(self):
...
+
+        return numpydoc
+
+    def _build_section(self, header, iterable, parameters=False):
+        section = ""
+
+        if len(iterable) != 0:
+            if not parameters:
+                section += "{}\n".format(header)
+                section += "{}\n".format(''.join(['-' for char in header]))
```
ebolyen: A trick you can use:
```
 '-' * len(some_iterable) 
```

qiime2/sdk/action.py
```diff
@@ -284,6 +282,47 @@ def _build_annotations(self):
...
+
+        sig = self.signature
+        params = True if len(sig.inputs) != 0 else False
+
+        numpydoc += "{}".format(self._build_section(
+            "Parameters", sig.inputs))
+        numpydoc += "{}".format(self._build_section(
+            "Parameters", sig.parameters, parameters=params))
+        numpydoc += ("\n" if len(sig.parameters) != 0 or
+                     len(sig.inputs) != 0 else "")
```
ebolyen: If you joined segments with `"\n".join(...)` you might have an easier time.

qiime2/sdk/action.py
```diff
@@ -284,6 +282,47 @@ def _build_annotations(self):
 
         return annotations
 
+    def _build_numpydoc(self):
+        numpydoc = ""
+        numpydoc += "{}\n\n".format(self.name)
+        numpydoc += "{}\n\n".format(self.description)
```
ebolyen: These descriptions may be longer than 75 characters, you should use `textwrap` here.

ebolyen: This is true of all of the descriptions (param, etc), but `textwrap` also understands indenting, which will make things easier.

qiime2/sdk/tests/test_method.py
```diff
@@ -444,14 +450,102 @@ def test_async_with_multiple_outputs(self):
...
+
+        params_only = params_only_method.__call__.__doc__
+        self.assertEqual(exp_params_only, params_only)
+
+
+exp_merge_calldoc = """\
+Merge mappings
+
+This method merges two mappings into a single new mapping. If a key is \
+shared between mappings and the values differ, an error will be raised.
```
ebolyen: This line is too long, it should be wrapped.

qiime2/sdk/action.py
```diff
@@ -284,6 +282,47 @@ def _build_annotations(self):
...
+    def _build_section(self, header, iterable, parameters=False):
+        section = ""
+
+        if len(iterable) != 0:
+            if not parameters:
+                section += "{}\n".format(header)
+                section += "{}\n".format(''.join(['-' for char in header]))
+            for item in iterable:
+                item_obj = iterable[item]
+                section += "{}: {}".format(item, item_obj.qiime_type)
```
ebolyen: I've never seen format used without keys. I would either add some keys or use string interpolation (`%`). Either is fine.

Also, you should probably use the `repr` call on the QIIME Type, it's closer to what you need semantically (the results are the same at present however). This can be done with `{!r}` or `%r` depending on the interpolation/formatting syntax you go with.

qiime2/sdk/action.py
```diff
@@ -284,6 +282,47 @@ def _build_annotations(self):
...
+    def _build_numpydoc(self):
+        numpydoc = ""
+        numpydoc += "{}\n\n".format(self.name)
+        numpydoc += "{}\n\n".format(self.description)
+
+        sig = self.signature
+        params = True if len(sig.inputs) != 0 else False
+
+        numpydoc += "{}".format(self._build_section(
+            "Parameters", sig.inputs))
```
ebolyen: Format isn't doing anything here. It is already a string.

qiime2/sdk/action.py
ebolyen: Calling `type(stuff).__anythinghere__` should be a hint that you are breaking someone's abstractions. Sometimes they are bad abstractions, or you are doing something very low level, but it should give you pause.

qiime2/sdk/tests/test_method.py
ebolyen: To clarify, it is the string object that is too long on that line, not the source representation itself.

qiime2/sdk/tests/test_method.py
Kleptobismol: done

qiime2/sdk/tests/test_method.py
Kleptobismol: done

qiime2/sdk/tests/test_method.py
Kleptobismol: done

qiime2/sdk/action.py
Kleptobismol: done

qiime2/sdk/action.py
Kleptobismol: fixed

qiime2/sdk/action.py
Kleptobismol: done

qiime2/sdk/action.py
Kleptobismol: done

Kleptobismol: fixed both occurrences of this mistake

qiime2/sdk/action.py
Kleptobismol: done

qiime2/sdk/tests/test_method.py
```diff
@@ -523,9 +516,7 @@ def test_docstring(self):
 
 Returns
 -------
-out: Mapping
-    No description available
-
+out : Mapping\
```
ebolyen: There should be a trailing empty line (I'm not 100% if that's the spec, but that's our convention that we use):
```
"""Docstring start
.... stuff ....
   <- single empty line
"""
```

qiime2/sdk/action.py
```diff
@@ -283,45 +284,54 @@ def _build_annotations(self):
...
     def _build_numpydoc(self):
-        numpydoc = ""
-        numpydoc += "{}\n\n".format(self.name)
-        numpydoc += "{}\n\n".format(self.description)
+        numpydoc = []
+        numpydoc.append(textwrap.fill(self.name, width=79))
+        numpydoc.append(textwrap.fill(self.description, width=79))
 
         sig = self.signature
         params = True if len(sig.inputs) != 0 else False
```
ebolyen: Missed this earlier, but this is just `bool(sig.inputs)`. In Python things that can be described as empty are also "falsey" (technical term) which means that in the context of a boolean expression they will be evaluated with respect to their truthy-ness and falsey-ness.

qiime2/sdk/action.py
```diff
@@ -283,45 +284,54 @@ def _build_annotations(self):
...
-            "Parameters", sig.inputs))
-        numpydoc += "{}".format(self._build_section(
-            "Parameters", sig.parameters, parameters=params))
-        numpydoc += ("\n" if len(sig.parameters) != 0 or
-                     len(sig.inputs) != 0 else "")
-        numpydoc += "{}\n".format(self._build_section(
-            "Returns", sig.outputs))
+        inputs = self._build_section("Parameters", sig.inputs)
+        parameters = self._build_section("Parameters", sig.parameters,
+                                         parameters=params)
```
ebolyen: The keyword might be better as `include_header=bool(inputs)`

qiime2/sdk/action.py
```diff
@@ -283,45 +284,54 @@ def _build_annotations(self):
...
+        returns = self._build_section("Returns", sig.outputs)
 
-        return numpydoc
+        if len(inputs) != 0 and len(parameters) != 0:
+            parameters = '\n'.join([inputs, parameters])
+            inputs = None
+
+        if inputs is not None:
```
ebolyen: The flow control between this block and the above `if` is not terribly clear, it's very much like a state-machine. You might have better luck merging `sig.inputs` and `sig.parameters` (they are both `OrderedDict` objects) and providing that to `_build_section`. Then the presence/absence of them isn't as interesting because you'll either have an empty section or you wont.

qiime2/sdk/action.py
ebolyen: This comes up in a lot of places `if len(thing) != 0:` you can replace that with just `if thing:`

qiime2/sdk/action.py
```diff
@@ -283,45 +284,54 @@ def _build_annotations(self):
...
-                    section += "\n"
-                if type(item_obj.description).__name__ == "_NoValue":
-                    description = "No description available"
-                else:
-                    description = item_obj.description
-                section += "    {}\n".format(description)
-
-        return section
+                info = {'item': item, 'type': item_obj.qiime_type}
+                section.append("{item} : {type}".format(**info))
```
ebolyen: Sorry, I confused you here (although your use of kwargs *is* correct), because of the variadic signature you can just write:
```python
"... stuff ...".format(name=key, type=value.qiime_type)
```

`format` internally has a definition like:
```python
def format(*args, **kwargs):
   ...
````
which is what makes what you did and the above snippet possible.

File: qiime2/sdk/action.py [hunk 15]
```diff
@@ -283,45 +284,54 @@ def _build_annotations(self):
...
-        section = ""
+        section = []
 
         if len(iterable) != 0:
             if not parameters:
-                section += "{}\n".format(header)
-                section += "{}\n".format(''.join(['-' for char in header]))
+                section.append(header)
+                section.append('-'*len(header))
             for item in iterable:
```
```python
for key, value in iterable.items():
```
(or any other variable name)

qiime2/sdk/action.py
```diff
@@ -283,45 +284,54 @@ def _build_annotations(self):
...
-                else:
-                    description = item_obj.description
-                section += "    {}\n".format(description)
-
-        return section
+                info = {'item': item, 'type': item_obj.qiime_type}
+                section.append("{item} : {type}".format(**info))
+                if item_obj.has_default():
+                    section[-1] += ", optional"
+                if str(item_obj.description) != "NOVALUE":
```
ebolyen: use `.has_description()`. 

Clever use of the (implicit) repr for the sentinel... I'm going to have to add the memory address to the end of that repr to prevent this.

qiime2/sdk/action.py
```diff
@@ -283,45 +284,54 @@ def _build_annotations(self):
...
+                    section.append("    {}".format(item_obj.description))
```
ebolyen: This will need textwrap as well, but luckily it can also indent for you.

qiime2/sdk/action.py
```diff
@@ -283,45 +284,54 @@ def _build_annotations(self):
...
+
+        for idx in range(len(section)):
+            section[idx] = textwrap.fill(section[idx], width=79)
```
ebolyen: text-wrapping the entire section isn't going to work because you have indentation. You'll need to text-wrap each section individually and then merge them.

ebolyen: I would recommend creating a method with really long descriptions everywhere so that you can write a test to confirm that each wrapping respects its indent-level.

qiime2/sdk/action.py
ebolyen: (it will also mean you don't need the `include_header` (`parameters` currently) flag)

qiime2/sdk/tests/test_method.py
Kleptobismol: done

qiime2/sdk/action.py
Kleptobismol: done

Kleptobismol: fixed elsewhere as well

qiime2/sdk/action.py
Kleptobismol: done

qiime2/sdk/action.py
Kleptobismol: done

qiime2/sdk/action.py
Kleptobismol: done

qiime2/sdk/action.py
Kleptobismol: done

qiime2/sdk/action.py
Kleptobismol: done

qiime2/sdk/action.py
Kleptobismol: @ebolyen are there any qiime objects with really long descriptions? because otherwise i'd have to make a custom qiime object. to clarify, by qiime objects, I mean something like IntSequence1, MetadataCategory, etc.

ebolyen: It isn't the type that has the description, but rather the parameter which is documented at registration of a method. You should create another simplistic method (doesn't need to do much) which is annotated with relatively long (>80 characters) descriptions.

Kleptobismol: thanks for clarifying

Kleptobismol: done

Kleptobismol:
@ebolyen ready for final pass

qiime2/core/testing/method.py
```diff
@@ -45,6 +45,10 @@ def no_input_method() -> dict:
     return {'foo': 42}
 
 
+def long_description_method(name: 'hello world', age: int) -> dict:
+    return {'foo': 42}
```
ebolyen: Maybe return `{name: age}` instead?

ebolyen: that way this "kind of" does something

qiime2/core/testing/plugin.py
```diff
@@ -152,6 +157,24 @@
...
 dummy_plugin.methods.register_function(
+    function=long_description_method,
+    inputs={},
+    parameters={
+        'name': qiime2.plugin.Str,
+        'age': qiime2.plugin.Int
+    },
+    parameter_descriptions={
+        'name': """This is a very long description. If asked about its length,\
+ I would have to say it is greater than 79 characters."""
```
ebolyen: Try using `("implicit string" "concatenation" "instead")` then you don't need the newline escape (`\`)

qiime2/core/testing/plugin.py
```diff
@@ -152,6 +157,24 @@
...
+    },
+    outputs=[
+        ('out', Mapping)
+    ],
+    name="Long Description",
+    description="This method does nothing, but includes a long annotation"
```
ebolyen: Make this description wrap as well (you can use lorem ipsum if that helps). Also add input and output descriptions. It's kind of a waste of a good method otherwise.

qiime2/sdk/action.py
```diff
@@ -276,6 +276,45 @@ def _build_annotations(self):
 
         return annotations
 
+    def _build_numpydoc(self):
+        numpydoc = []
+        numpydoc.append(textwrap.fill(self.name, width=79))
```
ebolyen: the width should actually be `75` and for parameter/returns descriptions I think they wrap at `71` because they're indented.

qiime2/sdk/action.py
```diff
@@ -276,6 +276,45 @@ def _build_annotations(self):
...
+        numpydoc.append(textwrap.fill(self.description, width=79))
+
+        sig = self.signature
+        params = collections.OrderedDict()
+        params.update(sig.inputs)
+        params.update(sig.parameters)
```
ebolyen: :+1:

qiime2/sdk/action.py
```diff
@@ -276,6 +276,45 @@ def _build_annotations(self):
...
+                if value.has_default():
+                    section[-1] += ", optional"
+                if value.has_description():
+                    section.append(textwrap.indent(textwrap.fill(
+                        str(value.description), width=79), '    '))
+
+        if section:
+            return '\n'.join(section).strip()
+        else:
+            return ""
```
ebolyen: This branch (and so the entire conditional) isn't necessary, the above will be an empty string if `section` is empty.

qiime2/sdk/action.py
```diff
@@ -276,6 +276,45 @@ def _build_annotations(self):
...
+            section.append(header)
+            section.append('-'*len(header))
+            for key, value in iterable.items():
+                section.append(
+                    "{item} : {type}".format(item=key, type=value.qiime_type))
+                if value.has_default():
+                    section[-1] += ", optional"
+                if value.has_description():
+                    section.append(textwrap.indent(textwrap.fill(
+                        str(value.description), width=79), '    '))
```
ebolyen: You can provide an `indent` to fill.

qiime2/sdk/action.py
```diff
@@ -276,6 +276,45 @@ def _build_annotations(self):
...
+        section = []
+
+        if iterable:
+            section.append(header)
+            section.append('-'*len(header))
+            for key, value in iterable.items():
+                section.append(
+                    "{item} : {type}".format(item=key, type=value.qiime_type))
+                if value.has_default():
+                    section[-1] += ", optional"
```
ebolyen: instead of reindexing the list, how about you append the finished string.

```python
figure_out_a_name = "{item} : {type}".format(item=key, type=value.qiime_type))
if value.has_default():
    figure_out_a_name += ", optional"
section.append(figure_out_a_name)
```

qiime2/sdk/action.py
```diff
@@ -276,6 +276,45 @@ def _build_annotations(self):
...
+
+    def _build_section(self, header, iterable):
+        section = []
+
+        if iterable:
+            section.append(header)
+            section.append('-'*len(header))
+            for key, value in iterable.items():
+                section.append(
+                    "{item} : {type}".format(item=key, type=value.qiime_type))
```
ebolyen: technically this has the ability to exceed 75 characters, but I'm ignoring that because I can't find anything about how one would wrap in that case anyways

@jairideout do you have any idea here?

qiime2/sdk/tests/test_visualizer.py
```diff
@@ -431,11 +431,6 @@ def test_docstring(self):
         visualizer = self.plugin.visualizers['mapping_viz']
         exp = "QIIME 2 Visualizer"
         self.assertEqual(exp, visualizer.__doc__)
-        obs = visualizer.__call__.__doc__.split('\n')
-        self.assertTrue(obs[0].startswith('Visualize two'))
-        self.assertEqual(len(obs[1]), 0)
-        self.assertTrue(obs[2].startswith('This visualizer produces an HTML'))
-        self.assertTrue(obs[3].startswith('sorted in alphabetical'))
```
ebolyen: I think you should write a test or two for the visualizers as well.

qiime2/sdk/action.py
jairideout: I'm not sure how to handle that in the numpydoc standard, fine to ignore

qiime2/core/testing/method.py
Kleptobismol: done

qiime2/core/testing/plugin.py
Kleptobismol: done

qiime2/core/testing/plugin.py
Kleptobismol: done

qiime2/sdk/action.py
Kleptobismol: done

qiime2/sdk/action.py
Kleptobismol: 👍

qiime2/sdk/action.py
Kleptobismol: done

qiime2/sdk/action.py
Kleptobismol: if I do that, any line that is wrapped is not indented

qiime2/sdk/action.py
Kleptobismol: done

qiime2/sdk/tests/test_visualizer.py
Kleptobismol: done

qiime2/sdk/action.py
ebolyen: Sorry I was almost certain that that keyword existed, but it doesn't seem to.

ebolyen:
Thanks @Kleptobismol :tada:

PR #303, opened 2017-07-11
BUG: Primitive collections had double constructor

ebolyen:


qiime2/core/type/primitive.py
```diff
@@ -82,8 +82,6 @@ def __init__(self, is_element, encode, decode, *args, **kwargs):
 
         super().__init__(*args, **kwargs)
 
-        super().__init__(*args, **kwargs)
-
```
ebolyen: ¯\\\_(ツ)\_/¯

PR #314, opened 2017-08-08
Metadata.filter

gregcaporaso:


jairideout:
Thanks @gregcaporaso! Nice tests!

PR #333, opened 2017-09-08
BUG: add output-name to provenance

### LINKED ISSUE #325: Provenance should indicate the output name according to the action
This is a hole right now which means if an action returns multiple artifacts of the same type, they cannot be disambiguated from the provenance.

ebolyen:
fixes #325

### DIFF
--- a/qiime2/sdk/action.py
+++ b/qiime2/sdk/action.py
@@ -348,14 +348,15 @@ class Method(Action):
                 % (len(output_views), len(output_types)))
 
         output_artifacts = []
-        for output_view, spec in zip(output_views, output_types.values()):
+        for output_view, (name, spec) in zip(output_views,
+                                             output_types.items()):
             if type(output_view) is not spec.view_type:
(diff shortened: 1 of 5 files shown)


qiime2/sdk/action.py
```diff
@@ -394,6 +395,7 @@ def _callable_executor_(self, callable, view_args, output_types,
                 raise TypeError(
                     "Visualizer %r should not return anything. "
                     "Received %r as a return value." % (self, ret_val))
+            provenance.output_name = 'visualization'
```
ebolyen: I'm not sold on this here. Using `fork()` when there's only one result feels silly, but so does this a little bit.

Would assignment of `provenance.fork(name)` to `provenance` then using `provenance.output_name` be a better move for the Method version of this? Then they would match.

thermokarst: I think it is clear as implemented what is going on here. That said, I think would be fine with either approach.

PR #337, opened 2017-09-25
ENH: adds validate protocol to plugin.model

ebolyen:
Still needs some tests for each branch (and current testing formats should be updated)


qiime2/core/testing/format.py
```diff
@@ -100,6 +101,19 @@ def single_ints_path_maker(self, num):
             return 'file%d.txt' % num
 
 
+class RedundantSingleIntDirectoryFormat(model.DirectoryFormat):
```
jairideout: :1st_place_medal: Directory Format MVP 2017 :1st_place_medal:

PR #348, opened 2017-10-11
ENH: adds basic Pipeline actions

ebolyen:
TODO:
  - [x] tighten up registration w.r.t. view annotations (or lack thereof)
  - [x] write tests
  - [x] actual lifetime management of artifacts
  - [x] write more tests

Pipelines work otherwise! And cleanup is handled even when calling with `.async`

### DIFF
--- a/qiime2/core/archive/tests/test_provenance.py
+++ b/qiime2/core/archive/tests/test_provenance.py
@@ -13,7 +13,7 @@ import pandas.util.testing as pdt
 
 import qiime2
 from qiime2.plugins import dummy_plugin
-from qiime2.core.testing.type import IntSequence1
+from qiime2.core.testing.type import IntSequence1, Mapping
 
 
 class TestProvenanceIntegration(unittest.TestCase):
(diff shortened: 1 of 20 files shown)


ebolyen:
An update on this: artifact lifetimes (or at least the artifact archivers/backing directories) are getting more complicated that expected when an async call of a pipeline fails. I need to overhaul the lifetime management again (3rd times the charm?).

ebolyen:
QIIME 2 has a call tree/stack now which keeps a scope of the artifacts/visualizations/provenance within.

Also there is a more primitive operation on actions called `.bind` which will return a callable if you hand it a context factory. This will be a useful lower-level SDK API, and should come in handy for parallelism as well.

qiime2/core/archive/tests/test_provenance.py
```diff
@@ -176,6 +176,25 @@ def test_no_output_name_import(self):
...
+        with (new_mapping_p_dir / 'action' / 'action.yaml').open() as fh:
+            new_mapping_yaml = fh.read()
+
+        # Basic sanity check
+        self.assertIn('type: pipeline', new_mapping_yaml)
+        self.assertIn('int_sequence: %s' % ints.uuid, new_mapping_yaml)
+        self.assertIn('mapping: %s' % mapping.uuid, new_mapping_yaml)
+        # Remembers the original mapping uuid
+        self.assertIn('alias-of: %s' % mapping.uuid, new_mapping_yaml)
+
```
ebolyen: add test for nested

qiime2/sdk/action.py
```diff
@@ -412,6 +405,56 @@ def _init(cls, callable, inputs, parameters, package, name, description,
...
+
+        if len(outputs) != len(output_types):
+            raise TypeError(
+                "Number of outputs must match number of output "
+                "semantic types: %d != %d"
+                % (len(outputs), len(output_types)))
+
+        results = []
+        for output, (name, spec) in zip(outputs, output_types.items()):
+            if not (output.type <= spec.qiime_type):
```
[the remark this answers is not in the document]
ebolyen: Turns out the errors are more informative if arity is checked before types, as an arity mismatch is usually also a type mismatch.

qiime2/core/archive/tests/test_provenance.py
ebolyen: done

PR #349, opened 2017-10-12
Added ValueError to Choices Constructor to prevent empty values in Ch…

patthehat033:
I raised an error in the Choices Constructor that says "Choices must have a minimum of one value."

### DIFF
--- a/qiime2/core/type/primitive.py
+++ b/qiime2/core/type/primitive.py
@@ -190,7 +190,8 @@ class Choices(Predicate):
     def __init__(self, choices):
 
         if not choices:
-            raise ValueError("'Choices' cannot be instantiated with an empty set.")
+            raise ValueError("'Choices' cannot be instantiated with an empty"
+                             " set.")
 
         self.choices = set(choices)
 
--- a/qiime2/core/type/primitive.py
+++ b/qiime2/core/type/primitive.py
@@ -188,6 +188,10 @@ class Range(Predicate):
 
 class Choices(Predicate):
     def __init__(self, choices):
+
+        if not choices:
+            raise ValueError("Choices must have a minimum of one value.")
+
         self.choices = set(choices)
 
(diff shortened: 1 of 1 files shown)


qiime2/core/type/primitive.py
```diff
@@ -188,6 +188,10 @@ def to_ast(self):
 
 class Choices(Predicate):
     def __init__(self, choices):
+
+        if not choices:
+            raise ValueError("Choices must have a minimum of one value.")
```
ebolyen: This is a pretty reasonable error, but maybe make it a little more mechanical in this case as our audience here are plugin developers (who are more or less familiar with Python).
```
'Choices' cannot be instantiated with an empty set.
```
The single-quotes are commonly used in Python exceptions to disambiguate words from actual objects/variables.

qiime2/core/type/primitive.py
```diff
@@ -188,6 +188,10 @@ def to_ast(self):
 
 class Choices(Predicate):
     def __init__(self, choices):
+
+        if not choices:
+            raise ValueError("'Choices' cannot be instantiated with an empty set.")
```
ebolyen: [Looks like `flake8` found an issue](https://travis-ci.org/qiime2/qiime2/builds/291293905?utm_source=github_status&utm_medium=notification). It checks compliance with [PEP 8](https://www.python.org/dev/peps/pep-0008/). We use style guides like these because they can be automatically checked, and it prevents fights over formatting.

In this case you can use Python's implicit string concatenation:
```python
raise ValueError("'Choices' cannot be instantiated with an empty"
                 " set.")
```

Python will mash those together because they are part of the same expression. Just be sure to mind the space (` `) otherwise you'll have two words mashed together as well.

You can also check locally by running `make lint`.

patthehat033:
No problem, @ebolyen! Sounds perfect!

PR #356, opened 2017-11-09
ENH: adds collection types

ebolyen:
`Set` and `List` are no longer primitive types, they are collection types. This
means they can be used on primitives (with the exception of metadata) or
semantic types. This enables variadic input artifacts in the framework.

~~**This is still under-tested. I'm going to be making changes to the CLI to make sure this works before returning to this PR**~~

### DIFF
--- a/qiime2/core/testing/method.py
+++ b/qiime2/core/testing/method.py
@@ -78,3 +78,17 @@ def optional_artifacts_method(ints: list, num1: int, optional1: list=None,
     if num2 is not None:
         result += [num2]
     return result
+
+
+def variadic_input_method(ints: list, int_set: int, nums: int,
+                          opt_nums: int=None) -> list:
+    results = []
+
+    for int_list in ints:
+        results += int_list
+    results += sorted(int_set)
+    results += nums
+    if opt_nums:
+        results += opt_nums
+
+    return results
--- a/qiime2/core/type/tests/test_collection.py
+++ b/qiime2/core/type/tests/test_collection.py
@@ -1,0 +1,153 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016-2017, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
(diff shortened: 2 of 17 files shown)


ebolyen:
Ready for review, as this PR is complete, however the CLI is being a pain, so for the sake of not blocking integration tests, this shouldn't be merged until the CLI is ready.

qiime2/core/testing/method.py
```diff
@@ -78,3 +78,17 @@ def optional_artifacts_method(ints: list, num1: int, optional1: list=None,
     if num2 is not None:
         result += [num2]
     return result
+
+
+def variadic_input_method(ints: list, int_set: int, nums: int,
```
thermokarst: Why doesn't the type annotation in the signature convey the collection type?

```python
from typing import List, Set

def variadic_input_methods(ints: List[list], int_set: Set[int], nums: Set[int], opt_nums: List[int]=None):
    ...
```

ebolyen: The only reason is that `typing.List` doesn't pickle well. This can be fixed outside of Python's stdlib, but we've never really sat down to take care of it. (This is also why we haven't bothered with `typing.Union` which would also be **super** handy.)

qiime2/core/type/tests/test_collection.py
```diff
@@ -0,0 +1,153 @@
...
+        self.assertFalse(Set[Int].is_concrete())
+
+    def test_to_ast_semantic(self):
+        Foo = SemanticType('Foo')
+
+        ast = List[Foo].to_ast()
+        self.assertEqual(ast['fields'][0], Foo.to_ast())
+        self.assertEqual(ast['type'], 'collection')
+
+    def test_to_ast_primtive(self):
```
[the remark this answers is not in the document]
ebolyen: oops!

qiime2/core/testing/method.py
thermokarst: Makes sense, thanks for the reminder!

PR #364, opened 2018-01-23
MAINT: remove case-insensitive duplicates check for IDs and column names

jairideout:
Discussed with @antgonza and @gregcaporaso. The requirement that IDs or column names cannot contain case-insensitive duplicates may be too restrictive (@antgonza had alleles as column names in his example).

The code now performs exact string matching when checking for duplicates on either axis.

### DIFF
--- a/qiime2/metadata/metadata.py
+++ b/qiime2/metadata/metadata.py
@@ -135,10 +135,12 @@ class _MetadataBase:
                     "Detected metadata %s that conflicts with a name reserved "
                     "for ID headers: %r" % (label, value))
 
-        # TODO use qiime2.core.util.find_duplicates in error message
-        if len(index) != len(set(e.lower() for e in index)):
+        if len(index) != len(set(index)):
+            duplicates = find_duplicates(index)
             raise ValueError(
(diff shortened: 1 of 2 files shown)


qiime2/metadata/metadata.py
```diff
@@ -135,10 +135,12 @@ def _validate_pandas_index(self, index, label):
...
 
-        # TODO use qiime2.core.util.find_duplicates in error message
-        if len(index) != len(set(e.lower() for e in index)):
+        if len(index) != len(set(index)):
+            duplicates = find_duplicates(index)
             raise ValueError(
-                "Detected duplicate metadata %s(s) (case-insensitive)" % label)
+                "Metadata %ss must be unique. The following %ss are "
+                "duplicated: %s" %
+                (label, label, ', '.join(repr(e) for e in sorted(duplicates))))
```
antgonza: I know it's not that trivial and this should not be a blocker but what about also adding the column numbers where those labels are found?

jairideout: That's a good idea, I'll put it on my TODO list. If it's not something I get to in the near-term (i.e. within the next week as I wrap up the Metadata project), I'll create an issue so we can track this feature.

PR #366, opened 2018-01-25
ENH: remove Metadata.from_artifact in favor of Artifact.view(Metadata)

jairideout:
Removed `Metadata.from_artifact` in favor of `Artifact.view(Metadata)`, which now tracks the source artifact (important for provenance).

Discussed with @ebolyen; there's no need to have two APIs that do the same thing, and using the view API is preferred since `Metadata` doesn't need to be special-cased.

Related to discussion in https://github.com/qiime2/q2-feature-table/issues/136 regarding `Artifact.view(Metadata)` not tracking the source artifact, which is now fixed.

### DIFF
--- a/qiime2/sdk/result.py
+++ b/qiime2/sdk/result.py
@@ -249,6 +249,10 @@ class Artifact(Result):
         return self._view(view_type)
 
     def _view(self, view_type, recorder=None):
+        if view_type is qiime2.Metadata and not self.has_metadata():
+            raise TypeError(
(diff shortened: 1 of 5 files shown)


qiime2/sdk/result.py
```diff
@@ -249,14 +249,18 @@ def view(self, view_type):
         return self._view(view_type)
 
     def _view(self, view_type, recorder=None):
+        if view_type is qiime2.Metadata and not self.has_metadata():
```
thermokarst: 👍

PR #383, opened 2018-03-08
ENH: adds notebook server-extension for viewing results

### LINKED ISSUE #340: Support viewing artifacts inside the Jupyter notebook
It would be wonderful if we could visualize artifacts from inside a Jupyter notebook. This presents lots of technical challenges (resource loading, unpacking, etc), but it would be valuable if we could find a way to do it.

Perhaps it is more feasible to make it happen with something like [Jupyter Lab](https://github.com/jupyterlab/jupyterlab).

ebolyen:
fixes #340 

This supports `base_url` config so it should work with JupyterHub. 
The `_html_repr_` is able to provides instructions:
  - when the extension isn't installed.
  - when the kernel has been restarted and the page reloaded (i.e. viz is out of scope)

I think this will be of interest to @ElDeveloper, @bassio, and @jakereps.

I've done a bunch of local testing, but this could use more!
To install, just run:
```bash
jupyter serverextension enable --py qiime2 --sys-prefix
```
  -or-
repr a QIIME 2 visualization which will produce the above instructions.

### DIFF
--- a/qiime2/jupyter/handlers.py
+++ b/qiime2/jupyter/handlers.py
@@ -1,0 +1,65 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016-2018, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import os
+import pathlib
+
+import tornado.web as web
+from notebook.base.handlers import IPythonHandler
+
+import qiime2.core.archive.archiver as archiver
+
+
+class _ArchiveCheck(archiver._Archive):
+    """This is only what is needed to verify a path is an archive"""
+    # TODO: make this part of the archiver API at some point
+    def open(self, relpath):
+        abspath = os.path.join(str(self.path), str(self.uuid), relpath)
+        return open(abspath, 'r')
+
+    def relative_iterdir(self, relpath='.'):
+        for p in pathlib.Path(self.path).iterdir():
+            yield str(p.relative_to(self.path))
+
+
+class QIIME2RedirectHandler(IPythonHandler):
+    """Add a location to location_store for later retrieval"""
+    def initialize(self, result_store):
+        self.result_store = result_store
+
+    def get(self):
+        location = self.get_query_argument('location')
+        if not os.path.exists(location):
+            # Client DOM should explain that the user should re-run the cell
+            self.send_error(428)  # Precondition Required
+            return
+        # is it actually a QIIME 2 result, or a random part of the filesystem
+        archive = _ArchiveCheck(pathlib.Path(location))
+        self.result_store[archive.uuid] = os.path.join(
+            location, str(archive.uuid), 'data')
+
+        self.redirect('view/%s/' % archive.uuid)
+
+
+class QIIME2ResultHandler(web.StaticFileHandler):
+    def initialize(self, path, default_filename):
+        super().initialize(path, default_filename)
+        self.result_store = path  # path is actually result_store
+
+    @classmethod
+    def get_absolute_path(cls, root, path):
+        uuid, path = path.split('/', 1)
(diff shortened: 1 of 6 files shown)


qiime2/jupyter/handlers.py
```diff
@@ -0,0 +1,65 @@
...
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import os
+import pathlib
+
+import tornado.web as web
+from notebook.base.handlers import IPythonHandler
```
ebolyen: This file isn't imported unless `load_jupyter_server_extension` is called, or you import it directly (but don't do that). So I don't think we need to add any of these as actual dependencies. They are optional, and exist if you have Jupyter installed.

ebolyen:
@wasade / @ElDeveloper, any interest in reviewing this PR? Or do you know of someone who would be (this involves a little bit of Tornado code)?

ElDeveloper:
I can have a look later today for sure, this looks great! Can't wait to 
try it out.

qiime2/jupyter/template.py
```diff
@@ -0,0 +1,51 @@
...
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import urllib.parse
+
+
+def make_html(location):
+    url = "/qiime2/redirect?location={location}".format(
+        location=urllib.parse.quote(location))
+    # This is dark magic. An image has an onload handler, which let's me
```
wasade: Yea, that's dark magic

wasade:
@ebolyen, I did a static review and nothing stood out. Testing Tornado handlers gets a bit interesting, will there be unit tests for these?

ebolyen:
@wasade, I haven't the slightest idea how I would test those handlers. They kind of expect to be sitting within a running notebook server.

jakereps:
Definitely better than my chicken scratch first attempt of viewing visualizations in notebooks, haha. Launching a `http.server` within every artifact directory was not a clean way of doing it.

Does this only require pulling this locally and running that extension enable command?

ebolyen:
Ideally that's all you should need to do. It will even "work" if you don't
install the extension.

qiime2/jupyter/template.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+# 404 - the extension isn't installed
+# 428 - the result went out of scope, re-run cell
+# 302->200 - set up the iframe for that location
+_anonymous_function = '''\
+function(div, url){
+var baseURL = require.toUrl('').split('/').slice(0, -2).join('/');
+url = baseURL + url;
+fetch(url).then(function(res) {
```
ElDeveloper: I assume that 404 could also occur when the notebook is loaded from GitHub or nbviewer, in which case perhaps it should note that server-less viewing is not supported.

ebolyen: good idea!

ebolyen:
Turns out this fails in a Jupyterlab notebook. Looks like the JS globals are much cleaner in JLab, so using `require` doesn't work anymore.

ebolyen:
Since I'm going to be refactoring a bit of q2view this cycle for citation tracking, it'll be a good opportunity to make those components more modular, which means they could be used in the notebook and jlab. I think that's maybe the best way to approach this. But this PR does act as a good example of how easy server extensions are to make, so that part is still pretty useful, the client-side repr not as much.

ebolyen:
Resurrecting with just "one more hack" for jupyter lab. It doesn't look like there's time to get the q2view magic refactored, but why let perfect be the enemy of good (or even just-ok)?

ebolyen:
image

ebolyen:
I'm probably not going to touch `Result` or other reprs until we have a better "entry" into the jupyter notebook/lab front-end.

ElDeveloper:
Agree!

PR #389, opened 2018-04-16
REF: shuffle citation API a little

ebolyen:
Some changes to accommodate q2cli


qiime2/plugin/plugin.py
```diff
@@ -57,9 +58,11 @@ def __init__(self, name, version, website, package, citation_text=None,
             self.description = description
 
         if citations is None:
-            self.citations = ()
-        else:
-            self.citations = tuple(citations)
+            citations = ()
+
+        self.citations = Citations([
+            ('plugin|%s:%s|%d' % (self.name, self.version, idx), entry)
```
ebolyen: Not in love with this, but I didn't have a "great" option either, so this seems like the least damage.

PR #395, opened 2018-04-20
ENH: override user pkg dirs for R/Python

ebolyen:
fixes https://github.com/qiime2/q2-dada2/issues/68


ebolyen:
It's probably easiest to let busywork build this out and double check that it works after that.

PR #405, opened 2018-06-11
add sdk.util.artifact_actions

antgonza:
Perhaps a description might be useful. Basically, this adds a new method to retrieve all plugins and their actions with a given input. This could be helpful to figure out which actions can be applied on a given input.

### DIFF
--- a/qiime2/sdk/tests/test_util.py
+++ b/qiime2/sdk/tests/test_util.py
@@ -17,9 +17,6 @@ class TestUtil(unittest.TestCase):
         obs = qiime2.sdk.util.artifact_actions(None)
         self.assertEqual(obs, [])
 
-        obs = qiime2.sdk.util.artifact_actions('ShouldBeEmpty')
-        self.assertEqual(obs, [])
-
         # For simplicity, we are gonna test the names of the plugin and
         # the actions
         obs = [(x.name, [yy.name for yy in y])
@@ -27,6 +24,14 @@ class TestUtil(unittest.TestCase):
         exp = [('dummy-plugin', [
             'Do stuff normally, but override this one step sometimes'])]
         self.assertEqual(obs, exp)
+
+        obs = [(x.name, [yy.name for yy in y])
+               for x, y in qiime2.sdk.util.artifact_actions('Kennel[Cat]')]
+        self.assertEqual(obs, [])
+
+        with self.assertRaises(qiime2.sdk.util.UnknownTypeError):
+            qiime2.sdk.util.artifact_actions('ShouldBeEmpty')
+
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


qiime2/sdk/util.py
```diff
@@ -92,3 +92,29 @@ def parse_format(format_str):
...
+    list of tuples: [(q2.plugin, [q2.actions, ...]), ...]
+    """
+    commands = []
+
+    if string is not None:
+        pm = qiime2.sdk.PluginManager()
+        for pgn, pg in pm.plugins.items():
+            actions = list({a for an, a in pg.actions.items()
+                            for iname, i in a.signature.inputs.items()
+                            if str(i.qiime_type) == string})
```
ebolyen: I think right here you want the condition to be:
```python
if i.qiime_type <= query_type
```
Which will be important for union-types, since their strings won't match, but they have the subtyping relationship.

qiime2/sdk/util.py
```diff
@@ -92,3 +92,29 @@ def parse_format(format_str):
...
+    ----------
+    string : str
+        QIIME2 artifact type
+
+    Returns
+    -------
+    list of tuples: [(q2.plugin, [q2.actions, ...]), ...]
+    """
+    commands = []
+
```
ebolyen: add something along the lines of this:
```python
query_type = qiime2.sdk.util.parse_type(string)
```
to convert the string into a type. I'm pretty sure it will no-op if its already a type, but I'm not sure.

qiime2/sdk/util.py
ebolyen: edit: operator was backwards, fixed in above comment

qiime2/sdk/util.py
```diff
@@ -92,3 +92,30 @@ def parse_format(format_str):
     except KeyError:
         raise TypeError("No format: %s" % format_str)
     return format_record.format
+
+
+def artifact_actions(string):
```
ebolyen: Maybe `actions_by_input_type` would be a better name. It's a little over-specific, but I could see this functionality evolving into a full API with mutliple inputs/output query/solving in the future, so this should be pretty future-proof.

qiime2/sdk/tests/test_util.py
```diff
@@ -0,0 +1,37 @@
...
+            'Do stuff normally, but override this one step sometimes'])]
+        self.assertEqual(obs, exp)
+
+        obs = [(x.name, [yy.name for yy in y])
+               for x, y in qiime2.sdk.util.artifact_actions('Kennel[Cat]')]
+        self.assertEqual(obs, [])
+
+        with self.assertRaises(qiime2.sdk.util.UnknownTypeError):
+            qiime2.sdk.util.artifact_actions('ShouldBeEmpty')
+
```
ebolyen: Let's add one more test to make sure that passing a concrete type like `IntSequence1` find actions that have unions like `IntSequence1 | IntSequence2`.

PR #406, opened 2018-06-15
BUG: Copy if no write access

### LINKED ISSUE #390: qiime.util.duplicate should fall back to copy when write permissions don't exist
Discovered [on the forum](https://forum.qiime2.org/t/qiime-tools-import-wrong/3791). It looks like write permissions are needed to hard-link, we should catch the permission error and fall back to copy, just like a cross-device link error.

ChrisKeefe:
fixes #390

### DIFF
--- a/qiime2/tests/test_util.py
+++ b/qiime2/tests/test_util.py
@@ -7,14 +7,16 @@
 # ----------------------------------------------------------------------------
 
 import os
+import errno
 import tempfile
 import unittest
 import unittest.mock as mock
 
 import qiime2.util as util
 
-EXDEV = OSError(18, "Invalid cross-device link")
-EPERM = PermissionError(1, "find this")
+EXDEV = OSError(errno.EXDEV, "Invalid cross-device link")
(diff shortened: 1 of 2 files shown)


qiime2/tests/test_util.py
```diff
@@ -59,13 +61,25 @@ def test_dst_dir(self):
...
+        with self.assertRaisesRegex(
+                PermissionError, "insufficient r/w permissions"):
             util.duplicate(self.src, self.dst1)
 
         assert mocked_link.called
 
+    @mock.patch('qiime2.util.os.link', side_effect=EPERM)
+    def test_perm_error_EPERM(self, mocked_link):
+        """Does util.duplicate still duplicate if "others" write permissions
+        are absent?"""
```
ebolyen: I suspect not, but it also isn't going to be an issue, because duplicate will refuse to overwrite an existing file (by design), so it won't matter what the permissions of the destination are.

ebolyen: P.S. this docstring will mask the test-name in most test-runners, which is why our unit-tests usually don't have docstrings.

PR #408, opened 2018-06-26
BUG: Fixed docstring order in API

### LINKED ISSUE #319: incorrect ordering of parameters in Artifact API generated numpydoc
In the numpydoc generated for an Artifact API method/visualizer, the `Parameters` section lists the parameters out of order, such that they don't match the function signature.

For example, this function signature declares `feature_table` followed by `max_depth`, but the generated `Parameters` section lists `feature_table` followed by `phylogeny`:

image

I think the code is putting all input artifacts first, followed by (primitive) parameters. We should preserve function signature order to avoid confusion.

ChrisKeefe:
fixes #319

### DIFF
--- a/qiime2/core/testing/method.py
+++ b/qiime2/core/testing/method.py
@@ -51,7 +51,7 @@ def long_description_method(mapping1: dict, name: str, age: int) -> dict:
 
 def docstring_order_method(req_input: dict, req_param: str,
                            opt_input: dict=None, opt_param: int=None) -> dict:
-    return{req_param: opt_param}
+    return {req_param: opt_param}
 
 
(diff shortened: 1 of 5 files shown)


qiime2/core/testing/method.py
```diff
@@ -49,6 +49,11 @@ def long_description_method(mapping1: dict, name: str, age: int) -> dict:
     return {name: age}
 
 
+def docstring_order_method(req_input: dict, req_param: str,
+                           opt_input: dict=None, opt_param: int=None) -> dict:
+    return{req_param: opt_param}
```
thermokarst: style request: can you put a space between `return` and the `dict`?

ChrisKeefe: Done. good catch, thanks!

PR #414, opened 2018-10-24
NEW: adds checksums to archive structure

ebolyen:
This commit adds a 'checksums.md5' file to the root of an archive.
The file is formatted with the md5sum format:

\<hex-hash>\<space>\<space or asterisk>\<escaped filename>\<newline>

Filename escaping handles backslash and newline. In the event those are
present, the backslash is doubled and the newline becomes `\n`.
Additionally if escaping is required, then an additional backslash
precedes the \<hex-hash> to indicate the filename is escaped.
See:
https://www.gnu.org/software/coreutils/manual/html_node/md5sum-invocation.html
for more details.

**TODO**:

- [x] Add tests

### DIFF
--- a/qiime2/core/archive/tests/test_archiver.py
+++ b/qiime2/core/archive/tests/test_archiver.py
@@ -86,6 +86,7 @@ class TestArchiver(unittest.TestCase, ArchiveTestingMixin):
         root_dir = str(archiver.uuid)
         expected = {
             'VERSION',
+            'checksums.md5',
             'metadata.yaml',
             'data/ints.txt',
             'provenance/metadata.yaml',
@@ -104,6 +105,7 @@ class TestArchiver(unittest.TestCase, ArchiveTestingMixin):
         root_dir = str(self.archiver.uuid)
         expected = {
             'VERSION',
+            'checksums.md5',
             'metadata.yaml',
             'data/ints.txt',
             'provenance/metadata.yaml',
@@ -148,6 +150,7 @@ class TestArchiver(unittest.TestCase, ArchiveTestingMixin):
                 '.hidden-file',
                 '.hidden-dir/ignored-file',
                 '%s/VERSION' % root_dir,
+                '%s/checksums.md5' % root_dir,
                 '%s/metadata.yaml' % root_dir,
                 '%s/data/ints.txt' % root_dir,
                 '%s/provenance/metadata.yaml' % root_dir,
@@ -194,6 +197,7 @@ class TestArchiver(unittest.TestCase, ArchiveTestingMixin):
             'data/',
             'data/nested/',
             'VERSION',
+            'checksums.md5',
             'metadata.yaml',
             'data/ints.txt',
             'data/nested/foo.txt',
@@ -217,6 +221,7 @@ class TestArchiver(unittest.TestCase, ArchiveTestingMixin):
         expected = {
             # Directory entries should not be present.
             'VERSION',
+            'checksums.md5',
             'metadata.yaml',
             'data/ints.txt',
             'data/nested/foo.txt',
@@ -280,6 +285,7 @@ class TestArchiver(unittest.TestCase, ArchiveTestingMixin):
             root_dir = str(self.archiver.uuid)
             expected = {
                 '%s/VERSION' % root_dir,
+                '%s/checksums.md5' % root_dir,
                 '%s/metadata.yaml' % root_dir,
                 '%s/data/ints.txt' % root_dir,
                 '%s/provenance/metadata.yaml' % root_dir,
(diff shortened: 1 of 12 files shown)


ebolyen:
Ok this should be ready for review.

ebolyen:
image

ebolyen:
@gregcaporaso @thermokarst @nbokulich,

Question:

Should we copy the `checksums.md5` file to `/provenance/`? That would have the effect of making this file inherited by downstream artifacts (the checksum was already performed, so we would have the fingerprint of the original `/data/` directory for basically free). 

The downside is, the checksums will refer to an archive that no longer exists (from the perspective of the child), and the filepaths will be wrong/unusable with `md5sum`. We also would require some kind of tooling to really make any use of that data, but it would be interesting.

Another option would be to save that for a rainy day (perhaps when we have some notion of "who" did a thing with optional digital signatures? Then a mechanism like this would be almost required...)

gregcaporaso:
> Should we copy the checksums.md5 file to /provenance/? 

@ebolyen, for the reasons you noted I think you should leave that one for a rainy day.

qiime2/core/archive/tests/test_archiver.py
```diff
@@ -331,6 +337,59 @@ def test_parse_uuid_invalid(self):
...
+        # version (which is an unknown commit sha1), so just check name
+        self.assertEqual(list(diff.removed.keys()), ['VERSION'])
+        self.assertEqual(diff.changed,
+                         {'data/ints.txt': ('c0710d6b4f15dfa88f600b0e6b624077',
+                                            'f47bc36040d5c7db08e4b3a457dcfbb2')
+                          })
+
+    def setup_old_archiver(self):
+        class OldArchiver(Archiver):
+            # Quickest way to get the old code running
```
thermokarst: Actually...

https://github.com/qiime2/qiime2/pull/198

qiime2/core/exceptions.py
```diff
@@ -0,0 +1,11 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016-2018, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+
+class ValidationError(Exception):
```
thermokarst: https://youtu.be/NGrLb6W5YOM

qiime2/core/archive/tests/test_archiver.py
```diff
@@ -331,6 +338,44 @@ def test_parse_uuid_invalid(self):
...
+        self.assertEqual(diff.removed, {})
+        self.assertEqual(diff.changed, {})
+
+    def test_checksums_mismatch(self):
+        with (self.archiver.root_dir / 'data' / 'ints.txt').open('w') as fh:
+            fh.write('999\n')
+        with (self.archiver.root_dir / 'tamper.txt').open('w') as fh:
+            fh.write('extra file')
+
+        (self.archiver.root_dir / 'VERSION').unlink()
```
thermokarst: spooky 👻

qiime2/sdk/tests/test_result.py
```diff
@@ -218,6 +221,37 @@ def test_artifact_has_metadata_false(self):
...
+
+        artifact.validate()
+        self.assertTrue(True)  # Checkpoint
+
+    def test_validate_artifact_bad(self):
+        artifact = Artifact.import_data('IntSequence1', [1, 2, 3, 4])
+        with (artifact._archiver.root_dir / 'extra.file').open('w') as fh:
+            fh.write('uh oh')
+
+        with self.assertRaisesRegex(exceptions.ValidationError, 'extra\.file'):
```
thermokarst:
```suggestion
        with self.assertRaisesRegex(exceptions.ValidationError, r'extra\.file'):
```

qiime2/sdk/tests/test_result.py
```diff
@@ -218,6 +221,37 @@ def test_artifact_has_metadata_false(self):
...
+        self.assertTrue(True)  # Checkpoint
+
+    def test_validate_vizualization_bad(self):
+        visualization = Visualization._from_data_dir(
+             self.data_dir, self.make_provenance_capture())
+
+        with (visualization._archiver.root_dir / 'extra.file').open('w') as fh:
+            fh.write('uh oh')
+
+        with self.assertRaisesRegex(exceptions.ValidationError, 'extra\.file'):
```
thermokarst:
```suggestion
        with self.assertRaisesRegex(exceptions.ValidationError, r'extra\.file'):
```

qiime2/sdk/tests/test_result.py
```diff
@@ -218,6 +221,37 @@ def test_artifact_has_metadata_false(self):
...
+
+        artifact.validate()
+        self.assertTrue(True)  # Checkpoint
+
+    def test_validate_artifact_bad(self):
+        artifact = Artifact.import_data('IntSequence1', [1, 2, 3, 4])
+        with (artifact._archiver.root_dir / 'extra.file').open('w') as fh:
+            fh.write('uh oh')
+
+        with self.assertRaisesRegex(exceptions.ValidationError, r'extra\.file'):
```
thermokarst:
```suggestion
        with self.assertRaisesRegex(exceptions.ValidationError,
                                    r'extra\.file'):
```

qiime2/sdk/tests/test_result.py
```diff
@@ -218,6 +221,37 @@ def test_artifact_has_metadata_false(self):
...
+        self.assertTrue(True)  # Checkpoint
+
+    def test_validate_vizualization_bad(self):
+        visualization = Visualization._from_data_dir(
+             self.data_dir, self.make_provenance_capture())
+
+        with (visualization._archiver.root_dir / 'extra.file').open('w') as fh:
+            fh.write('uh oh')
+
+        with self.assertRaisesRegex(exceptions.ValidationError, r'extra\.file'):
```
thermokarst:
```suggestion
        with self.assertRaisesRegex(exceptions.ValidationError,
                                    r'extra\.file'):
```

PR #415, opened 2018-10-24
MAINT: update citation to use preprint

ebolyen:


thermokarst:
What is going on with the failing travis run here? Linting errors and failing tests...

qiime2/citations.bib
```diff
@@ -1,11 +1,12 @@
...
-  title={QIIME allows analysis of high-throughput community sequencing data},
-  author={Caporaso, J Gregory and Kuczynski, Justin and Stombaugh, Jesse and Bittinger, Kyle and Bushman, Frederic D and Costello, Elizabeth K and Fierer, Noah and Peña, Antonio Gonzalez and Goodrich, Julia K and Gordon, Jeffrey I and Huttley, Gavin A and Kelley, Scott T and Knights, Dan and Koenig, Jeremy E and Ley, Ruth E and Lozupone, Catherine A and McDonald, Daniel and Muegge, Brian D and Pirrung, Meg and Reeder, Jens and Sevinsky, Joel R and Turnbaugh, Peter J and Walters, William A and Widmann, Jeremy and Yatsunenko, Tanya and Zaneveld, Jesse and Knight, Rob},
-  journal={Nature methods},
-  volume={7},
-  number={5},
-  pages={335},
-  year={2010},
-  publisher={Nature Publishing Group},
-  doi={10.1038/nmeth.f.303}
+@article{10.7287/peerj.preprints.27295v1,
```
thermokarst: Missing test updates in `qiime2.core.archive.tests.test_citations.TestCitationsTracked` (`test_action` and `test_import`).

thermokarst:
`flake8` 3.6.0 was released yesterday, which includes updates for python 3.7, which is why we are seeing all kinds of new linting errors. @ebolyen, thoughts on proceeding? Seems like we have three options:

- pin flake8
- add `#noqa` directives as needed
- deal with the underlying language changes

Seems like the last option is a lot of work probably best reserved for another day.

qiime2/citations.bib
ebolyen: >_< thanks!

ebolyen:
@thermokarst, what about having @ChrisKeefe work on cleaning up 3? This PR can wait a little bit, and then we're on reasonable footing moving forward. It looks like most of the lint errors are straight-forward and maybe for now we just say `async` becomes `asynchronous`? Then we don't really have to pick a new word or think too hard. q2studio is the only one that uses that API as far as I know.

thermokarst:
Sounds like a great plan, @ebolyen (👍 to `asynchronous`)! @ChrisKeefe - can you connect directly with @ebolyen about this if you decide to start working on it before I am back in the office? Thanks!

PR #429, opened 2019-01-14
MAINT: set up host section in build recipe

thermokarst:


ci/recipe/meta.yaml
```diff
@@ -13,11 +13,12 @@ build:
 
 requirements:
   build:
-    - python {{ python }}
     - setuptools
```
ebolyen: I think setuptools goes in the host section:
https://conda.io/docs/user-guide/tutorials/build-pkgs.html#editing-the-meta-yaml-file

So build is strictly for compilers now (and hopefully still npm for webpack).

PR #431, opened 2019-01-17
ENH: Allow plugin registration sans entrypoint

thermokarst:
Pair-programmed with @ebolyen

### DIFF
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -38,14 +38,14 @@ class PluginManager:
 
     # This class is a singleton as it is slow to create, represents the
     # state of a qiime2 installation, and is needed *everywhere*
-    def __new__(cls):
+    def __new__(cls, install_plugins=True):
         if cls.__instance is None:
             self = super().__new__(cls)
-            self._init()
+            self._init(install_plugins=install_plugins)
             cls.__instance = self
         return cls.__instance
 
-    def _init(self):
+    def _init(self, install_plugins):
         self.plugins = {}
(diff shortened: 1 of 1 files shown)


qiime2/sdk/plugin_manager.py
```diff
@@ -38,29 +38,30 @@ def iter_entry_points(cls):
...
+            # These are all dependent loops, each requires the loop above it to
+            # be completed.
+            for entry_point in self.iter_entry_points():
+                plugin = entry_point.load()
+                self.plugins[plugin.name] = plugin
 
-        for plugin in self.plugins.values():
-            self._integrate_plugin(plugin)
+            for plugin in self.plugins.values():
+                self._integrate_plugin(plugin)
```
ebolyen: So it has occurred to me that this loop might be happening after setting up the plugin dict on purpose, which `install_plugin` can't do without a `finalize()` method of some kind.

ebolyen:
Looking at this again, I don't see any reason this won't work. Any dependence between plugin namespaces is checked in `_integrate_plugin` which presumes previous plugins have already run through `_integrate_plugin`. So the difference between updating the plugin dict and then integrating per-plugin, and filling the plugins dict and then integrating over all shouldn't matter.

thermokarst:
Here is an example invocation, @Oddant1:

https://github.com/thermokarst/q2galaxy/blob/dummy/q2galaxy/__init__.py#L59-L64

thermokarst:
Thanks for looking into this, @Oddant1! This PR has been superseded by #458.

PR #442, opened 2019-03-07
ENH: ActionGraph

Sleightly:


thermokarst:
Hey there @Sleightly! The error on travis is because the testing environment doesn't have networkx available. Want to open a _second_ PR that adds networkx to our [build recipe](https://github.com/qiime2/qiime2/blob/master/ci/recipe/meta.yaml#L20-L27)?

Sleightly:
Sure thing!

qiime2/sdk/typemap.py
```diff
@@ -0,0 +1,48 @@
+import random
+from itertools import combinations
+import networkx as nx
```
thermokarst: Oops, sorry! When I was looking at the failing travis log earlier I saw a message about a missing networkx dependency. Looks like you aren't actually using networkx here, just left the import in. Can you please remove this?

qiime2/sdk/typemap.py
```diff
@@ -0,0 +1,48 @@
...
+import matplotlib.pyplot as plt
```
thermokarst: Please remove unused dependency.

qiime2/sdk/typemap.py
```diff
@@ -0,0 +1,48 @@
...
+
+import qiime 2
```
thermokarst: Hmm, looks like a typo, and also appears to be unused. If that is the case, please remove.

qiime2/sdk/typemap.py
```diff
@@ -0,0 +1,48 @@
...
+
+q2methods = qiime2.sdk.util.actions_by_input_type('FeatureTable[Frequency]')
+
+#############testing##################
```
thermokarst: Please remove "testing" code.

qiime2/sdk/typemap.py
```diff
@@ -0,0 +1,48 @@
+import random
+from itertools import combinations
+import networkx as nx
+import matplotlib.pyplot as plt
+
+import qiime 2
+
+q2methods = qiime2.sdk.util.actions_by_input_type('FeatureTable[Frequency]')
```
thermokarst: This doesn't appear to be used anywhere --- please remove.

qiime2/sdk/typemap.py
```diff
@@ -0,0 +1,48 @@
...
+#############testing##################
+q2p = qiime2.sdk.PluginManager()
+
+#for plugin, value in q2p.plugins.items():
+#    print(value.methods.keys())
+##############testing#################
+
+
+
+class TypeMap:
```
thermokarst: We probably want to think of a new name for this ---`TypeMap` is already taken in [this PR](https://github.com/qiime2/qiime2/pull/444) that @ebolyen is working on. Historically we have used this term in QIIME 2 to refer to a mapping of outputs that belong to a specific set of inputs and parameters.

I don't actually see this code being tested anywhere in your tests above --- can you comment on this?

Sleightly:
Hi Matthew,

Thank you for the comments! I'll get to cleaning up the code. As for
networkx, I do use it in my buildGraph method.

Sincerely,
Edward

thermokarst:
Thanks @Sleightly  --- you might want to double-check that you have checked in all of your changes --- the `buildGraph` function isn't defined _anywhere_ in this PR.

Sleightly:
My apologies, refactoring the code now, should be up soon!

thermokarst:
Thanks @Sleightly, please just ping us here when you are ready for another round of review. Thanks!

Sleightly:
Ready for my second round of code review!

qiime2/sdk/actiongraph.py
```diff
@@ -0,0 +1,217 @@
...
+                    action_list.append(pg.actions[str(action)])
+
+    for action in action_list:
+        node_combs = generate_nodes_by_action(action)
+
+        # each action node is a str(dict)
+        # each Qiime2.type node is the type itself
+        for dict in node_combs:
+            if not G.has_node(str(dict)):
+                G.add_node(str(dict), value=action)
```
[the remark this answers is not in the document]
ebolyen: Ok, after looking at the networkx docs, I understand why you are using this as the key.

It is a reasonable workaround, however, another approach would be to use `id(dict)` which is the object identity (in CPython the memory address). I think that's a reasonable thing to do as looking up specific nodes isn't as useful as looking up by `value`.

qiime2/sdk/actiongraph.py
```diff
@@ -0,0 +1,217 @@
...
+            for k in dict.keys():
+                for type in dict[k]:
+                    if not G.has_node(type):
+                        G.add_node(type, value=type)
+                    if k == 'outputs':
+                        G.add_edge(str(dict), type)
+                    elif k == 'inputs':
+                        G.add_edge(type, str(dict))
+                    elif k == 'params':
+                        G.add_edge(type, str(dict))
```
ebolyen: **Change Requested**

A useful thing to store on the edge would be `name` which is the parameter name of that particular argument type. This is important because some actions have two parameters (cs definition os param) with the same type and it would be nice to be able to grab all incoming edges on a node and to have the parameter names that should be used.

ebolyen: This change will require your `dict[k]` to itself be a dictionary, as a list certainly loses this context.

qiime2/sdk/actiongraph.py
```diff
@@ -0,0 +1,217 @@
...
+    for i in input:
+        for p in param:
+            if non_req:
+                non_req = get_combinations(non_req)
+                for nr in non_req:
+                    d = {}
+                    d['inputs'] = i
+                    d['params'] = p
+                    d['outputs'] = output
+                    d['non_req'] = nr
```
ebolyen: **Change Requested**

This is 99% of what we're looking for, but the `non_req` section should be merged back into `inputs/params` depending on where the content of that came from.

qiime2/sdk/actiongraph.py
```diff
@@ -0,0 +1,217 @@
...
+    param = unravel_union(param)
+
+    for x in input_nr + param_nr:
+        non_req += unravel_union(x)
+
+    for i in input:
+        for p in param:
+            if non_req:
+                non_req = get_combinations(non_req)
+                for nr in non_req:
```
ebolyen: **Implementation Bug**

Because you don't have the empty set included in the powerset from `get_combinations`, this loop will miss the situations where you can omit _all_ optional arguments.

qiime2/sdk/actiongraph.py
```diff
@@ -0,0 +1,217 @@
...
+                    d['outputs'] = output
+                    d['non_req'] = nr
+                    results.append(d)
+            else:
+                d = {}
+                d['inputs'] = i
+                d['params'] = p
+                d['outputs'] = output
+                d['non_req'] = []
+                results.append(d)
```
ebolyen: **Discussion**
I think you could simplify this entire loop if the following was done:

1. Create a function `unravel` such that for every parameter (cs-definition of the term) of an actions signature such that:
   - when it encounters an optional value, it yields None and then yields the value
   - when it encounters a union, it unravels that (and if it is optional it also yields None).

2. zip or otherwise combine the unraveled iterable with the parameter name, so that you now have an iterable of `(name, type)` for each parameter

3. use `itertools.product` on *all* iterables, this will result in a new iterable of tuples where the tuple contains a `(name, type)` value for each parameter from the step above.

4. For each tuple-of-tuple convert that to a dict (you can actually just use `dict(tuple-of-tuple)` to construct it)

5. yield that dict

qiime2/sdk/actiongraph.py
```diff
@@ -0,0 +1,217 @@
...
+            for action in pg.actions.keys():
+                if str(action) in sigs:
+                    action_list.append(pg.actions[str(action)])
+
+    for action in action_list:
+        node_combs = generate_nodes_by_action(action)
+
+        # each action node is a str(dict)
+        # each Qiime2.type node is the type itself
+        for dict in node_combs:
```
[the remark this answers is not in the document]
Sleightly: oops - vim syntaxing didn't highlight that ^^||

qiime2/sdk/actiongraph.py
Sleightly: get_combinations is essentially getting the power set of a set.  

>>> from actiongraph import *
>>> get_combinations([1,2,3])
[[], [1], [2], [3], [1, 2], [1, 3], [2, 3], [1, 2, 3]]

Sleightly: Ended up not needing get_combinations!

qiime2/sdk/actiongraph.py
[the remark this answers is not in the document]
Sleightly: Not sure why id() isn't returning the same thing when the dictionaries created are exactly the same (could be because id == memory location of that object?) 

Would you still like to use ids but store the dictionaries as a value and somehow check if that node exists if the dictionaries match?

Sleightly:
I've updated most of the changes specified! Thanks for the code review @ebolyen and @thermokarst 

Please let me know if there are any additional changes that need to be made!

qiime2/sdk/actiongraph.py
```diff
@@ -0,0 +1,215 @@
...
+    # get all actions or specifc actions if specified in sigs
+    pm = qiime2.sdk.PluginManager()
+    if not sigs:
+        for _, pg in pm.plugins.items():
+            action_list += list(pg.actions.values())
+    else:
+        for _, pg in pm.plugins.items():
+            for action in pg.actions.keys():
+                if action in sigs:
+                    action_list.append(pg.actions[str(action)])
```
ebolyen: This block can become `action_list = sigs` (also, it might be good to call `sigs` just `action_list` instead in the `build_graph` signature).

qiime2/sdk/actiongraph.py
```diff
@@ -0,0 +1,215 @@
...
+    Returns
+    -------
+    nx.DiGraph - networkx graph connected based on all or specified methods
+
+    """
+
+    G = nx.DiGraph()
+    G.edges(data=True)
+    action_list = []
+    sigs = [x.__name__ for x in sigs]
```
ebolyen: This line actually breaks when you provide a list of actions/methods. (see comment below for solution)

qiime2/sdk/actiongraph.py
```diff
@@ -0,0 +1,215 @@
...
+                dict_[k] = action_node
+
+                if k == 'inputs':
+                    for in_k, in_v in v.items():
+                        if not in_v:
+                            continue
+                        if in_k[0] == '.':
+                            name = "opt_"+str(in_v)
+                            G.add_edge(name, str(dict_))
+                            G[name][str(dict_)]['name'] = in_k[1:]
```
ebolyen: at least it is stripped! :+1:

qiime2/sdk/actiongraph.py
```diff
@@ -0,0 +1,215 @@
...
+                        action_node[x[1:]] = y
+                    else:
+                        action_node[x] = y
+                dict_[k] = action_node
+
+                if k == 'inputs':
+                    for in_k, in_v in v.items():
+                        if not in_v:
+                            continue
+                        if in_k[0] == '.':
```
ebolyen: Bit of a hack, but networkx is clearly not making life easy here

qiime2/sdk/actiongraph.py
```diff
@@ -0,0 +1,215 @@
...
+
+    opt_in_list = []
+
+    if opt:
+        opt_in_list += input_nr
+        opt_in_list += param_nr
+        opt_in_list = unravel(opt_in_list)
+        ins = [dict(x) for x in
+               [list(chain.from_iterable(i)) for i in
+                list(product(input, param, opt_in_list))]]
```
ebolyen: This took me a few to track down, but I think there's an error here:

The product should be calculated on something like:
```python
product(*[  # the star here unpacks the list into N arguments for product
   [[key1, type1], [key1, type2]],  # maybe this is an input which was a union of type1/2
   [[key2, typex]],  # maybe these are params
   [[key3, typey]]
])
```
This is subtly different from what you have, which from unravel will create the product of only 3 lists, one of which might look like this:
```python
[  # this outer list is messing things up
[['table', FeatureTable[Frequency]],
  ['representative_sequences', FeatureData[Sequence]],
  ['stats', DeblurStats]]  # this inner list is saying that all of these are part of the same set to take the product of
                           # but we want to take the product of sets where each set is a single key/param name
]
```
Because your three lists typically contain a single list, the cartesian product is just the same 3 lists back out, which why the results _seem_ sensible at first glance.

qiime2/sdk/actiongraph.py
Sleightly: So passing in methods and actions into sigs, are we passing in the <method > or <visualizer >  object that has the action.signature.inputs and action.signature.parameters or are we passing in a list of <functions> that is an object of the function. 

In the later case, we would need to search the plugin for the <method > or <visualizer > object to find the action's attributes.

ebolyen: You would get the actual action object, so you'd have a list of things that have the `.signature` property ready to go

qiime2/sdk/actiongraph.py
ebolyen: Update: I had misunderstood

input/param/opt_in_list are themselves cartesian products, so the product of a product is still the same product as the the product of the individual elements

ebolyen:
Finally figured out how to use networkX. Edges and Nodes look good to me, I added another attribute named `node` which will let users of the graph quickly check whether its an action or type node.

I also fixed a typo on `optional` for output (which was always True).

ebolyen:
Thanks @Sleightly, sorry for the delay, and congrats on your first contribution to QIIME 2! We'll be releasing with this new feature shortly (in the next day or so).

PR #446, opened 2019-03-26
DEP: Add networkx dependency

Sleightly:
Adding networkx for ~TypeMap~ ActionGraph


thermokarst:
This dependency appears to be unused in the new method introduced in https://github.com/qiime2/qiime2/pull/442. It is however used in the tests, but chances are that it might not be necessary.

thermokarst:
Cool, looks like the changeset in #442 has been updated to include the use of this dependency.

PR #447, opened 2019-04-08
BUG: Fixes type mismatch exception for primitives

### LINKED ISSUE #437: Type checking errors can fail for primitive types
**Bug Description**
The type mismatch exception message uses QIIME 2 API specific to Artifacts (`kwarg.type`), which breaks down when `kwarg` is a primitive type (like `float`).

This bug was introduced in #417.

```python
import qiime2
from qiime2.plugins import metadata
md = qiime2.Metadata.load('some_md.tsv')
metadata.actions.tabulate(input=md, page_size=0.5). # page_size should be an int...
```

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-6-46d72e1c9e55> in <module>
----> 1 metadata.actions.tabulate(input=md, page_size=0.5)

<decorator-gen-338> in tabulate(input, page_size)

~/src/qiime2/qiime2/qiime2/sdk/action.py in bound_callable(*args, **kwargs)
    197
    198                 # Type management
--> 199                 self.signature.check_types(**user_input)
    200                 output_types = self.signature.solve_output(**user_input)
    201                 callable_args = {}

~/src/qiime2/qiime2/qiime2/core/type/signature.py in check_types(self, **kwargs)
    299                         "Parameter %r received an argument of type %r. An "
    300                         "argument of subtype %r is required." % (
--> 301                             name, kwargs[name].type, spec.qiime_type))
    302
    303     def solve_output(self, **input_types):

AttributeError: 'float' object has no attribute 'type'
```

**Expected behavior**
The exception should work for primitive types, too.

**References**
1. [Original forum post](https://forum.qiime2.org/t/error-float-object-has-no-attribute-type-in-otu-clustering/8087)

ChrisKeefe:
fixes #437 

- Fixes type mismatch exception for primitives (e.g. when a string is passed where an int is expected)
- Customizes type mismatch error messages to better describe the user error
- Adds tests to support these changes

### DIFF
--- a/qiime2/core/type/signature.py
+++ b/qiime2/core/type/signature.py
@@ -310,7 +310,7 @@ class PipelineSignature:
 
                 elif isinstance(parameter, qiime2.Metadata):
                     raise TypeError(
-                        "Parameter %r received a Metadata object as an "
+                        "Parameter %r received Metadata as an "
                         "argument, which is incompatible with parameter "
                         "type: %r" % (name, spec.qiime_type))
 
--- a/qiime2/core/type/signature.py
+++ b/qiime2/core/type/signature.py
@@ -307,10 +307,15 @@ class PipelineSignature:
                         "Parameter %r requires an argument of type %r. An "
                         "argument of type %r was passed." % (
                             name, spec.qiime_type, kwargs[name].type))
+                elif isinstance(kwargs[name], qiime2.Metadata):
+                    raise TypeError(
+                        "Parameter %r received a Metadata file as an argument,"
+                        " which is incompatible with parameter type: %r" % (
+                            name, spec.qiime_type))
                 else:  # kwarg is a primitive type
                     raise TypeError(
                         "Parameter %r received %r as an argument, which is "
-                        "incompatible with parameter type %r"
+                        "incompatible with parameter type: %r"
                         % (name, kwargs[name], spec.qiime_type))
 
     def solve_output(self, **input_types):
--- a/qiime2/core/type/signature.py
+++ b/qiime2/core/type/signature.py
@@ -302,16 +302,18 @@ class PipelineSignature:
                         "argument. Visualizations may not be used as inputs."
                         % name)
 
-                # Check if is Artifact
-                isArtifact = isinstance(kwargs[name], qiime2.sdk.Artifact)
-                # kwargHasTypeAttribute = is_semantic_type(kwargs[name])
-                passed_type = (kwargs[name].type if isArtifact
+                # store parameter type for later use
+                kwargIsArtifact = isinstance(kwargs[name], qiime2.sdk.Artifact)
+                passed_type = (kwargs[name].type if kwargIsArtifact
                                else type(kwargs[name]).__name__)
+
                 raise TypeError(
-                    "Parameter %r received an argument of type %r. An "
-                    "argument of subtype %r is required." % (
-                        name, passed_type,
-                        spec.qiime_type))
+                    "Parameter %r requires an argument of type %r. An "
+                    "argument of type %r was passed." % (
+                        name, spec.qiime_type,
+                        passed_type))
+
+                # handle artifacts with same types, different subtypes:
 
     def solve_output(self, **input_types):
         # TODO implement solving here. The check for concrete output types may
--- a/qiime2/core/type/signature.py
+++ b/qiime2/core/type/signature.py
@@ -302,18 +302,16 @@ class PipelineSignature:
                         "argument. Visualizations may not be used as inputs."
                         % name)
 
-                # store parameter type for later use
-                kwargIsArtifact = isinstance(kwargs[name], qiime2.sdk.Artifact)
-                passed_type = (kwargs[name].type if kwargIsArtifact
-                               else type(kwargs[name]).__name__)
-
-                raise TypeError(
-                    "Parameter %r requires an argument of type %r. An "
-                    "argument of type %r was passed." % (
-                        name, spec.qiime_type,
-                        passed_type))
-
-                # handle artifacts with same types, different subtypes:
+                if isinstance(kwargs[name], qiime2.sdk.Artifact):
+                    raise TypeError(
+                        "Parameter %r requires an argument of type %r. An "
+                        "argument of type %r was passed." % (
+                            name, spec.qiime_type, kwargs[name].type))
+                else:  # kwarg is a primitive type
+                    raise TypeError(
+                        "Parameter %r received %r as an argument, which is "
+                        "incompatible with parameter type %r"
+                        % (name, kwargs[name], spec.qiime_type))
 
     def solve_output(self, **input_types):
         # TODO implement solving here. The check for concrete output types may
--- a/qiime2/core/type/signature.py
+++ b/qiime2/core/type/signature.py
@@ -294,10 +294,24 @@ class PipelineSignature:
             if ((kwargs[name] not in spec.qiime_type) and
                     not (spec.has_default() and spec.default is None
                          and kwargs[name] is None)):
+
+                # handle visualization passed as parameter
+                if isinstance(kwargs[name], qiime2.sdk.Visualization):
+                    raise TypeError(
+                        "Parameter %r received a Visualization (*.qzv) as an "
+                        "argument. Visualizations may not be used as inputs."
+                        % name)
+
+                # Check if is Artifact
+                isArtifact = isinstance(kwargs[name], qiime2.sdk.Artifact)
+                # kwargHasTypeAttribute = is_semantic_type(kwargs[name])
+                passed_type = (kwargs[name].type if isArtifact
+                               else type(kwargs[name]).__name__)
                 raise TypeError(
                     "Parameter %r received an argument of type %r. An "
                     "argument of subtype %r is required." % (
-                        name, kwargs[name].type, spec.qiime_type))
+                        name, passed_type,
+                        spec.qiime_type))
 
     def solve_output(self, **input_types):
         # TODO implement solving here. The check for concrete output types may
--- a/qiime2/core/type/signature.py
+++ b/qiime2/core/type/signature.py
@@ -289,16 +289,15 @@ class PipelineSignature:
 
     def check_types(self, **kwargs):
         for name, spec in self.signature_order.items():
-            if kwargs[name] not in spec.qiime_type:
-                # A type mismatch is unacceptable unless the value is None
-                # and this parameter's default value is None.
-                if not (spec.has_default() and
-                        spec.default is None and
-                        kwargs[name] is None):
-                    raise TypeError(
-                        "Parameter %r received an argument of type %r. An "
-                        "argument of subtype %r is required." % (
-                            name, kwargs[name].type, spec.qiime_type))
+            # A type mismatch is unacceptable unless the value is None
+            # and this parameter's default value is None.
+            if ((kwargs[name] not in spec.qiime_type) and
+                    not (spec.has_default() and spec.default is None
+                         and kwargs[name] is None)):
+                raise TypeError(
+                    "Parameter %r received an argument of type %r. An "
+                    "argument of subtype %r is required." % (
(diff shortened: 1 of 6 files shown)


qiime2/core/type/signature.py
```diff
@@ -289,16 +289,36 @@ def decode_parameters(self, **kwargs):
...
-                if not (spec.has_default() and
-                        spec.default is None and
-                        kwargs[name] is None):
+            # A type mismatch is unacceptable unless the value is None
+            # and this parameter's default value is None.
+            if ((kwargs[name] not in spec.qiime_type) and
+                    not (spec.has_default() and spec.default is None
+                         and kwargs[name] is None)):
+
+                # handle visualization passed as parameter
```
thermokarst: Do you think this comment adds anything to the discussion? My concern is that it basically echoes the line below. My rule of thumb: code declares _how_, comments declare _why_.

One way to make the code "speak" more clearly is to "rename" `kwargs[name]` somewhere up here:

```python
parameter = kwargs[name]

if isinstance(parameter, qiime2.sdk.Visualization):
    ...
```

Now I read that line as "if the parameter is an instance of `Visualization`..."

qiime2/core/type/signature.py
```diff
@@ -289,16 +289,36 @@ def decode_parameters(self, **kwargs):
...
+                if isinstance(kwargs[name], qiime2.sdk.Visualization):
                     raise TypeError(
-                        "Parameter %r received an argument of type %r. An "
-                        "argument of subtype %r is required." % (
-                            name, kwargs[name].type, spec.qiime_type))
+                        "Parameter %r received a Visualization (*.qzv) as an "
```
thermokarst: I like this error message! One small note though - not all `Visualization`s are qzv files --- if you are an Artifact API user you can generate a `Visualization` and "view" it in a notebook without ever needing to save it as a qzv.

qiime2/core/type/signature.py
```diff
@@ -289,16 +289,36 @@ def decode_parameters(self, **kwargs):
...
+                # handle Artifacts
+                if isinstance(kwargs[name], qiime2.sdk.Artifact):
+                    raise TypeError(
+                        "Parameter %r requires an argument of type %r. An "
+                        "argument of type %r was passed." % (
+                            name, spec.qiime_type, kwargs[name].type))
+                # handle Metadata without displaying long Metadata repr
+                elif isinstance(kwargs[name], qiime2.Metadata):
+                    raise TypeError(
+                        "Parameter %r received a Metadata file as an argument,"
```
thermokarst: Similar to my comment above --- not all `Metadata` is a file --- an Artifact API user can construct `Metadata` and provide as input without ever saving to disk (in this example, it would be called a "Metadata object"). Make sense?

qiime2/sdk/tests/test_bad_inputs.py
```diff
@@ -0,0 +1,161 @@
...
+from qiime2.core.testing.type import IntSequence1, IntSequence2, SingleInt
+from qiime2.core.testing.visualizer import most_common_viz
+from qiime2 import Metadata
+from qiime2.metadata.tests.test_io import get_data_path
+
+
+class TestBadInputs(TestPluginBase):
+
+    def make_provenance_capture(self):
+        # Copy-pasted from qiime2.sdk.TestResult.make_provenance_capture.
```
thermokarst: Since this is a within-project copy-paste I don't think you need this comment. You could always just re-write the original comment to say something like "importing visualizations is not supported, but we do that here to simplify testing machinery".

qiime2/sdk/tests/test_bad_inputs.py
```diff
@@ -0,0 +1,161 @@
...
+from qiime2.plugin.testing import TestPluginBase
+
+from qiime2.sdk import Artifact, Visualization
+from qiime2.core.testing.type import IntSequence1, IntSequence2, SingleInt
+from qiime2.core.testing.visualizer import most_common_viz
+from qiime2 import Metadata
+from qiime2.metadata.tests.test_io import get_data_path
+
+
+class TestBadInputs(TestPluginBase):
```
thermokarst: I'm not sure this test case needs to be in a new test file (`test_bad_inputs.py`) --- maybe it could get moved into `test_method.py`?

qiime2/sdk/tests/test_method.py
```diff
@@ -396,9 +396,10 @@ def test_call_with_optional_artifacts(self):
 
         self.assertEqual(obs.view(list), [0, 42, 43, 42, 99, -22, 43, 43, 111])
 
+        # TODO: break this out into section with other "bad input" tests?
```
thermokarst: Leftover dev comment?

qiime2/sdk/tests/test_visualizer.py
```diff
@@ -95,7 +95,7 @@ def test_from_function_with_parameters_only(self):
             inputs={},
             parameters={
                 'name': qiime2.plugin.Str,
-                'age': qiime2.plugin.Int
+                'age': qiime2.plugin.Int % Range(0, None)
```
thermokarst: :+1:

qiime2/sdk/tests/test_bad_inputs.py
```diff
@@ -0,0 +1,161 @@
...
+        self.test_dir = tempfile.TemporaryDirectory(prefix='qiime2-test-temp-')
+        self.data_dir = os.path.join(self.test_dir.name, 'viz-output')
+        os.mkdir(self.data_dir)
+        most_common_viz(self.data_dir, collections.Counter(range(42)))
+
+    def tearDown(self):
+        self.test_dir.cleanup()
+
+    def test_viz_passed_as_input(self):
+        # generate a sample viz and other params
```
thermokarst: Please see my comment above about _how_ vs _why_ w.r.t. comments.

qiime2/sdk/tests/test_bad_inputs.py
thermokarst: These tests all look good to me, I _think_ you have all the cases covered! I will take a closer look in round 2 of review!

qiime2/core/type/signature.py
```diff
@@ -289,16 +289,36 @@ def decode_parameters(self, **kwargs):
...
+                if isinstance(kwargs[name], qiime2.sdk.Visualization):
                     raise TypeError(
-                        "Parameter %r received an argument of type %r. An "
-                        "argument of subtype %r is required." % (
-                            name, kwargs[name].type, spec.qiime_type))
+                        "Parameter %r received a Visualization (*.qzv) as an "
+                        "argument. Visualizations may not be used as inputs."
+                        % name)
+
+                # handle Artifacts
```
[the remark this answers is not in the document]
ChrisKeefe: 🗸

qiime2/core/type/signature.py
ChrisKeefe: Love this. Thanks!

qiime2/core/type/signature.py
ChrisKeefe: 🗸

qiime2/core/type/signature.py
ChrisKeefe: 🗸

qiime2/sdk/tests/test_bad_inputs.py
ChrisKeefe: > I'm not sure this test case needs to be in a new test file (`test_bad_inputs.py`) --- maybe it could get moved into `test_method.py`?

These tests use a couple of visualizers and pipelines with features not available in our current dummy-plugin methods. Am I right to assume that these non-method tests shouldn't live in `test_method.py'? I kinda like breaking these out thematically, but understand the organizational slippery slope there. 

Is it worth adding/modifying a method with `choices` and `range` predicates to fill the gap, then bringing these into `test_method`?

qiime2/sdk/tests/test_method.py
ChrisKeefe: Yes. And related to the comment above (combining with `test_method.py`), is there value in digging through `test_method`, `test_pipeline`, etc to consolidate or remove type mismatch tests that are covered more systematically by the tests in this PR?

qiime2/sdk/tests/test_bad_inputs.py
ChrisKeefe: 🗸

qiime2/sdk/tests/test_bad_inputs.py
ChrisKeefe: 🗸

qiime2/sdk/tests/test_bad_inputs.py
thermokarst: How about you split these new tests up? Methods into `test_method.py`, Visualizers into `test_visualizer.py`, Pipelines into `test_pipeline.py`. These tests could still be "grouped" in a `TestBadInputs` suite in each respective file.

qiime2/sdk/tests/test_method.py
thermokarst: I vote for leaving the existing actions alone.

ChrisKeefe: 🗸

qiime2/sdk/tests/test_bad_inputs.py
ChrisKeefe: Renamed file to `test_action.py` per our discussion, to align with sdk test naming conventions and remove the need to break up test functions that use more than one type of Action to test type mismatches.

qiime2/core/type/signature.py
thermokarst: I think we overcorrected in the wrong direction here --- for all the same reasons why we might not have "qzv files", we also can't guarantee that we have "objects" here, either.

qiime2/core/type/signature.py
thermokarst: I think we overcorrected in the wrong direction here --- for all the same reasons why we might not have "qzv files", we also can't guarantee that we have "objects" here, either.

qiime2/core/type/signature.py
```diff
@@ -289,16 +289,36 @@ def decode_parameters(self, **kwargs):
...
+
+                elif isinstance(parameter, qiime2.sdk.Artifact):
+                    raise TypeError(
+                        "Parameter %r requires an argument of type %r. An "
+                        "argument of type %r was passed." % (
+                            name, spec.qiime_type, parameter.type))
+
+                elif isinstance(parameter, qiime2.Metadata):
+                    raise TypeError(
+                        "Parameter %r received a Metadata object as an "
```
thermokarst:
```suggestion
                        "Parameter %r received Metadata as an "
```

qiime2/core/type/signature.py
ChrisKeefe: Handled. Thanks!

qiime2/core/type/signature.py
ChrisKeefe: done.

thermokarst:
@ChrisKeefe, please update the changelog with these changes. Thanks!

thermokarst:
BTW @ChrisKeefe, at least one instance of collateral damage due to this PR: https://busywork.qiime2.org/teams/main/pipelines/master-distribution/jobs/unit-test-q2cli/builds/28

Looks like at least one test in q2cli needs an update.

thermokarst:
I pushed up a change, q2cli is back in business.

ChrisKeefe:
Thanks @thermokarst. Do you expect we're going to see similar breakages in the API or studio tests? Is it worth my digging for them, or just wait until they come up in busywork?

thermokarst:
I wouldn't expect any issues with the Artifact API, since that is co-located in the framework --- you would've seen test failures for that locally. Or maybe you have something else in mind when you say "API"? Let me know!

As far as q2studio goes, unfortunately it isn't wired up to busywork, so we might be waiting for a long time. The jury is still out on q2studio this release, so I would just wait and see.

PR #461, opened 2019-06-06
IMP: Artifacts now max validate on import

Oddant1:


qiime2/sdk/result.py
```diff
@@ -237,10 +237,12 @@ def import_data(cls, type, view, view_type=None):
...
 
         provenance_capture = archive.ImportProvenanceCapture(format_, md5sums)
-        return cls._from_view(type_, view, view_type, provenance_capture)
+        return cls._from_view(type_, view, view_type, provenance_capture,
+                              level='max')
 
     @classmethod
-    def _from_view(cls, type, view, view_type, provenance_capture):
+    def _from_view(cls, type, view, view_type, provenance_capture,
+                   level='min'):
```
ebolyen: maybe call it `validate_level`, because we are pretty far removed from the code that does that, so out of context, `level` may not make very much sense.

qiime2/sdk/result.py
```diff
@@ -262,7 +264,7 @@ def _from_view(cls, type, view, view_type, provenance_capture):
         recorder = provenance_capture.transformation_recorder('return')
         transformation = from_type.make_transformation(to_type,
                                                        recorder=recorder)
-        result = transformation(view)
+        result = transformation(view, level)
```
ebolyen: Maybe `validate_level` here as well, but in the `validate` methods, `level` on its own is perfectly fine.

qiime2/core/transform.py
```diff
@@ -63,14 +63,14 @@ def make_transformation(self, other, recorder=None):
...
+        def transformation(view, validate_level='min'):
             view = self.coerce_view(view)
-            self.validate(view)
+            self.validate(view, validate_level)
 
             new_view = transformer(view)
 
             new_view = other.coerce_view(new_view)
-            other.validate(new_view)
+            other.validate(new_view, validate_level)
```
ebolyen: Sorry @Oddant1, realized as I was walking in, we probably don't want to validate the result of the transform at `max`. As then we would be scanning sequence data twice on import.

qiime2/sdk/tests/test_artifact.py
```diff
@@ -496,7 +502,15 @@ def test_artifact_validate_max(self):
...
         A = Artifact.import_data('IntSequence1', [1, 2, 3, 4, 5, 6, 7, 10])
         with self.assertRaisesRegex(ValidationError, '3 more'):
-            A.validate('max')
+            A.validate(level='max')
+
+    def test_artifact_validate_max_on_import(self):
+        A = Artifact.import_data('IntSequence1', [1, 2, 3, 4, 5, 8])
+        A.validate('min')
+        self.assertTrue(True)  # Checkpoint assertion
+        fp = get_data_path('intsequence-fail-max-validation.txt')
```
ebolyen: I would recommend adding
```python
fmt = IntSequenceFormat(fp)
fmt.validate(level='min')
self.assertTrue(True)  # Checkpoint assertion
```

As this demonstrates that the file is valid at a minimum level, therefore the error raised on import _must_ be because it is validating max now.

qiime2/sdk/tests/test_artifact.py
```diff
@@ -496,7 +502,15 @@ def test_artifact_validate_max(self):
         self.assertTrue(True)  # Checkpoint assertion
         A = Artifact.import_data('IntSequence1', [1, 2, 3, 4, 5, 6, 7, 10])
         with self.assertRaisesRegex(ValidationError, '3 more'):
-            A.validate('max')
+            A.validate(level='max')
+
+    def test_artifact_validate_max_on_import(self):
+        A = Artifact.import_data('IntSequence1', [1, 2, 3, 4, 5, 8])
+        A.validate('min')
```
ebolyen: This doesn't verify the format quite the way we want, since it doesn't prove anything about `intsequence-fail-max-validation.txt`.

ebolyen:
Thanks @Oddant1!

PR #464, opened 2019-06-19
IMP: Merging metadata handles overlapping metadata column names

Oddant1:
Metadata files with overlapping column names can now be merged.
NOTE: This breaks some tests in q2cli, and it was not immediately apparent for myself or @ebolyen as to exactly where and why it broke those tests.


thermokarst:
One thing that concerns me about this approach is that now the order of files specified really matters. If a user changes their specification order, well, they can wind up with totally different results. Some options:

- maybe we can come up with deterministic sorting within the merge functionality?
- generate some kind of hash-based id to refer to each metadata

Thoughts?

thermokarst:
> NOTE: This breaks some tests in q2cli, and it was not immediately apparent for myself or @ebolyen as to exactly where and why it broke those tests.

BTW, the test failures in q2cli maybe aren't surprising, there are a few tests specific to that interface's automatic merging functionality: https://github.com/qiime2/q2cli/blob/master/q2cli/tests/test_cli.py

thermokarst:
What I am concerned about above is the case where a user accidentally swaps the order of their files in an invocation (using the cli here as an example, but the same issue applies when merging manually in the artifact API):

```bash
qiime longitudinal anova \
  --m-metadata-file faith_pd_a.qza \
  --m-metadata-file faith_pd_b.qza \
  --m-metadata-file sample_metadata.tsv \
  --p-formula " 'faith_pd [1]' ~ some_tsv_col * 'faith_pd [2]'" \
  --o-visualization foo.qzv
```

vs

```bash
qiime longitudinal anova \
  --m-metadata-file faith_pd_b.qza \
  --m-metadata-file faith_pd_a.qza \
  --m-metadata-file sample_metadata.tsv \
  --p-formula "'faith_pd [1]' ~ some_tsv_col * 'faith_pd [2]'" \
  --o-visualization foo.qzv
```

The second one swaps the order of the alpha diversity results, but doesn't handle updating the labels in the formula. This will change the interpretation of the formula, with no warning to the user.

If we did something to uniquely identify which file the overlapping columns came from, this would allow for the merge to not be influenced by the order:

```bash
qiime longitudinal anova \
  --m-metadata-file faith_pd_a.qza \
  --m-metadata-file faith_pd_b.qza \
  --m-metadata-file sample_metadata.tsv \
  --p-formula " 'faith_pd [45de1dd]' ~ some_tsv_col * 'faith_pd [31eb03b]'" \
  --o-visualization foo.qzv
```

vs 

```bash
qiime longitudinal anova \
  --m-metadata-file faith_pd_b.qza \
  --m-metadata-file faith_pd_a.qza \
  --m-metadata-file sample_metadata.tsv \
  --p-formula " 'faith_pd [45de1dd]' ~ some_tsv_col * 'faith_pd [31eb03b]'" \
  --o-visualization foo.qzv
```

Both of these are the same. Now, the issue here is: "how do I know which hash belongs to which column? Maybe there is some utility exposed elsewhere (like a new `.hash` or `.id` property). 

Other variations on this: still render out `[1]`, `[2]`, but sort the metadata files internally in some deterministic way, that way the order specified doesn't matter. Finally, we could also append the counter to the hash --- this would cause the command to error out on the user if they changed the order specified, which would be ideal IMO.

Oddant1:
That last commit message should be "SQUASH: Fixed an `if` statement" but I guess the commit message didn't like the `if`. . . or I forgot it

ebolyen:
That's a really good example @thermokarst, thanks! We weren't thinking of any actions like anova. I think I agree this could be more problematic than we thought originally.

ebolyen:
What about something like this:

What if `Metadata.merge` required suffixes when dups are found? In the python API, this could look like anything, but the CLI will choose to use the filename (up to a distinct path) as the suffixes. This isn't initially customizable behavior in the CLI, if you want specific suffixes, then you have to use the Python API.

Then the CLI always does something sane (and relatively predictable), and in Python, that's totally up to you. In this case it's just an argument the CLI knows to pass automatically.

I think this sidesteps the problem of object artifacts not having a filepath and of needing to immediately creating a suffix syntax for the CLI right away

ebolyen:
cc @gregcaporaso who was part of the initial discussion.

Oddant1:
> What if `Metadata.merge` required suffixes when dups are found?

@ebolyen do you mean that if `merge` finds duplicates it blocks and prompts the user to enter suffixes? Or that when `merge` finds duplicates it breaks and tells the user to rerun the command with suffixes for "xyz" files?

ebolyen:
> Or that when merge finds duplicates it breaks and tells the user to rerun the command with suffixes for "xyz" files?

The latter, it would probably expect a list in the same order as the arguments:
```
md1.merge(md2, md3, suffixes=['GG', 'SILVA', 'NewThing'])

**Comment by ebolyen** [comment]
Actually, one more thought on this:

List currently works this way as well:
```
--p-list 1
--p-list 2
```
is different from 
```
--p-list 2
--p-list 1
```

However, the new CLI parser can do this now:
```
--p-list 1 2
```
which is kind of way better, should we be emphasizing similar things for metadata merging?
```
--m-whatever-file example.tsv artifacta.qza
```

I get that generally options are re-orderable, but that isn't true of of _some_ of them.

**Comment by Oddant1** [comment]
@ebolyen would that look like all of the `--m-metadata` and other arguments for the method followed by a `--p-list "file-ids"`? Wouldn't that have the same issue @thermokarst brought up before where it isn't robust to them accidentally changing the order of the files they pass because the ids they provide aren't uniquely associated with that file? Come to think of it, doesn't them passing the ids through the Python API also have that same issue?

**Comment by ebolyen** [comment]
Sorry I didn't mean to suggest that they would do anything with the list as a suffix, rather we already have a `List[X]` type which has the same problem @thermokarst was describing.

And yes, if you pass a list in the wrong order, it will be wrong... So in that way, the hash is superior, since you can't mess that up, but you could make the anova formula wrong too.

**Comment by Oddant1** [comment]
> Sorry I didn't mean to suggest that they would do anything with the list as a suffix, rather we already have a `List[X]` type which has the same problem @thermokarst was describing.

Ah ok

**Comment by ebolyen** [comment]
I will say, I've always kind of felt like the UUID is the natural key here, so if metadata files had a hash as well, it would make the columns _super_ traceable, which I like.

```
Taxonomy [12c9566f-d05f-4067-af7d-8c830ac3cd3f]
ArbitraryDup [a79439c986e60c11601aeabadf4ec963]
```

I think @gregcaporaso mentioned disliking this since users probably don't know where these numbers come from. But I would be completely fine with that. Like @thermokarst mentioned, we could make a `.hash` and maybe `inspect-metadata` is a natural place for it. The only trouble is what to hash, it seems like maybe we serialize it pre-emptively and hash that? So that the metadata is normalized?

**Comment by Oddant1** [comment]
@ebolyen As long as we have a clear way for the user to know which file which column came from. I don't like the idea of them having to do any real digging to figure that out

**Comment by gregcaporaso** [comment]
So it sounds like the two options being discussed here are: 

1) CLI (or more generally, interfaces that always operate on qza files?) defaults to using filenames or filepaths as suffixes; API has the user provide suffixes explicitly. 
2) hash/uuid is used _and_ new functionality is added to make that info easily accessible. 

2 seems more labor intensive for the user (e.g., if a user has four different artifacts with the same column names, such as a volatility plot generated for four PCoA Results, they're going to have to look up the mapping each time they're trying to interpret the results). 1 is a little clunkier (we probably have to extend filenames out to filepaths sometimes so they're unique), but probably less labor intensive for the user - though sometimes this could get annoying too because they have to find the parts of the suffix that differ, which might be only a couple of characters in a big path. 

What if we start with 2 since that sounds like reasonable default behavior? If it becomes a pain for users (we get a lot of forum questions about it), we could add optional suffixes in the future that could be included in addition to or in place of hash/uuid suffixes.

**Comment by ebolyen** [comment]
Sounds good to me! How do we want to define this hash? I'll throw out using `.save()` and the `md5sum` on it as a first approach.

**Comment by thermokarst** [comment]
How about converting the backing dataframe to bytes, then hashing that? We could memoize that to reduce the cost.

**Comment by ebolyen** [comment]
Wouldn't that look different every time as the python strings would be on the heap at different memory addresses each time you load the same file?

**Comment by thermokarst** [comment]
Not sure, I am AFK, but a quick Google search earlier implied that this would be deterministic

**Comment by ebolyen** [comment]
Oh, I see: https://stackoverflow.com/a/47800021/579416

I am a little nervous tying ourselves to a pandas API here, as that does eventually break, and this seems like less trodden territory, but it would definitely be faster.

@Oddant1, want to experiment with these approaches a bit and see how well they work?

**Comment by Oddant1** [comment]
https://github.com/qiime2/qiime2/pull/464/commits/70df03cca1c7f638fba87e55c2e9a487fe4c6e9c : This commit begins to implement some of the additional ideas we have discussed. Duplicated columns no longer append a simple number, they now append a hash of the file(s) that column came from. If the source file is a single artifact file the current behavior is to use that artifact's uuid. If the source file is a single .tsv it uses a hash of the DataFrame view of the .tsv. If the source is a file with actual history, it uses an XOR of all the hashes in the provenance. This means the metadata now keeps a record of both the artifacts it came from and all of the non-artifacts in its history.

File: qiime2/metadata/metadata.py [hunk 1]
```diff
@@ -100,6 +116,7 @@ def __init__(self, index):
         self._ids = tuple(index)
 
         self._artifacts = []
+        self._non_artifact_metadata = []
```
thermokarst: I think we will need to update the eq/neq dunder methods to include these, right?

qiime2/metadata/metadata.py
```diff
@@ -362,6 +386,43 @@ def __init__(self, dataframe):
         super().__init__(dataframe.index)
 
         self._dataframe, self._columns = self._normalize_dataframe(dataframe)
+        self.contains_renamed_columns = False
+
+    @property
+    def hash(self):
```
thermokarst: Is `hash` the right word here? To me this seems more like a metadata `.id` property, where the id is either a UUID (in the case of Artifact-sourced MD), or a hash of the TSV representation of MD.

qiime2/metadata/metadata.py
```diff
@@ -87,6 +89,20 @@ def artifacts(self):
         """
         return tuple(self._artifacts)
 
+    @property
+    def non_artifact_metadata(self):
+        """Non artifacts that are the source of the metadata.
```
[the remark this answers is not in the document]
thermokarst: Maybe we don't need this property at all? One option is to refactor this and the `artifacts` property above as a new `source` (or similar) property. That property could expose some new list-like data made up of records that keep track of the data, and the type of data (native object vs Artifact). I think this might allow you to clean up some logic down below (but, I haven't looked too closely). The pre-existing `artifacts` property could remain, but it could just return a filtered view of the `source` property.

Again, just spitballing an idea, but, might be worth looking into.

qiime2/metadata/metadata.py
```diff
@@ -362,6 +386,43 @@ def __init__(self, dataframe):
...
+    def hash(self):
+        if not isinstance(self._hash, uuid.UUID):
+            return f'{self._hash:x}'
+        else:
+            return self._hash
+
+    @property
+    def _hash(self):
+        hash = None
+        if self.artifacts:
```
thermokarst: One (subjective) pattern that I think improves readability is to go from this:

```python
if self.artifacts:
    hash = self.artifacts[0].uuid
    for artifact in self.artifacts[1:]:
        hash = int(hash) ^ int(artifact.uuid)
```

to this:

```python
if self.artifacts:
    artifacts = iter(self.artifacts)

    # always at least one here
    hash = getattr(next(artifacts), 'uuid')

    for artifact in artifacts:
        hash = int(hash) ^ int(artifact.uuid)
```

YMMV.

qiime2/metadata/metadata.py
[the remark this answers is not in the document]
Oddant1: I've been playing with that idea, and I think I like it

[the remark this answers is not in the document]
thermokarst: Okay, how about we keep track of three separate sources:

- `artifact_sources`
- `dataframe_sources`
- `tsv_sources`

?

I haven't looked too closely at the constructors in a while, so I might be forgetting some "gotcha" w.r.t disambiguating the dataframe and tsv sources, but my hunch is that with some updates to the constructors, we can keep track of "who is who". I think this will be useful, especially down the line.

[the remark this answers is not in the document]
Oddant1: Would we have three separate lists for those three things or would we still have only a single `source` list that we would filter those out of?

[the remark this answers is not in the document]
Oddant1: It shouldn't be to hard to differentiate between those three sources either by the way. Sources from dataframes call the `Metadata` constructor and pass it a `DataFrame` while sources from tsvs call `Metadata.load` on a tsv. Checking if it came from an artifact can be done just by typechecking.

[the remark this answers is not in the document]
Oddant1: Previously we were adding the `Metadata` objects to the `sources` list for all `non_artifact` sources. If we're going to distinguish between .tsvs and `DataFrame` object sources now, should we be adding something that more accurately reflects those to the `sources` list. Like say the actual `DataFrame` object for `Dataframe` sources and the filepath to the .tsv for .tsv sources or something?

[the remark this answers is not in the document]
Oddant1: @thermokarst https://github.com/qiime2/qiime2/pull/464/commits/8ef8dda515213b43cbc4aa1cde5b7e9f5c13f528 this commit begins to roughly implement what we discussed above. Not really complete or tested, but something along those lines.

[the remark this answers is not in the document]
thermokarst: Thanks @Oddant1, I will respond inline.

qiime2/metadata/metadata.py
```diff
@@ -85,7 +87,123 @@ def artifacts(self):
...
-        return tuple(self._artifacts)
+        artifacts = []
+        for source in self._sources:
+            if isinstance(source, qiime2.Artifact):
+                artifacts.append(source)
+        return tuple(artifacts)
+
+    # This is to be gotten rid of
+    @property
+    def non_artifacts(self):
```
thermokarst: Do we need this property now that we have our "partitioned" sources?

qiime2/metadata/metadata.py
```diff
@@ -85,7 +87,123 @@ def artifacts(self):
...
+        Returns
+        -------
+        tuple of qiime2.Metadata and qiime2
+            Source metadata of this metadata.
+
+        """
+        return tuple(self._sources)
+
+    @property
+    def source_type(self):
```
thermokarst: Do we need this? How does this work when the current MD instance is the result of a merging operation?

qiime2/metadata/metadata.py
```diff
@@ -99,32 +217,44 @@ def __init__(self, index):
         self._validate_index(index, axis='id')
         self._ids = tuple(index)
 
-        self._artifacts = []
+        self._sources = []
```
thermokarst: I think now that we are tracking the sources in their own partitions, we can walk back on this "pooled" approach, and instead just track the lists in private attributes (`self._artifacts`, `self._tsvs`, etc).

Oddant1: @thermokarst, for some reason none of your last three comments were showing up for me until just now

qiime2/metadata/metadata.py
Oddant1: I've gone back and forth on that, initially I had it commented "This is to be gotten rid of" up there, I left it because some other things were dependent on it, and I didn't have time to change that yesterday, but now I feel like it probably should be there. It certainly doesn't hurt anything by being there

qiime2/metadata/metadata.py
Oddant1: The source type is "merge" without this and with the single pooled `_source` attribute how would we know specifically what the source of a given object was?

qiime2/metadata/metadata.py
Oddant1: Should we still have single `@property sources` that returns a tuple of all of those different sources or something similar as well as `@property` methods that return those different sources individually?

Oddant1: Additionally, are you talking about adding the actual DataFrames/tsvs themselves to these properties? If so how? The actual `DataFrame` objects go in the `DataFrames` property and then the filepath to the tsv or something goe in the `tsvs` property? Currently the artifacts are `Artifact` objects and everything else is a `Metadata` object.

qiime2/metadata/metadata.py
thermokarst: One thing to consider is public API for these objects --- once you set something loose in the wild it can be complicated changing or recalling it. I am going to write up a more detailed comment later today with some skeleton code so that we can discuss.

qiime2/metadata/metadata.py
thermokarst: I think I suggested _replacing_ `_source` with these new partitioned source properties, so that you can still keep track of where things came from without quite so much effort. I will write up my thoughts in more detail later today.

qiime2/metadata/metadata.py
thermokarst: > Should we still have single @property sources that returns a tuple of all of those different sources or something similar as well as @property methods that return those different sources individually?

Maybe, I need to stew on that a bit.

> Additionally, are you talking about adding the actual DataFrames/tsvs themselves to these properties? If so how

I was envisioning either the `Metadata` objects themselves, or, perhaps a list of IDs.

I will write up my thoughts in more detail later.

thermokarst:
Okay, the following skeleton :skull:  is an attempt to summarize some of the ideas in my last round or two of discussion, please take a look and let me know if you have any questions! :question: 

```python
class _MetadataBase:
    @property
    def artifacts(self):
        return tuple(self._artifacts)

    @property
    def tsvs(self):
        return tuple(self._tsvs)

    @property
    def dataframes(self):
        return tuple(self._dataframes)

    def __init__(self, index):
        ...
        self._artifacts = []
        self._tsvs = []
        self._dataframes = []
        ...

    # Note, this replaces the current `_add_artifacts`
    def _add_source_from_artifact(self, artifact: qiime2.Artifact):
        ...
        self._artifacts.append(artifact)

    def _add_source_from_tsv(self, md: qiime2.Metadata):
        ...
        self._tsvs.append(md)

    def _add_source_from_dataframe(self, md: qiime2.Metadata):
        ...
        self._dataframes.append(md)

    @property
    def _id(self):
        id_ = 0
        for source in self._artifacts:
            id_ = int(id_) ^ int(source.uuid)

        sum_file = tempfile.NamedTemporaryFile(prefix='md5-')
        for source in itertools.chain(self._dataframes, self._tsvs):
            source.save(sum_file.name)
            id_ = int(id_) ^ int(md5sum(sum_file.name), 16)
        sum_file.close()

        return id_


class Metadata(_MetadataBase):
    def __init__(self, dataframe):
        ...
        super().__init__(dataframe.index)
        ...
        self._add_source_from_dataframe(self)


class MetadataReader:
    def read(self, into, column_types=None):
        ...
        try:
            md = into(df)
            md._add_source_from_tsv(md)
            return md
        ...


class Artifact(Result):
        ...
        if view_type is qiime2.Metadata:
            result._add_source_from_artifact(self)
        ...
```

This is untested code, so YMMV, but the point is to illustrate the idea! :bulb:

Oddant1:
@thermokarst the only reason I did it the way I did it and not that way is because using that method the metadata contains a reference to itself as a source which I didn't think made sense, but if you're fine with that happening/want that to happen I'm happy to implement it that way. It definitely simplifies some pieces of logic! Additionally, do you want a list of sources that are metadata from previous merges as well? I just kind of threw the "Merge" source_type in the last implementation because it seemed like something that was missing.

thermokarst:
> @thermokarst the only reason I did it the way I did it and not that way is because using that method the metadata contains a reference to itself as a source which I didn't think made sense

I understand your concern! It does feel a bit weird, but, we have been doing it for the `._artifacts` for a while now and things generally seem to work, so I think I will take that success as motivation for expanding the idea.

> do you want a list of sources that are metadata from previous merges as well?

Yes, I would like to do that (I left that part out of my skeleton, I didn't want to step on your toes too much! :feet:)

Oddant1:
>     @property
>     def _id(self):
>         id_ = 0
>         for source in self._artifacts:
>             id_ = int(id_) ^ int(source.uuid)
> 
>         sum_file = tempfile.NamedTemporaryFile(prefix='md5-')
>         for source in itertools.chain(self._dataframes, self._tsvs):
>             source.save(sum_file.name)
>             id_ = int(id_) ^ int(md5sum(sum_file.name), 16)
>         sum_file.close()
> 
>         return id_

With the current implementation of the `id` property, we want it to return the uuid of the source artifact if its source is only a single artifact, this is something @ebolyen decided upon. I think there are some arguments for changing it namely
1. All metadata ids will look the same if it is changed to a normal hex int
2. This sort of "initialize to 0 then xor with 0" logic can be used simplifying the id calculation

qiime2/core/archive/tests/test_provenance.py
```diff
@@ -100,6 +100,7 @@ def test_chain_with_merged_artifact_metadata(self):
                                              md_artifact2.uuid)
 
         # Check action files for uuid-metadata values
+        # FAIL: This test currently fails right here
```
Oddant1: This test fails, by the looks of it the artifacts on the merged metadata aren't actually being added to provenance, but I'm not too familiar with how this stuff works, and haven't had the time to really dig into why this may be happening

qiime2/metadata/metadata.py
```diff
@@ -469,17 +324,74 @@ def column_count(self):
...
+            for column in self._columns:
+                self._column_names[column] = column
+                self._column_sources[column] = None
+        else:
+            self._column_names = column_names
+            self._column_sources = column_sources
+
+        # This is set post facto at the point of the creation of this metadata
+        # from an artifact
+        self._source_artifact = None
```
Oddant1: Should we make this a kwarg that defaults to `None`?

qiime2/metadata/metadata.py
```diff
@@ -899,8 +815,8 @@ def filter_ids(self, ids_to_keep):
         """
         filtered_df = self._filter_ids_helper(self._dataframe, self.get_ids(),
                                               ids_to_keep)
-        filtered_md = self.__class__(filtered_df)
-        filtered_md._add_artifacts(self.artifacts)
+        filtered_md = self.__class__(filtered_df, self._column_names,
```
Oddant1: Do we want `filter_ids` and `filter_columns` to remove the columns from the filtered metadata's `column_names` and `column_sources`? Removal from `column_sources` would potentially remove artifacts from provenance. If this metadata is the result of a merge, and all columns from one of the metadatas merged in were filtered out, that metadata's `_source_artifact` would no longer be tracked.

qiime2/metadata/metadata.py
```diff
@@ -1036,7 +952,42 @@ def name(self):
...
+                if old_md._source_artifact is not None:
+                    return (old_md._source_artifact,)
+                return ()
+            if new_name in new_md._column_sources:
+                old_md = new_md
+                new_md = new_md._column_sources[new_name]
+            elif old_name in new_md._column_sources:
+                old_md = new_md
+                new_md = new_md._column_sources[old_name]
+            else:
```
Oddant1: Like the comment says, if we somehow reach this point in code something went very wrong, I'm thinking we just let that crash and burn (it shouldn't even be possible anyway), but something made me think putting this here might be worthwhile.

qiime2/metadata/metadata.py
```diff
@@ -1083,7 +1035,11 @@ def __eq__(self, other):
         return (
             super().__eq__(other) and
             self.name == other.name and
-            self._series.equals(other._series)
+            self._series.equals(other._series) and
+            # My initial inclination was to make this test for equalty of
+            # _source_metadata, but this seriously messes with some of the
+            # rountrip tests in test_io, so it looks like this for now
+            self.artifacts == other.artifacts
```
Oddant1: Again, same as comment. Do we want this to test for equality of `artifacts` or equality of `_source_metadata`.?

qiime2/metadata/tests/test_metadata.py
```diff
@@ -413,41 +412,51 @@ def test_no_source_artifacts(self):
...
-        artifact2 = Artifact.import_data('IntSequence1', [1, 2, 3, 4])
-        self.md._add_artifacts([artifact1, artifact2])
-
-        with self.assertRaisesRegex(
-                ValueError, "Duplicate source artifacts.*artifact: Mapping"):
-            self.md._add_artifacts([artifact1])
-
-        # Test that the object hasn't been mutated.
-        self.assertEqual(self.md.artifacts, (artifact1, artifact2))
+#    Because of the changed way of tracking source artifacts, this isn't really
```
Oddant1: Do this test and it's corresponding test in `test_metadata_column.py` serve a purpose at all now? Given the new way artifacts are tracked, this doesn't really seem relevant anymore.

qiime2/metadata/tests/test_metadata.py
```diff
@@ -1393,14 +1324,14 @@ def test_with_artifacts(self):
...
-        md._add_artifacts([artifact1, artifact2])
+        md._add_artifacts((artifact1, artifact2))
 
         obs = md.filter_ids({'a', 'c'})
 
         exp = Metadata(pd.DataFrame(
             {'col1': [1, 3], 'col2': ['foo', 'baz']},
             index=pd.Index(['a', 'c'], name='id')))
-        exp._add_artifacts([artifact1, artifact2])
+        exp._source_artifact = (artifact1, artifact2)
```
Oddant1: This should probably still be `_add_artifacts` not that it matters because this line of code will never be reached because this will break on line 1327. I left this test failing like this because I'm not really sure what to do with it right now. We would need to create a `Metadata` object with both of those artifacts via a merge that didn't contain the column that was filtered out. This is completely doable, but it feels somewhat fishy to me. I'm probably just overthinking this though.

qiime2/metadata/tests/test_metadata.py
```diff
@@ -1625,20 +1556,22 @@ def test_no_artifacts(self):
...
+        md._column_sources = {'col1': artifact1,
+                              'col2': artifact2}
 
         obs = md.filter_columns(column_type='categorical')
 
         exp = Metadata(pd.DataFrame(
             {'col2': ['foo', 'bar', 'baz']},
             index=pd.Index(['a', 'b', 'c'], name='id')))
-        exp._add_artifacts([artifact1, artifact2])
+        exp._add_artifacts(artifact1, artifact2)
```
Oddant1: This goes back to the "should the filters remove filtered columns from `_column_sources`" deal. If filtered columns are removed from `column_sources` then `artifact1` is no longer in the expected metadata's `artifacts` property.

qiime2/metadata/tests/test_metadata_column.py
```diff
@@ -258,49 +258,45 @@ def test_case_insensitive_duplicate_ids(self):
...
+
+        md1 = artifact1.view(Metadata)
+        md2 = artifact2.view(Metadata)
+        md3 = artifact3.view(Metadata)
+        merged_md = md1.merge(md2, md3)
+
+        mdc = merged_md.get_column('e')
+        self.assertEqual(mdc.artifacts, (artifact3,))
+
+#    Same as with test_metadata.py this isn't really a h=thing anymore
```
Oddant1: As the comment says, same as the version of this test in `test_metadata.py`. Is this even relevant now?

qiime2/metadata/tests/test_metadata.py
Oddant1: Same as above goes for the equivalent test in `test_metadata_column.py`

qiime2/core/archive/tests/test_provenance.py
Oddant1: The assert right below this one does pass, it's only this one assertion that fails

qiime2/metadata/metadata.py
```diff
@@ -353,7 +324,49 @@ def column_count(self):
...
+
+        This property is read-only.
+
+        Returns
+        -------
+        qiime2.Artifact
+            Source artifact of the metadata.
+
+        """
+        return self.source_artifact
```
thermokarst: This appears to be a circular reference, should it be the following, instead?

```suggestion
        return self._source_artifact
```

qiime2/metadata/metadata.py
```diff
@@ -353,7 +324,49 @@ def column_count(self):
...
+
+    # Try to ensure the source artifact is actually an artifact
+    def _add_source_artifact(self, artifact):
+        if not isinstance(artifact, qiime2.Artifact):
+            raise TypeError('Source Artifact must a an Artifact object not '
```
thermokarst:
```suggestion
            raise TypeError('Source Artifact must an Artifact object not '
```

qiime2/metadata/metadata.py
```diff
@@ -668,34 +712,80 @@ def merge(self, *others):
...
+        columns = {}
+        df_changes = {}
+
+        mds = [self]
+        mds.extend(others)
+        dupes = set()
+        for md in mds:
+            if md._id in dupes:
+                raise ValueError(
+                    'Your input contained duplicate metadata files. Merging '
```
thermokarst:
```suggestion
                    'Merging '
```

qiime2/metadata/metadata.py
```diff
@@ -353,7 +324,49 @@ def column_count(self):
...
+        return self.source_artifact
+
+    # Try to ensure the source artifact is actually an artifact
+    def _add_source_artifact(self, artifact):
+        if not isinstance(artifact, qiime2.Artifact):
+            raise TypeError('Source Artifact must a an Artifact object not '
+                            f'{artifact}.')
+        self._source_artifact = artifact
+
+    def __init__(self, dataframe, column_names=None, column_sources=None):
```
thermokarst: Should we use a private constructor for handling `column_names` & `column_sources`, instead? I think the current signature will be confusing to Artifact API users.

qiime2/metadata/metadata.py
thermokarst: I think that retroactively altering the provenance of the metadata is a dangerous game to play, maybe better to track in some other property for when these things are filtered out?

qiime2/metadata/metadata.py
thermokarst: I haven't looked closely at this code, but if this endpoint represents a terminal failure condition, I think it would be best to raise an exception.

qiime2/metadata/metadata.py
```diff
@@ -862,7 +952,42 @@ def name(self):
...
+                new_md = new_md._column_sources[new_name]
+            elif old_name in new_md._column_sources:
+                old_md = new_md
+                new_md = new_md._column_sources[old_name]
+            else:
+                # If we get here something really bad happened, should this
+                # else exist?
+                pass
+
+    def __init__(self, series, md=None):
```
thermokarst: Same question as above - should the new kwargs be exposed through a private constructor?

qiime2/metadata/metadata.py
thermokarst: Hmm, let's stick with this at the moment, but please leave a `TODO` note for us to revisit before merging.

qiime2/metadata/tests/test_metadata.py
thermokarst: I agree, this seems to not be relevant, now.

qiime2/metadata/tests/test_metadata.py
```diff
@@ -968,6 +978,7 @@ def test_merging_two(self):
             {'a': [1, 2, 3], 'b': [4, 5, 6],
              'c': [7, 8, 9], 'd': [10, 11, 12]},
             index=pd.Index(['id1', 'id2', 'id3'], name='id')))
+
```
thermokarst: Lots of whitespace changes in this section of this file - was that intentional? If it is for improving readability, let's tackle in a separate (non-squashed) commit.

qiime2/metadata/tests/test_metadata.py
thermokarst: Yeah, this test (`test_with_artifacts`) is probably not relevant now, since we are removing the ability to construct metadata from multiple artifacts at once.

qiime2/metadata/tests/test_metadata.py
thermokarst: Please see my comment, above.

qiime2/metadata/tests/test_metadata_column.py
thermokarst: You could repurpose this test for the `merging` test suite --- worth testing that merging two artifact-based metadata explodes.

qiime2/core/archive/tests/test_provenance.py
thermokarst: Sounds like provenance recording might not be wired up quite right anymore. Maybe one way for you to debug is to reproduce this data in an REPL, then look at `action/action.yaml` to see how things are being recording w.r.t. metadata.

Oddant1:
This is what the metadata looks like in that test in provenance right now. Notice the 'metadata.tsv'

image

This is the value that is supposed to be there

image

Looks like it has somehow gone from 'artifact1.uuid,artifact2.uuid:metadata.tsv' to just 'metadata.tsv'

Oddant1:
I poked around a bit, and I think I know what's causing the failure in `test_provenance.py` the merged metadata column has no artifacts in its `artifacts` property. With the old way of doing things it should have had 2 artifacts in it's `artifacts` list, now it should only have one. Instead it has nothing. I'm not sure why because there are tests in `test_metadata_column.py` ensuring that columns taken from merged metadata retain their correct artifact.

thermokarst:
I am co-assigning this PR to me, I will pull the code down and take a closer look at your question about the test failures. In the meantime you can continue to work on addressing my latest round of review.

Oddant1:
I'm pretty sure I have new commits addressing all of those? I think I've identified the issue anyway. I never finished the `artifacts` property for metadata columns, I got side tracked and that code got left in a mangled state.

Oddant1:
Note, as far as the bug in provenance is concerned, the previous test had two artifacts making up the source of the merged metadata column, we now only have one. Previously, the column had both artifacts the merged metadata it came from had, now it only has the artifact that produced the metadata it originally came from prior to any merging. If we do still want both of those artifacts there, we're going to need to change our approach

Oddant1:
Ok, The provenance issue is not fixed, in fact it seems more bizarre now. It looks like sometimes it tracks both artifacts, and sometimes it doesn't? I'll leave that to you @thermokarst and do another pass over your review. I think the changes I made in my last commit made things a bit better.

Oddant1:
Additionally, the test failure is now on the next assertion. It used to fail on the assertion on line 104, now it is failing on the assertion on line 107 (when it fails)

Oddant1:
A bit late, but for clarification that (I think) at the end of https://github.com/qiime2/qiime2/pull/464/commits/340c51304612671b4a1fa3975ee212699782d0e6 is confirmed. The last commit did fix the issue.

lizgehret:
Hey @Oddant1 - I'm going back through old PRs to try and get some things cleaned up and merged. Can you resolve the conflicts above? Once that's fixed and pending successful CI, we can get a final review on this and get it merged!

qiime2/metadata/metadata.py
```diff
@@ -438,6 +409,48 @@ def column_count(self):
...
+        qiime2.Artifact
+            Source artifact of the metadata.
+
+        """
+        return self._source_artifact
+
+    # Try to ensure the source artifact is actually an artifact
+    def _add_source_artifact(self, artifact):
+        if not isinstance(artifact, qiime2.Artifact):
+            raise TypeError('Source Artifact must an Artifact object not '
```
Oddant1, line 450:
```suggestion
            raise TypeError('Source Artifact must be an Artifact object not '
```

qiime2/metadata/metadata.py
```diff
@@ -552,7 +598,8 @@ def __eq__(self, other):
         return (
             super().__eq__(other) and
             self._columns == other._columns and
-            self._dataframe.equals(other._dataframe)
+            self._dataframe.equals(other._dataframe) and
+            self._id == other._id
```
Oddant1, line 602: Getting the ids can involve getting the md5sum of a file which can be expensive

gregcaporaso:
We've decided to close this PR, now that we have a `merge` action in q2-metadata. That action supports merging of metadata when there are either overlapping ids or overlapping columns, but not both. This was partly enabled by the addition of the `ImmutableMetadata` type, which didn't exist when work on this PR started. 

The underlying goal of this PR is still of interest (merge on the fly, without creation of a new artifact, and when there are potentially overlapping ids and overlapping columns), but it's considerably more work than adding the new action. We may circle back to reference this PR if we decide we need it in the future.

PR #466, opened 2019-07-10
BUG: Improve error message for overlong metadata column names

### LINKED ISSUE #463: alpha-group-significance breaks on long metadata column headers
**Bug Description**
`qiime diversity alpha-group-significance` produces an OSError: `[Errno 36] File name too long` when users pass in metadata with column header(s) that exceed os/filesystem length limits. 

1. Modify a metadata file so that one column header is greater than 256 characters long. (This should cover most if not all file systems). 
2. Pass this file to `qiime diversity alpha-group-significance` as metadata. (Other input data can be downloaded from [Moving Pictures](https://docs.qiime2.org/2019.4/tutorials/moving-pictures/#alpha-and-beta-diversity-analysis)

**Expected behavior**
QIIME 2 should handle this gracefully, likely by catching the over-long string and rendering a helpful error message. 

**Computation Environment**
Unknown linux/mac

**Questions**
1. Does this issue affect other commands/is this worth dealing with more broadly?
2. Are we serving users better by failing with a clear error message (and encouragement to clean up their metadata), or by "fixing" the issue and proceeding (e.g. by truncating overlong column-header strings when found).  

**Comments**
I'm inclined to let the command fail loudly, because I doubt there are many use cases in which columns benefit from 200-character titles. I'd welcome more-experienced opinions on this.

**References**
1. [forum xref](https://forum.qiime2.org/t/oserror-errno-36-file-name-too-long/10226)

Oddant1:
Closes #463. Currently all this does is check if any of the metadata column names are >200 characters long and tells the user they're too long if they are. Two potential issues with this approach are:
1. If the directory the file is saving to has a filename length limit of less than 200 (not likely)
2. If whatever we add to the header name before making it a filename makes it longer than acceptable  (example https://github.com/qiime2/q2-diversity/blob/88d3f42b34b2b8e74e79c2a48b3ce794aa596f1d/q2_diversity/_alpha/_visualizer.py#L113), but this is also unlikely because the maximum filename length seems to generally be 255 characters and I doubt we'll ever be adding 56+ characters.
I'll add some tests to this once we've finalized an approach. @ebolyen proposed some potentially more sweeping changes to this sort of system when this issue was initially brought up, so we will see.

### DIFF
--- a/qiime2/metadata/io.py
+++ b/qiime2/metadata/io.py
@@ -48,7 +48,6 @@ class MetadataReader:
 
         self._filepath = filepath
 
-        # Used by `read()` to store an iterator yielding rows with
         # leading/trailing whitespace stripped from their cells (this is a
         # preprocessing step that should happen with *every* row). The iterator
         # protocol is the only guaranteed API on this object.
@@ -183,7 +182,10 @@ class MetadataReader:
                     "Metadata column name %r conflicts with a name reserved "
                     "for the ID column header. Reserved ID column headers:"
                     "\n\n%s" % (column_name, FORMATTED_ID_HEADERS))
-
+            if len(column_name) > 200:
+                raise MetadataFileError(
+                    f"Metadata column name {column_name} exceeds maximum "
+                    "column name length of 200")
         return header
 
     def _read_directives(self, header):
--- a/qiime2/metadata/io.py
+++ b/qiime2/metadata/io.py
@@ -184,7 +184,7 @@ class MetadataReader:
                     "\n\n%s" % (column_name, FORMATTED_ID_HEADERS))
             if len(column_name) > 200:
                 raise MetadataFileError(
-                    f"Metadata column name {column_name} exceeds maximum "
+                    f"Metadata column name {column_name!r} exceeds maximum "
                     "column name length of 200")
         return header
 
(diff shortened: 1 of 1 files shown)


qiime2/metadata/io.py
```diff
@@ -48,7 +48,6 @@ def __init__(self, filepath):
 
         self._filepath = filepath
 
-        # Used by `read()` to store an iterator yielding rows with
```
thermokarst: Oops, looks like you deleted part of this pre-existing comment in your changeset.

Oddant1: @thermokarst these attempts at using vim commands have me doing that a lot lately :laughing:

thermokarst:
Thanks for putting this together, @Oddant1. I don't want to limit the length of metadata column names, that seems unnecessarily restrictive, in my opinion. What if we added a new `hash` property on metadata columns, then updated actions (like the one implicated in https://github.com/qiime2/qiime2/issues/463) to use the hash of the name, instead? This would be beneficial too, because we have often run into issues with actions (visualizers, in particular), producing invalid filepaths due to "special" characters leaking through.

Oddant1:
@thermokarst given the changes currently being made to Metadata and Metadata Columns in #464 I'm inclined to wait until that's done to make further changes to Metadata Columns

thermokarst:
Closing this to make way for a future implementation. Thanks!

PR #471, opened 2019-07-22
HACK: pinning pandas 0.24.2 for 2019.7 release

thermokarst:
0.25.x is out, but let's save that jump for the 2019.10 release cycle...


thermokarst:
I like that I linked to a pandas PR above. Yikes...

PR #473, opened 2019-07-23
IMP: Add ability to deprecate plugin functionality

Oddant1:
Begins to implement ability to deprecate plugin functionality

### DIFF
--- a/qiime2/core/testing/method.py
+++ b/qiime2/core/testing/method.py
@@ -45,6 +45,10 @@ def no_input_method() -> dict:
     return {'foo': 42}
 
 
+def deprecated_method() -> dict:
+    return {'foo': 43}
+
+
 def long_description_method(mapping1: dict, name: str, age: int) -> dict:
     return {name: age}
 
--- a/qiime2/core/testing/plugin.py
+++ b/qiime2/core/testing/plugin.py
@@ -345,7 +345,8 @@ dummy_plugin.methods.register_function(
         ('out', Mapping)
     ],
     name='Parameters only method',
-    description='This method only accepts parameters.'
+    description='This method only accepts parameters.',
+    deprecated=True
 )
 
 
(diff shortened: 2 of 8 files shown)


thermokarst:
@ebolyen, what do you think about making this new `deprecated` flag a simple boolean value for now? If it is True, then the sdk (action.py) could issue a `FutureWarning` about the method being deprecated in "some future version of the plugin." Interfaces can catch the warning and do something fancy, if they care. No hiding of methods or anything at this stage, just a nudge that "hey, this action is disappearing soon." (note, the code in this PR is WIP, I don't think @Oddant1 and @Davrod29 have implemented any of this, yet).

In the future, this param could become a version string, representing the plugin version that will deprecate the action. If the version identifier was PEP440-compatible, we could do some math on the value and compare to the current plugin version --- if less than the dep ver, warnings for the user, if gte, hide the action and a stricter warning.

Just spitballing ideas, mostly just want to get your approval for a minimal changeset for 2019.7 (which, I _think_ is the bool flag). Thanks!

ebolyen:
I agree with all of this. The flag is just fine for now, and PEP 440 support would be great. And it seems a colored deprecation warning is the easiest and most obvious thing to do.

Maybe for now we leave it at that, and then I think in the future we can get fancier with the plugin help where we could split the deprecated commands out in the list

thermokarst:
@Oddant1 & @Davrod29 - can you two:

1. Fix the conflict
2. Change the warning type to a `FutureWarning`? 

Thanks!

Oddant1:
@thermokarst With the conflicts, do we want both `deprecated_method` and `unioned_primitives`

thermokarst:
Hey @antgonza & @wasade, just a head's up, this PR introduces a new warning that interfaces will want to be aware of. Right now this changeset will raise a `FutureWarning` when deprecated plugin actions are executed. You might want to consider catching these Warnings in Qiita, although at the moment the API for this is still very fresh (i.e. we probably need to think about creating a new class of `QIIME2Warning`, deprecation targets, deprecating params/inputs/outputs, etc.). Just wanted to get this on your radar as soon as possible.

I am going to work on drafting up a PR for q2cli that makes that interface aware of this new deprecation warning --- that will be a good changeset to look at if you decide to throw this into Qiita.

PR #474, opened 2019-07-24
definitive QIIME 2 citation

nbokulich:


nbokulich:
collapsed to one line — I agree that citation looks ugly in q2view

PR #483, opened 2019-08-12
IMP: Make path to BoundFile easily accessible through path attribute

### LINKED ISSUE #481: The file path to a `BoundFile` object should be easily accessible
**Improvement Description**
Make `BoundFile` filepaths easily accessible.

**Current Behavior**
Their doesn't seem to be any easy way to access the path to a `BoundFile` object. Their repr just shows them as a Python object, and their `name` attribute isn't a filepath either.

**Proposed Behavior**
Create an easy way to access the path to a `BoundFile`

**Question**
1. Are there any technical limitations restricting the ability to make this a thing?

**Comments**
1. Probably just do this via the exposure of a new `path` parameter.

Oddant1:
Closes #481 by exposing a new `path` parameter on `BoundFile` objects that makes the filepath directly accessible without having to do any digging.

### DIFF
--- a/qiime2/plugin/model/directory_format.py
+++ b/qiime2/plugin/model/directory_format.py
@@ -118,6 +118,10 @@ class BoundFile:
             return path
         return bound_path_maker
 
+    @property
+    def path(self):
+        return str(self._directory_format.path)
+
 
 class BoundFileCollection(BoundFile):
     def __init__(self, name, pathspec, format, directory_format, path_maker):
--- a/qiime2/plugin/model/directory_format.py
+++ b/qiime2/plugin/model/directory_format.py
@@ -8,6 +8,7 @@
 
 import re
 import pathlib
+from pathlib import Path
 
 from qiime2.core import transform
 from .base import FormatBase, ValidationError, _check_validation_level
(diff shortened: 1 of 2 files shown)


Oddant1:
Looks like Travis kinda died during the testing of this PR? Unittests for this code will be added in future commits by the way.

Oddant1:
As far as those unittests are concerned, I'm not really sure where to put them. I'm not really seeing tests for `directory_format.py` at all. There's a fairly limited `test_file_format.py`, but no `test_directory_format.py`

qiime2/plugin/model/directory_format.py
```diff
@@ -118,6 +118,10 @@ def bound_path_maker(**kwargs):
             return path
         return bound_path_maker
 
+    @property
+    def path(self):
```
ebolyen: Drive-by-comment on the name, we use `.path` for `FileFormat` and it is a `pathlib` object instead of a str. We might want to make that the same here. Also it would be nice if we got the file's actual path, instead of the directory format's path.

Oddant1: The paths look something like this '/tmp/qiime2-test-tempmdun4tfw/file', is this not the path to the file?

Oddant1: Also, the default type of `self._directory_format.path` is `qiime2.core.path.InPath`, should I just return that, or should I convert it into a `pathlib.Path`

thermokarst:
Please see my comment on the parent issue: https://github.com/qiime2/qiime2/issues/481#issuecomment-523152365

PR #484, opened 2019-08-12
IMP: Add .view method to _FileFormat

### LINKED ISSUE #482: `_FileFormat` Should Have A `.view` Method
**Improvement Description**
`_FileFormat` should have a `.view` method

**Current Behavior**
`BoundFile` objects have a `.view` method allowing for easy transformation between classes, but `_FileFormat` objects do not. This means the easiest way to transform `_FileFormat` objects is to create a directory format with them as a bound file and then use their `.view`.

**Proposed Behavior**
Add a `.view` method to `_FileFormat`

Oddant1:
Closes #482 `_FileFormat` objects now have a `.view` method allowing for easy transformation of `_FileFormat` subclasses without them needing to be turned into `BoundFile` objects. I am unsure of how to test this code. Should `.view` tests be added to the tests for every `_FileFormat` subclass testing each of their transformers? Should the `dummy_plugin` be used to test the transformers of the subclasses of `_FileFormat` in `dummy_plugin`?

### DIFF
--- a/qiime2/plugin/model/file_format.py
+++ b/qiime2/plugin/model/file_format.py
@@ -8,6 +8,7 @@
 
 import abc
 
+from qiime2.core import transform
 from .base import FormatBase, ValidationError, _check_validation_level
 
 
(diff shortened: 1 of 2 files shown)


ebolyen:
This is a very nice way to solve the problem, I like it!

PR #485, opened 2019-08-20
IMP: Metadata constructor now strips whitespace

### LINKED ISSUE #478: Allow Metadata constructor(s) to optionally strip whitespace
**Improvement Description**
A kwarg in the various Metadata constructors to strip whitespace in all columns could be really convenient for transformers et al.

**Current Behavior**
Whitespace is not stripped.

**Proposed Behavior**
If some kwarg (say, `strip_whitespace`) was True, the MD constructor would strip whitespace.

**Questions**
1. This could also just be the default behavior, right? Am I missing something here?

Oddant1:
Closes #478. Strips all white space from all pieces of the Metadata

### DIFF
--- a/qiime2/metadata/metadata.py
+++ b/qiime2/metadata/metadata.py
@@ -156,11 +156,6 @@ class _MetadataBase:
                     "Detected empty metadata %s. %ss must consist of at least "
                     "one character." % (label, label))
 
-            if value != value.strip():
-                raise ValueError(
-                    "Detected metadata %s with leading or trailing "
-                    "whitespace characters: %r" % (label, value))
-
             if axis == 'id' and value.startswith('#'):
                 raise ValueError(
                     "Detected metadata %s that begins with a pound sign "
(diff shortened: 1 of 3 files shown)


qiime2/metadata/metadata.py
```diff
@@ -877,7 +876,11 @@ def __init__(self, series):
                 "%s %r does not support a pandas.Series object with dtype %s" %
                 (self.__class__.__name__, series.name, series.dtype))
 
+        series.index = series.index.str.strip()
+        series.name = series.name.strip()
         self._series = self._normalize_(series)
+        if isinstance(self, CategoricalMetadataColumn):
```
ebolyen: This particular check would be unnecessary if the below branch was a part of `_normalize_` as method dispatch solves the "who" we are problem.

qiime2/metadata/metadata.py
```diff
@@ -361,11 +356,15 @@ def __init__(self, dataframe):
 
         super().__init__(dataframe.index)
 
+        self._validate_index(dataframe.columns, axis='column')
+        dataframe.index = dataframe.index.str.strip()
```
ebolyen: This actually mutates the input, which is usually something to avoid. If you look inside `_normalize_dataframe` you'll see it creates a copy to avoid changing the input.

qiime2/metadata/metadata.py
```diff
@@ -877,7 +876,11 @@ def __init__(self, series):
                 "%s %r does not support a pandas.Series object with dtype %s" %
                 (self.__class__.__name__, series.name, series.dtype))
 
+        series.index = series.index.str.strip()
+        series.name = series.name.strip()
```
ebolyen: Same situation here re: mutation.

Oddant1:
@ebolyen damn, I committed the cardinal sin of mutating things I shouldn't be where I shouldn't be :laughing:

qiime2/metadata/metadata.py
Oddant1: But we want to normalize the `series` prior to stripping don't we? I've been poking at it, and unless I'm missing something this isn't really doable in `_normalize_` given its current implementation with the inner function and all

Oddant1: Wait I've had an epiphany and realized I am dumb, nevermind.

Oddant1:
Looks good to me

PR #489, opened 2019-08-23
IMP: convert MetadataColumn to Metadata

### LINKED ISSUE #487: allow `MetadataColumn` to convert to `Metadata`
**Improvement Description**
Add a method to `MetadataColumn` to convert to `Metadata`.

**Current Behavior**
There is no direct conversion, so need to use `Metadata(column.to_dataframe())`

**Proposed Behavior**
Either add a method to convert to Metadata, or alternatively MetadataColumn should be substitutable for Metadata in all cases.

**References**
For an example use case, see discussion here: https://github.com/qiime2/q2-sample-classifier/pull/168#discussion_r316852560

Oddant1:
Closes #487 Adds a method to convert `MetadataColumns` to `Metadata`

### DIFF
--- a/qiime2/metadata/metadata.py
+++ b/qiime2/metadata/metadata.py
@@ -994,6 +994,16 @@ class MetadataColumn(_MetadataBase, metaclass=abc.ABCMeta):
         """
         return self._series.to_frame()
 
+    def to_metadata(self):
+        """Create a metadata object from the metadata column
+
+        Returns
+        -------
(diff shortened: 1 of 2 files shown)


lizgehret:
How do we want to handle this one and [#464] @ebolyen?

ebolyen:
Closing as stale. This should be updated someday to handle `missing_schemes` and `_artifacts`.

PR #491, opened 2019-08-30
HACK: Unpin pandas to v0.25.1 to determine plugin issues

David-Rod:
Unpin pandas to determine build issues regarding QIIME 2 plugins and busywork.

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -20,7 +20,6 @@ requirements:
     - python {{ python }}
     - pyyaml
     - decorator
-    - pandas 0.25.1
     - tzlocal
     - python-dateutil
     - pyparsing ==2.3.0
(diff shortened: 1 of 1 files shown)


ci/recipe/meta.yaml
```diff
@@ -21,7 +21,7 @@ requirements:
     - pyyaml
     - decorator
     # TODO: unset this pin once pandas 0.25.x is sorted out
-    - pandas 0.24.2
+    - pandas 0.25.1
```
thermokarst: Thanks @David-Rod! Can you update this to _unpin_ pandas? The proposed change just pins to a different version of pandas. If there is _no pin_ specified, conda will usually grab the newest version. Thanks

thermokarst: Also, please remove the `# TODO` comment above. Thanks!

PR #494, opened 2019-09-11
IMP: TestPlugin.transform_format now validates output

### LINKED ISSUE #492: `TestPluginBase.transform_format` does not validate it output
**Improvement Description**
Validate the output of `TestPluginBase.transform_format`

**Current Behavior**
The standard transformation API ensures the validity of the formats output by the transformer., but `TestPluginBase.transform_format` skips the API and calls the transfomer directly allowing for transformers tested in this manner to potentially produce invalid outputs.

**Proposed Behavior**
Ensure outputs of transformers called via `TestPluginBase.transform_format` are valid

**References**
1. This was first noticed by @ebolyen in the issue qiime2/q2-types#226

Oddant1:
Closes #492 with a minimum working example of `TestPlugin.transform_format` validating its output, funnily enough this addition caused 2 test failures in current q2-types tests, on top of catching the instance in qiime2/q2-types#226 that led to this PR.

I'm not sure why this needs to be a thing though. Is there any reason we can't just transform the test data via the normal transformation API?

Also, in this PR, validation is being run on max because that is the default level, but all tests fail at max or min. Is there any reason we might want to validate on min here and not max?


thermokarst:
> I'm not sure why this needs to be a thing though. Is there any reason we can't just transform the test data via the normal transformation API?

Not to my knowledge (we probably just did it "wrong" the first time), and I suspect that is what @ebolyen was driving at in https://github.com/qiime2/q2-types/pull/226#discussion_r319311408

Oddant1:
Awesome, so should this `transform_format` method just be changed into a more normal implementation of the transformation API (leading to it min validating everything with a `.validate`)? Additionally, if the q2-types tests are anything to go off of, this could lead to some generalized test failures due to `.validate` failures. Not like that isn't something that shouldn't be fixed, but it's something to keep in mind in case a significant number of things break.

thermokarst:
Hey there @ChrisKeefe, can you please review this PR when you get a chance? Please address any changes you might want made to @ebolyen & I - we are teaming up to run this one across the finish line!

ebolyen:
@thermokarst and/or I will add a `validate="max"` before and after the transformation

thermokarst:
@ebolyen, this is ready for review (please take a look at https://github.com/qiime2/q2-types/pull/234 before making any decisions as to the fate of this PR). Thanks!

thermokarst:
I believe that the conclusion of this investigation was that we actually _want_ to skip the validation machinery, that way tests can do things with bad data, easily. The compromise is that test data needs to be deliberately selected and manually verified, which is a bit error-prone. See #527 for a superseding PR.

PR #495, opened 2019-09-18
IMP: Max validate FileFormat on construction

### LINKED ISSUE #490: _FileFormat should self-validate on construction
**Improvement Description**
It is currently possible to create invalid file formats via the format constructor. This isn't generally a problem, since users typically "import" their data, which _does_ call the format validator. One common place that formats are instantiated via the constructor though is in unit tests, and as such, if the caller forgets to call `.validate()` on the instance, they can potentially be working with "invalid" data.

**Current Behavior**
The `_FileFormat` constructor knows nothing about `.validate()`.

**Proposed Behavior**
The `_FileFormat`  constructor calls `.validate()` as part of its set up.

**Questions**
1. I suspect we left this off of the initial validate API because we were hesitant to start validating data willy-nilly, before understanding the performance implications. We have slowly been filling in some [validation gaps](https://github.com/qiime2/qiime2/pull/461), but I wonder if I am forgetting something else about this.

Oddant1:
Closes #490 

This PR should get the ball rolling on validating  `_FileFormat` on construction, it should also serve to expose some of the issues currently facing this idea. This current PR works by max validating all objects created in 'r' mode using the `FormatBase` constructor
1. First of all as @ebolyen said here https://github.com/qiime2/qiime2/issues/490#issuecomment-530002158 we may not want to max validate every `FormatBase` that comes out of a transformer, we already min validate all of them by default when using the transformation API. I'm not sure why we chose min over max though. I suppose time efficiency.
2. We can no longer create invalid test data of any `FormatBase` derivative. This is only problematic when we are intentionally creating invalid objects, and usually all it means is putting the object creation inside of the `assertRaisesRegex` instead of an explicit `object.validate`, but it does prevent us from actually doing anything with these intentionally broken `FormatBase` objects. For example, one of the tests I changed created an object, asserted that it passed min validation, and then that it failed max validation. This cannot be done anymore as max validation is called on creation.
3. As a result of point 2, all tests using intentionally broken `FormatBase` objects in all repos (no idea how many of these there are) will fail and then have to be changed. Fortunately, all tests using unintentionally broken `FormatBase` objects will also fail.
4. This only works to validate objects created in 'r' mode because those created in 'w' mode have nothing to validate and will always fail until after initialization as they have yet to be set up in any meaningful way.

### DIFF
--- a/qiime2/core/format.py
+++ b/qiime2/core/format.py
@@ -19,6 +19,8 @@ class FormatBase:
             if mode != 'r':
                 raise ValueError("A path must be omitted when writing.")
 
+        self._mode = mode
+
         if mode == 'w':
             self.path = qpath.OutPath(
                 # TODO: parents shouldn't know about their children
@@ -26,8 +28,10 @@ class FormatBase:
                 prefix='q2-%s-' % self.__class__.__name__)
         else:
             self.path = qpath.InPath(path)
-
-        self._mode = mode
+            # Check for _validate_ because some classes have a dummy
+            # implementation of validate to raise NotImplemented errors
+            if hasattr(self, '_validate_'):
+                self.validate()
 
     def __str__(self):
         return str(self.path)
--- a/qiime2/core/format.py
+++ b/qiime2/core/format.py
@@ -19,6 +19,8 @@ class FormatBase:
             if mode != 'r':
                 raise ValueError("A path must be omitted when writing.")
 
+        # Mode has to be set here so objects don't fail validation for not
+        # having a `.mode` attribute
         self._mode = mode
 
         if mode == 'w':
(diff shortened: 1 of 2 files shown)


qiime2/core/format.py
```diff
@@ -19,15 +19,19 @@ def __init__(self, path=None, mode='w'):
             if mode != 'r':
                 raise ValueError("A path must be omitted when writing.")
 
+        self._mode = mode
```
thermokarst: DBC: Can you explain why this move was necessary, it isn't immediately clear to me? Perhaps putting a comment inline would help, too.

Oddant1: I added a comment in the new commit. Basically with `self._mode = mode` down at the bottom, we would get validation errors due to objects not have a `._mode` field.

ebolyen:
closing in favor of https://github.com/qiime2/qiime2/issues/492

PR #497, opened 2019-10-11
NEW: Expose plugin level importable types set through sdk

Oddant1:
Exposes plugin level list (set technically) of importable types for easy access in coming changes to docs/cli/artifact API. I can make this happen with formats as well on this or another branch.

### DIFF
--- a/qiime2/plugin/plugin.py
+++ b/qiime2/plugin/plugin.py
@@ -207,8 +207,8 @@ class Plugin:
             format=artifact_format, plugin=self)
 
         self.type_formats.append(record)
-        for type in record.type_expression:
-            self.importable_types.add(type)
+        for type_ in record.type_expression:
+            self.importable_types.add(type_)
 
 
 class PluginActions(dict):
--- a/qiime2/plugin/plugin.py
+++ b/qiime2/plugin/plugin.py
@@ -70,6 +70,7 @@ class Plugin:
         self.types = {}
         self.transformers = {}
         self.type_formats = []
+        self.importable_types = set()
 
     @property
     def actions(self):
@@ -201,9 +202,13 @@ class Plugin:
                 raise ValueError("%r has a predicate, differentiating format"
                                  " on predicate is not supported.")
 
-        self.type_formats.append(TypeFormatRecord(
+        record = TypeFormatRecord(
             type_expression=semantic_type,
-            format=artifact_format, plugin=self))
+            format=artifact_format, plugin=self)
+
+        self.type_formats.append(record)
+        for type in record.type_expression:
+            self.importable_types.add(type)
 
 
 class PluginActions(dict):
--- a/qiime2/plugin/plugin.py
+++ b/qiime2/plugin/plugin.py
@@ -22,7 +22,7 @@ TransformerRecord = collections.namedtuple(
 SemanticTypeRecord = collections.namedtuple(
     'SemanticTypeRecord', ['semantic_type', 'plugin'])
 SemanticTypeFragmentRecord = collections.namedtuple(
-    'SemanticTypeRecord', ['semantic_type', 'plugin'])
+    'SemanticTypeRecord', ['fragment', 'plugin'])
 FormatRecord = collections.namedtuple('FormatRecord', ['format', 'plugin'])
 ViewRecord = collections.namedtuple(
     'ViewRecord', ['name', 'view', 'plugin', 'citations'])
@@ -184,7 +184,7 @@ class Plugin:
                 raise ValueError("%r is not a semantic type symbol."
                                  % type_fragment)
 
-            if type_fragment.name in self.types:
+            if type_fragment.name in self.type_fragments:
                 raise ValueError("Duplicate semantic type symbol %r."
                                  % type_fragment)
 
--- a/qiime2/plugin/plugin.py
+++ b/qiime2/plugin/plugin.py
@@ -21,6 +21,8 @@ TransformerRecord = collections.namedtuple(
     'TransformerRecord', ['transformer', 'plugin', 'citations'])
 SemanticTypeRecord = collections.namedtuple(
     'SemanticTypeRecord', ['semantic_type', 'plugin'])
+SemanticTypeFragmentRecord = collections.namedtuple(
+    'SemanticTypeRecord', ['semantic_type', 'plugin'])
 FormatRecord = collections.namedtuple('FormatRecord', ['format', 'plugin'])
 ViewRecord = collections.namedtuple(
     'ViewRecord', ['name', 'view', 'plugin', 'citations'])
@@ -67,7 +69,7 @@ class Plugin:
 
         self.formats = {}
         self.views = {}
-        self.types = {}
+        self.type_fragments = {}
         self.transformers = {}
         self.type_formats = []
         self.importable_types = set()
@@ -171,23 +173,24 @@ class Plugin:
             # Apply the decorator as we were applied with a single function
             return decorator(_fn)
 
-    def register_semantic_types(self, *semantic_types):
-        for semantic_type in semantic_types:
-            if not is_semantic_type(semantic_type):
-                raise TypeError("%r is not a semantic type." % semantic_type)
+    def register_semantic_type_fragments(self, *semantic_type_fragments):
+        for type_fragment in semantic_type_fragments:
+            if not is_semantic_type(type_fragment):
+                raise TypeError("%r is not a semantic type." % type_fragment)
 
-            if not (isinstance(semantic_type, grammar.IncompleteExp) or
-                    (semantic_type.is_concrete() and
-                    not semantic_type.fields)):
+            if not (isinstance(type_fragment, grammar.IncompleteExp) or
+                    (type_fragment.is_concrete() and
+                    not type_fragment.fields)):
                 raise ValueError("%r is not a semantic type symbol."
-                                 % semantic_type)
+                                 % type_fragment)
 
-            if semantic_type.name in self.types:
+            if type_fragment.name in self.types:
                 raise ValueError("Duplicate semantic type symbol %r."
-                                 % semantic_type)
+                                 % type_fragment)
 
-            self.types[semantic_type.name] = SemanticTypeRecord(
-                semantic_type=semantic_type, plugin=self)
+            self.type_fragments[type_fragment.name] = \
+                SemanticTypeFragmentRecord(
+                    fragment=type_fragment, plugin=self)
 
     def register_semantic_type_to_format(self, semantic_type, artifact_format):
         if not issubclass(artifact_format, DirectoryFormat):
--- a/qiime2/plugin/plugin.py
+++ b/qiime2/plugin/plugin.py
@@ -72,7 +72,7 @@ class Plugin:
         self.type_fragments = {}
         self.transformers = {}
         self.type_formats = []
-        self.importable_types = set()
+        self.types = set()
 
     @property
     def actions(self):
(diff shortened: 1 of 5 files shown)


qiime2/plugin/plugin.py
```diff
@@ -201,9 +202,13 @@ def register_semantic_type_to_format(self, semantic_type, artifact_format):
...
 
-        self.type_formats.append(TypeFormatRecord(
+        record = TypeFormatRecord(
             type_expression=semantic_type,
-            format=artifact_format, plugin=self))
+            format=artifact_format, plugin=self)
+
+        self.type_formats.append(record)
+        for type in record.type_expression:
+            self.importable_types.add(type)
```
thermokarst: `type` -> `type_`

Oddant1:
https://github.com/qiime2/qiime2/pull/497/commits/b99fc68e3cf6f7d050d6e79f17bec29653260ad4 breaks a significant number of tests due to changes in the names of properties

qiime2/plugin/plugin.py
```diff
@@ -67,9 +69,10 @@ def __init__(self, name, version, website, package, citation_text=None,
 
         self.formats = {}
         self.views = {}
-        self.types = {}
+        self.type_fragments = {}
         self.transformers = {}
         self.type_formats = []
+        self.importable_types = set()
```
thermokarst: Please rename:

```suggestion
        self.types = set()
```

:pineapple:

qiime2/plugin/plugin.py
```diff
@@ -170,23 +173,24 @@ def decorator(transformer):
             # Apply the decorator as we were applied with a single function
             return decorator(_fn)
 
-    def register_semantic_types(self, *semantic_types):
-        for semantic_type in semantic_types:
-            if not is_semantic_type(semantic_type):
-                raise TypeError("%r is not a semantic type." % semantic_type)
+    def register_semantic_type_fragments(self, *semantic_type_fragments):
```
thermokarst: Please find and replace globally across the entire framework:

`s/register_semantic_types/register_semantic_type_fragments/g`

This rename is at least part of what is causing travis to fail at the moment. You should be able to verify by running the tests locally. :baggage_claim:

qiime2/plugin/plugin.py
```diff
@@ -201,9 +205,13 @@ def register_semantic_type_to_format(self, semantic_type, artifact_format):
...
 
-        self.type_formats.append(TypeFormatRecord(
+        record = TypeFormatRecord(
             type_expression=semantic_type,
-            format=artifact_format, plugin=self))
+            format=artifact_format, plugin=self)
+
+        self.type_formats.append(record)
+        for type_ in record.type_expression:
+            self.importable_types.add(type_)
```
thermokarst: Don't forget to update this when you rename `importable_types` to `types`! Also, instead of adding the type to this property, how about you create and add a `SemanticTypeRecord`, instead (I will explain more detail below, in `plugin_manage.py`).

qiime2/sdk/plugin_manager.py
```diff
@@ -52,6 +52,7 @@ def _init(self):
         self.formats = {}
         self.views = {}
         self.type_formats = []
+        self.importable_types = set()
```
thermokarst: Please rename and change the data type:

```suggestion
        self.types = {}
```

:mount_fuji:

qiime2/sdk/plugin_manager.py
```diff
@@ -100,11 +101,28 @@ def _integrate_plugin(self, plugin):
 
             self.formats[name] = record
         self.type_formats.extend(plugin.type_formats)
+        self.importable_types.update(plugin.importable_types)
```
thermokarst: Don't forget to update to the new datatype. Let's use the `plugin.name` value as the key (although I am open to other ideas, like using the plugin object outright, instead).

```suggestion
        self.types[plugin.name] = plugin.types
```

qiime2/sdk/plugin_manager.py
```diff
@@ -100,11 +101,28 @@ def _integrate_plugin(self, plugin):
...
 
     # TODO: Should plugin loading be transactional? i.e. if there's
     # something wrong, the entire plugin fails to load any piece, like a
     # databases rollback/commit
 
+    @property
```
thermokarst: `get_semantic_types` should be a method (a callable function on an instance of `PluginManager`, not a property. Please remove this!

qiime2/sdk/plugin_manager.py
```diff
@@ -100,11 +101,28 @@ def _integrate_plugin(self, plugin):
...
+    def get_semantic_types(self, *, plugin=None):
+        from qiime2.plugin import SemanticTypeRecord
+
+        types = set()
+
+        plugins = [plugin] if plugin else self.plugins
+
+        for plugin in plugins:
+            for name, record in plugin.type_fragments.items():
```
thermokarst: Please refactor this to grab the `SemanticTypeRecord`s directly off of the plugin manager's `self.types` property. Since we discussed pushing this logic into `plugin.py` above, the records will already be created and ready for use by the time you get here. Please let me know if you need any clarification!

:electric_plug:

Oddant1:
https://github.com/qiime2/qiime2/pull/497/commits/65b9821d106df35c641c6cfb200b3038a42fc041 of course, I fix the tests and Travis decides it wants to not work and fail anyway :laughing:

qiime2/plugin/plugin.py
```diff
@@ -67,9 +69,10 @@ def __init__(self, name, version, website, package, citation_text=None,
 
         self.formats = {}
         self.views = {}
-        self.types = {}
+        self.type_fragments = {}
         self.transformers = {}
         self.type_formats = []
+        self.types = set()
```
ChrisKeefe: This is a set, because a type may be registered to more than one format, but these types should not be duplicated in our collection?

qiime2/plugin/plugin.py
```diff
@@ -201,9 +205,14 @@ def register_semantic_type_to_format(self, semantic_type, artifact_format):
...
                                  " on predicate is not supported.")
 
-        self.type_formats.append(TypeFormatRecord(
+        record = TypeFormatRecord(
             type_expression=semantic_type,
-            format=artifact_format, plugin=self))
+            format=artifact_format, plugin=self)
+
+        self.type_formats.append(record)
+        for type_ in record.type_expression:
```
ChrisKeefe: When we iterate over record.type_expression, we are stepping through a "semantic type". Does this mean each iteration is a type fragment? If so, is it appropriate to add these to `self.types` instead of `self.type_fragments`? 

If we're stepping over the `semantic_type` itself, is iteration necessary at all?

qiime2/plugin/plugin.py
Oddant1: I believe that is the idea behind it being a set yes.

qiime2/plugin/plugin.py
Oddant1: They are being added to self.types, for each fragment in the type_expression we construct a `SemanticTypeRecord` of the full type. . . Or something like that. I'm a bit fuzzy on it honestly, but that seems to be what's happening and is definitely the intent.

qiime2/plugin/plugin.py
thermokarst: Formats don't factor in here. I believe we suggested a set here to indicate that there is no order to the collection, we are less concerned with uniqueness.

qiime2/plugin/plugin.py
ChrisKeefe: I must be misunderstanding something here. If each `type_` we iterate over is a fragment, won't passing `type_` to `SemanticTypeRecord()` create records of type _fragments_, which we are then putting in `self.types`?

Oddant1: That's the part that I don't really understand

Oddant1: Correction, now that I actually look at where this actually is, `TypeFormatRecord` associates a full type with its canonical format. So `type_` is a full type expression at this point.

ChrisKeefe: Sorry if I'm completely missing the boat here. Is it possible for `record.type_expression` to hold more than one "full type expression"? If not, why are we iterating over it?

thermokarst:
> I don't love that PluginManager now has types and semantic_types attributes. Is there a better way to differentiate the two collections?

I agree - I would rather see a mirroring of what has been done in `Plugin` - `types` and `type_fragments`.

qiime2/plugin/plugin.py
thermokarst: > Is it possible for `record.type_expression` to hold more than one "full type expression"?

Yes. For example, this type expression can be [found in the wild](https://github.com/qiime2/q2-types/blob/22aa622f311881459373b9286b6adf526d7ef5df/q2_types/feature_table/_type.py#L38-L43): `FeatureTable[Frequency | RelativeFrequency | PresenceAbsence | Balance | Composition | PercentileNormalized]`

A MWE:

```python
import qiime2
from q2_types.feature_table import FeatureTable, Frequency, RelativeFrequency, PresenceAbsence

exp = FeatureTable[Frequency | RelativeFrequency | PresenceAbsence]
for type_ in exp:
    print(type_)
```

```
FeatureTable[Frequency]
FeatureTable[RelativeFrequency]
FeatureTable[PresenceAbsence]
```

Also, I just want to point out, @Oddant1 didn't add this loop, he moved it from the `PluginManager` (not that it really matters, I just want to clarify that this code didn't appear out of thin air).

ChrisKeefe: Thanks for the example, @thermokarst!

Oddant1:
> Right now there are `types` and `semantic_type_fragments` properties on `PluginManager` - I would like to see this mirror what he has done in `Plugin` - `types` and `type_fragments`.

Mirror in what way?

thermokarst:
Rename these properties to _match_ those on `Plugin`: `PluginManager.types` & `PluginManager.type_fragments` instead of what you currently have `PluginManager.types` & `PluginManager.semantic_type_fragments`.

qiime2/plugin/plugin.py
```diff
@@ -173,8 +173,8 @@ def decorator(transformer):
             # Apply the decorator as we were applied with a single function
             return decorator(_fn)
 
-    def register_semantic_type_fragments(self, *semantic_type_fragments):
-        for type_fragment in semantic_type_fragments:
+    def register_type_fragments(self, *type_fragments):
```
thermokarst: I don't think I am ready to deal with renaming this public API, please revert the method name to `register_semantic_types`.

qiime2/plugin/plugin.py
```diff
@@ -173,7 +173,7 @@ def decorator(transformer):
             # Apply the decorator as we were applied with a single function
             return decorator(_fn)
 
-    def register_type_fragments(self, *type_fragments):
+    def register_semantic_type_fragments(self, *type_fragments):
```
thermokarst: Please carefully re-read my request from before: https://github.com/qiime2/qiime2/pull/497#discussion_r336539448

qiime2/plugin/tests/test_plugin.py
```diff
@@ -151,7 +151,7 @@ def test_pipelines(self):
     # TODO test registration of directory formats.
 
     def test_types(self):
```
thermokarst: Please rename:

```suggestion
    def test_type_fragments(self):
```

Also, please add tests for `plugin.types` (if you need some ideas on how to go about this please let me know - for now, I would suggest trying to implement a test similar to the one described on L153).

PR #499, opened 2019-10-14
BUG: fix broken decode in MetadataColumn[...]

ebolyen:
This should correct the issue encountered here: 
https://forum.qiime2.org/t/error-running-qiime2-studio-bad-request/11548

### DIFF
--- a/qiime2/core/type/primitive.py
+++ b/qiime2/core/type/primitive.py
@@ -419,12 +419,12 @@ class _MetadataColumn(_PrimitiveTemplateBase):
             raise TypeError("Unsupported type in field: %r"
                             % (field.get_name(),))
 
-    def decode(self, metadata):
+    def decode(self, value):
         # This interface should have already retrieved this object.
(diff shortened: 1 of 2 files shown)


qiime2/core/type/primitive.py
```diff
@@ -419,12 +419,12 @@ def validate_field(self, name, field):
             raise TypeError("Unsupported type in field: %r"
                             % (field.get_name(),))
 
-    def decode(self, metadata):
+    def decode(self, value):
         # This interface should have already retrieved this object.
-        if not self.is_element(metadata):
+        if not isinstance(value, metadata.MetadataColumn):
```
ebolyen: @thermokarst, realized I was overthinking it, this change gets 99% of the intention of the original implementation for basically no additional work.

Normal type-checking will handle the rest.

thermokarst: :sake: LGTM!

PR #504, opened 2019-11-07
Import wizard

thermokarst:


qiime2/plugin/plugin.py
```diff
@@ -21,6 +21,8 @@
     'TransformerRecord', ['transformer', 'plugin', 'citations'])
 SemanticTypeRecord = collections.namedtuple(
     'SemanticTypeRecord', ['semantic_type', 'plugin'])
+SemanticTypeFragmentRecord = collections.namedtuple(
+    'SemanticTypeRecord', ['fragment', 'plugin'])
```
ebolyen:
```suggestion
    'SemanticTypeFragmentRecord', ['fragment', 'plugin'])
```

qiime2/plugin/plugin.py
```diff
@@ -201,9 +205,14 @@ def register_semantic_type_to_format(self, semantic_type, artifact_format):
...
 
-        self.type_formats.append(TypeFormatRecord(
+        record = TypeFormatRecord(
             type_expression=semantic_type,
-            format=artifact_format, plugin=self))
+            format=artifact_format, plugin=self)
+
+        self.type_formats.append(record)
+        for type_ in record.type_expression:
+            self.types.add(SemanticTypeRecord(
```
ebolyen: I would match the older `self.types` structure here (using a dict) as it make life marginally easier if you don't care about the `SemanticTypeRecord` objects (e.g. `list(self.types)` vs `[r.semantic_type for r in self.types]`).

This is a good approach otherwise.

qiime2/plugin/tests/test_plugin.py
```diff
@@ -150,15 +152,27 @@ def test_pipelines(self):
...
+        types = self.plugin.types
+        # Get just the types out of the SemanticTypeRecord namedtuples
+        types = {type_.semantic_type for type_ in types}
+
+        exp = {IntSequence1, IntSequence2, FourInts, Mapping, Kennel[Dog],
+               Kennel[Cat], SingleInt}
+        self.assertLessEqual(exp, types)
+        self.assertNotIn(Cat, types)
+        self.assertNotIn(Dog, types)
+        self.assertNotIn(Kennel, types)
```
ebolyen: Love it!

qiime2/sdk/plugin_manager.py
```diff
@@ -47,11 +46,17 @@ def __new__(cls):
...
         self.transformers = collections.defaultdict(dict)
+        self._reverse_transformers = collections.defaultdict(dict)
         self.formats = {}
         self.views = {}
         self.type_formats = []
+        self._ff_to_sfdf = {}
+        self._importable = set()
+        self._exportable = set()
+        self._canonical_formats = set()
+        self.types = {}
```
ebolyen: Consider the order of these instance attributes, there's so many some comments would be nice. For example, public attributes could be at the top, with the private ones after a line break.

qiime2/sdk/plugin_manager.py
```diff
@@ -62,22 +67,39 @@ def _init(self):
...
+
+        for importable_format in self._importable:
+            if(isinstance(importable_format,
+                          qiime2.plugin.model.file_format._FileFormat)):
+                self._importable.add(self._ff_to_sfdf[importable_format])
+
+        for exportable_format in self._exportable:
+            if(isinstance(exportable_format,
+                          qiime2.plugin.model.file_format._FileFormat)):
+                self._exportable.add(self._ff_to_sfdf[exportable_format])
```
ebolyen: My understanding is the three loops here are currently under revision, so I won't comment too much on them right now.

qiime2/sdk/plugin_manager.py
```diff
@@ -62,22 +67,39 @@ def _init(self):
...
 
-            self.semantic_types[type_name] = type_record
+            self.type_fragments[type_name] = type_record
 
         for (input, output), transformer_record in plugin.transformers.items():
             if output in self.transformers[input]:
                 raise ValueError("Transformer from %r to %r already exists."
                                  % transformer_record)
             self.transformers[input][output] = transformer_record
+            self._reverse_transformers[output][input] = transformer_record
```
ebolyen: The idea is good, the name could be misleading. The transformer still runs in the same direction, it is the dictionary that is reversed. Maybe `_transformers_by_output` would help contrast with `transformers` (and maybe that should be `_transformers_by_input` in all actuality).

qiime2/sdk/plugin_manager.py
```diff
@@ -89,8 +111,13 @@ def _integrate_plugin(self, plugin):
             self.views[name] = record
 
         for name, record in plugin.formats.items():
-            # TODO: remove this when `sniff` is removed
             fmt = record.format
+
+            if issubclass(
+                    fmt, qiime2.plugin.model.SingleFileDirectoryFormatBase):
+                self._ff_to_sfdf[fmt.file.format] = fmt
```
ebolyen: :+1:

qiime2/sdk/plugin_manager.py
```diff
@@ -100,44 +127,52 @@ def _integrate_plugin(self, plugin):
 
             self.formats[name] = record
         self.type_formats.extend(plugin.type_formats)
+        self.types[plugin.name] = plugin.types
+
+        for type_format in plugin.type_formats:
+            self._canonical_formats.add(type_format.format)
+            if isinstance(type_format.format,
+                          qiime2.plugin.model.SingleFileDirectoryFormatBase):
+                self._canonical_formats.add(type_format.format.file.format)
```
ebolyen: I think this chunk is superfluous, as `self._canonical_formats` isn't very useful, but this is currently being revised, so just don't forget to come back to this.

qiime2/sdk/plugin_manager.py
```diff
@@ -100,44 +127,52 @@ def _integrate_plugin(self, plugin):
 
             self.formats[name] = record
         self.type_formats.extend(plugin.type_formats)
+        self.types[plugin.name] = plugin.types
```
ebolyen: I'm not sure if this structure has any real advantage over `self.plugins[<name>].types`, so I would drop it.

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -86,9 +89,43 @@ def test_importable_types(self):
...
+                                  exportable=True, canonical_format=False)
+
+        self.assertEqual(obs, exp)
+
+    def test_get_formats_canonical_formats(self):
+        obs = self.pm._canonical_formats
+        exp = self.pm.get_formats(include_all=False, importable=False,
+                                  exportable=False, canonical_format=True)
+
+        self.assertEqual(obs, exp)
```
ebolyen: These tests kind of "beg their own question", so I would drop them and think about how to test the implementation of *where* these results come from instead (i.e., why should we expect `self.pm._importable/_exportable/etc` to be correct). Also, those properties are currently being worked on, so this is all likely to change anyhow.

I think we've discussed this, so it's probably on someones list already, but just wanted to mention it again here.

PR #505, opened 2019-11-08
MAINT: Remove params from get_semantic_types

Oddant1:


qiime2/sdk/plugin_manager.py
```diff
@@ -135,12 +135,10 @@ def _integrate_plugin(self, plugin):
...
                 self._canonical_formats.add(type_format.format.file.format)
 
-    def get_semantic_types(self, *, plugin=None):
+    def get_semantic_types(self):
         types = set()
 
-        plugins = [plugin] if plugin else self.plugins
-
-        for plugin in plugins.values():
+        for plugin in self.plugins.values():
```
thermokarst: [L130](https://github.com/qiime2/qiime2/pull/505/files#diff-81b344e767b72470d737b8c2972a9ffeR130) has an attribute that does the same thing already defined - how about you either use the attribute here, or, keep this implementation as-is, but remove the attribute above. Make sense?

Oddant1: Thinking back to the meeting yesterday, should we instead remove that `types` dictionary attribute altogether? Since, as pointed out during the meeting yesterday, it's entirely redundant to accessing the plugin from the `plugins` attribute, then accessing its types from there?

Oddant1:
> LGTM! Don't forget to merge this updated branch into any of your existing feature branches.

@David-Rod

PR #507, opened 2019-11-15
Merge import wizard and new changes to get_formats method

David-Rod:


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/Users/davidrodriguez/miniconda3/envs/q2dev/bin/python"
```
thermokarst: Hey @David-Rod, even though you added this file to the gitignore, you will still have to manually remove it from this changeset. Let me know if you need a hand with that.

qiime2/plugins.py
```diff
@@ -98,5 +98,11 @@ def exec_module(self, module):
                 for key, value in actions.items():
                     setattr(module, key, value)
 
+    def inspect_formats(filter=None, semantic_type=None):
+        import qiime2.sdk
+        pm = qiime2.sdk.PluginManager()
+
+        return pm.get_formats(filter=None, semantic_type=None)
```
thermokarst: Should this be:

```suggestion
        return pm.get_formats(filter, semantic_type)
```

?

Otherwise it is impossible to set the `filter` or `semantic_type`.

qiime2/sdk/plugin_manager.py
```diff
@@ -11,6 +11,8 @@
 import pkg_resources
 import qiime2.core.type
 
+from enum import Flag, auto
```
thermokarst: Can you change this to

```suggestion
import enum
```

then, update the code below as appropritate

David-Rod: Can `Flag` and `auto` be removed? I thought those are necessary due to their use in the `FormatFilter` class that is created on line `147`?

thermokarst: No, they can't be removed. You can access them through the `enum` package like this: `enum.Flag`, `enum.auto`.

qiime2/plugins.py
```diff
@@ -98,5 +98,11 @@ def exec_module(self, module):
                 for key, value in actions.items():
                     setattr(module, key, value)
 
+    def inspect_formats(filter=None, semantic_type=None):
```
thermokarst: This utility function should be moved into the "root" of the module, so that Artifact API users can import like this:

```python
from qiime2.plugins import inspect_formats
```

thermokarst: Second note, once this function is moved, you will need to add a corresponding entry to the `__all__` list of public objects (up around line 12).

thermokarst: Users will need some way of specifying the enum values for the `filter` flag --- they should be able to import those variables from `qiime2.plugins`, I think.

qiime2/sdk/plugin_manager.py
```diff
@@ -142,34 +144,75 @@ def get_semantic_types(self):
 
         return types
 
+    class FormatFilters(enum.Flag):
```
thermokarst: These enum values will need to be accessible for SDK developers to import and actually _use_ them (see my comments above about the Artifact API, similar story).

thermokarst: Maybe the name of the this class should be `GetFormatFilters`, or something along those lines. Any thoughts?

qiime2/sdk/plugin_manager.py
```diff
@@ -142,34 +144,75 @@ def get_semantic_types(self):
...
-                    exportable=False, canonical_format=False):
-        if include_all is True and canonical_format is True or include_all \
-                is True and importable is True or include_all is True and \
-                exportable is True:
-            raise ValueError("If all formats are requested, other formats "
-                             "cannot be included as a result.")
-
-        elif include_all is True:
-            return self.formats
+    def get_formats(self, *, filter=None, semantic_type=None):
```
thermokarst: This method will need a [numpydoc-compatible](https://numpydoc.readthedocs.io/en/latest/format.html) docstring.

qiime2/sdk/plugin_manager.py
```diff
@@ -142,34 +144,75 @@ def get_semantic_types(self):
...
 
         result_formats = set()
 
-        if importable is True:
-            result_formats = result_formats.union(self._importable)
+        if semantic_type is None:
```
thermokarst: I think we should have a clearly documented "order of operations" for the filtering operations. My gut instinct is that the `semantic_type` filter should be applied _first_, which means that the logic could be cleaned up a bit. I think you can take a strategy of progressively filtering. Start with the semantic type param, and filter the list of formats as necessary. Then, taking that resulting (semantic-type-filtered) list of formats, apply any of the filter param filters, as appropriate. By the end you wind up up one return. If you want a more detailed example just let me know!

qiime2/sdk/plugin_manager.py
```diff
@@ -142,34 +144,75 @@ def get_semantic_types(self):
...
-        if canonical_format is True:
-            result_formats = result_formats.union(self._canonical_formats)
+                    if self.FormatFilters.EXPORTABLE in filter:
+
+                        result_formats = result_formats.union(
+                            self.get_format_from(type_format.format))
 
         return result_formats
 
+    def get_format_from(self, canonical_format):
```
thermokarst: This method's name doesn't quite tell me what this does, can you think of a more descriptive name? Perhaps something that indicates that transformers are involved (like `get_formats_transformable_from`).

thermokarst: This method will need a [numpydoc-compatible](https://numpydoc.readthedocs.io/en/latest/format.html) docstring.

qiime2/sdk/plugin_manager.py
```diff
@@ -142,34 +144,75 @@ def get_semantic_types(self):
...
+
+        formats = set()
+
+        for format in self._reverse_transformers[canonical_format]:
+            formats.add(format)
+        formats.add(canonical_format)
+
+        return formats
+
+    def get_format_to(self, canonical_format):
```
thermokarst: See my comment above about the descriptive names.

thermokarst: This method will need a [numpydoc-compatible](https://numpydoc.readthedocs.io/en/latest/format.html) docstring.

qiime2/sdk/plugin_manager.py
```diff
@@ -142,34 +144,75 @@ def get_semantic_types(self):
...
+                    if self.FormatFilters.EXPORTABLE in filter:
+
+                        result_formats = result_formats.union(
+                            self.get_format_from(type_format.format))
 
         return result_formats
 
+    def get_format_from(self, canonical_format):
+
+        formats = set()
```
thermokarst: Stylistic, but I just wanted to point out that a set comprehension could work here, and is generally seen as a more "pythonic" or idiomatic approach (but again, that is really a minor "style" point).

qiime2/sdk/plugin_manager.py
thermokarst: This should actually be a private method, a docstring is optional.

qiime2/sdk/plugin_manager.py
thermokarst: This should actually be a private method, a docstring is optional.

PR #458, opened 2019-05-22
WIP: Usage API

thermokarst:

PR #511, opened 2019-12-12
IMP: plugin manager improved API, internal cleanup

thermokarst:
Splitting this out into it's own PR. Still needs tests. This PR is ported from https://github.com/qiime2/qiime2/pull/458.

### DIFF
--- a/qiime2/core/archive/provenance.py
+++ b/qiime2/core/archive/provenance.py
@@ -11,7 +11,6 @@ import collections
 import pkg_resources
 import uuid
 import copy
-import importlib
 import shutil
 import sys
 from datetime import datetime
@@ -348,9 +347,11 @@ class ImportProvenanceCapture(ProvenanceCapture):
 
 
 class ActionProvenanceCapture(ProvenanceCapture):
-    def __init__(self, action_type, import_path, action_id):
+    def __init__(self, action_type, plugin_id, action_id):
+        from qiime2.sdk import PluginManager
+
         super().__init__()
-        self._plugin = importlib.import_module(import_path).__plugin__
+        self._plugin = PluginManager().get_plugin(id=plugin_id)
         self.action = self._plugin.actions[action_id]
         self.action_type = action_type
         self.inputs = OrderedKeyValue()
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -38,29 +38,49 @@ class PluginManager:
 
     # This class is a singleton as it is slow to create, represents the
     # state of a qiime2 installation, and is needed *everywhere*
-    def __new__(cls):
+    def __new__(cls, install_plugins=True):
         if cls.__instance is None:
             self = super().__new__(cls)
-            self._init()
+            self._init(install_plugins=install_plugins)
             cls.__instance = self
         return cls.__instance
 
-    def _init(self):
+    def _init(self, install_plugins):
         self.plugins = {}
+        self._plugin_by_id = {}
         self.semantic_types = {}
         self.transformers = collections.defaultdict(dict)
         self.formats = {}
         self.views = {}
         self.type_formats = []
 
-        # These are all dependent loops, each requires the loop above it to
-        # be completed.
-        for entry_point in self.iter_entry_points():
-            plugin = entry_point.load()
-            self.plugins[plugin.name] = plugin
+        if install_plugins:
+            # These are all dependent loops, each requires the loop above it to
+            # be completed.
+            for entry_point in self.iter_entry_points():
+                project_name = entry_point.dist.project_name
+                package = entry_point.module_name.split('.')[0]
+                plugin = entry_point.load()
 
-        for plugin in self.plugins.values():
-            self._integrate_plugin(plugin)
+                self.install_plugin(plugin, package, project_name)
+
+    def install_plugin(self, plugin, package=None, project_name=None):
+        self.plugins[plugin.name] = plugin
+        self._plugin_by_id[plugin.id] = plugin
+        if plugin.package is None:
+            plugin.package = package
+        if plugin.project_name is None:
+            plugin.project_name = project_name
+        self._integrate_plugin(plugin)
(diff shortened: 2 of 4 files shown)


qiime2/sdk/plugin_manager.py
```diff
@@ -38,29 +38,57 @@ def iter_entry_points(cls):
...
+            # These are all dependent loops, each requires the loop above it to
+            # be completed.
+            for entry_point in self.iter_entry_points():
+                project_name = entry_point.dist.project_name
+                package = entry_point.module_name.split('.')[0]
+                plugin = entry_point.load()
+
+                self.install_plugin(plugin, package, project_name)
+
+    def install_plugin(self, plugin, package=None, project_name=None):
```
ebolyen: I wonder if there's another word we can use. Since installing things is usually such a challenge they are probably not going to think of this method in the context of adding it to an instance.

What about `add_plugin` and `add_plugins=True`?

thermokarst: Good point. I like `add` vs `install`.

qiime2/sdk/plugin_manager.py
```diff
@@ -38,29 +38,57 @@ def iter_entry_points(cls):
 
     # This class is a singleton as it is slow to create, represents the
     # state of a qiime2 installation, and is needed *everywhere*
-    def __new__(cls):
+    def __new__(cls, add_plugins=True):
```
ebolyen: we should add an `else` branch to complain about `add_plugins=False` if the singleton already exists.

qiime2/sdk/plugin_manager.py
```diff
@@ -38,29 +38,57 @@ def iter_entry_points(cls):
...
+        self.plugins[plugin.name] = plugin
+        self._plugin_by_id[plugin.id] = plugin
+        if plugin.package is None:
+            plugin.package = package
+        if plugin.project_name is None:
+            plugin.project_name = project_name
+        self._integrate_plugin(plugin)
+        plugin.freeze()
+
+    def get_plugin(self, id=None, name=None):
```
ebolyen: let's keyword-only these

qiime2/core/archive/provenance.py
```diff
@@ -348,9 +347,11 @@ def make_action_section(self):
 
 
 class ActionProvenanceCapture(ProvenanceCapture):
-    def __init__(self, action_type, import_path, action_id):
+    def __init__(self, action_type, plugin_id, action_id):
+        from qiime2.sdk import PluginManager
+
         super().__init__()
-        self._plugin = importlib.import_module(import_path).__plugin__
+        self._plugin = PluginManager().get_plugin(id=plugin_id)
```
ebolyen: this is so much less cludgy it's unreal

qiime2/sdk/plugin_manager.py
```diff
@@ -38,29 +38,57 @@ def iter_entry_points(cls):
...
+
+                self.add_plugin(plugin, package, project_name)
+
+    def add_plugin(self, plugin, package=None, project_name=None):
+        self.plugins[plugin.name] = plugin
+        self._plugin_by_id[plugin.id] = plugin
+        if plugin.package is None:
+            plugin.package = package
+        if plugin.project_name is None:
+            plugin.project_name = project_name
```
ebolyen: we need to check if these are still `None` after applying arguments, if so, we'll want to error and force the caller to pass them.

qiime2/core/archive/provenance.py
ebolyen: ok, so the thing I want to test is if `concurrent.futures` does something particular odd about the memory mapping for Python modules (which would cause the plugin to be the same address) vs the new approach, which may create a new PluginManager, with potentially new state (such that the plugin that was added, is not available in the "copy" as it was never added in the clone).

qiime2/sdk/plugin_manager.py
ebolyen: philosophical question: do we worry about mutating the object we import, that is to say, do we ever want to share the same plugin object between plugin-managers? It sounds like nonsense, but we are ruling that future out by doing this.

qiime2/core/archive/provenance.py
ebolyen: ok, empirically, this works fine, I created a PluginManager and added only the `dummy_plugin`, then did an async call on `parameter_only_pipelines` and got all of the provenance I expected, and the dummy_plugin was identified (which would be missing if we had a different PluginManager).

```pytb
In [1]: from qiime2.sdk import PluginManager                                                                                                  

In [2]: pm = PluginManager(add_plugins=False)                                                                                                 

In [3]: pm.plugins                                                                                                                            
Out[3]: {}

In [4]: from qiime2.core.testing.plugin import dummy_plugin                                                                                   

In [5]: pm.add_plugin(dummy_plugin)                                                                                                           

In [6]: pm.plugins                                                                                                                            
Out[6]: {'dummy-plugin': <qiime2.plugin.plugin.Plugin at 0x7fd29b753da0>}

In [7]: p = dummy_plugin.actions['parameter_only_pipeline']                                                                                   

In [8]: fut = p.asynchronous(1, 2)                                                                                                            

In [9]: r = fut.result()                                                                                                                      

In [10]: r                                                                                                                                    
Out[10]: 
Results (name = value)
-------------------------------------------------------------------------
foo = <artifact: IntSequence2 uuid: 4338de5f-ae26-44b5-b883-e8567d0168ef>
bar = <artifact: IntSequence1 uuid: 32c1b8c8-d385-4e5e-ae98-f2432d2d075e>
```

qiime2/sdk/plugin_manager.py
thermokarst: I might not have understood your request, check out https://github.com/qiime2/qiime2/pull/511/files#diff-81b344e767b72470d737b8c2972a9ffeR46-R50

ebolyen: That looks right (the final diff at least, the link is being weird for me)!

Issue #510, opened 2019-12-12
Gracefully fall back on tzlocal failures


thermokarst:
**Improvement Description**
Currently, if a deployment environment isn't [configured correctly](https://github.com/regebro/tzlocal/issues/79), tzlocal will fail with a message like:

```python
ValueError: Timezone offset does not match system offset: 0 != -25200. Please, check your config files.
```

The timezone information is potentially useful for decentralized provenance (at least in theory), but not strictly necessary, so perhaps a solution is to fall back to UTC if tzlocal isn't able to initialize properly.

**Alternatives**
1. Timezone info could be scrapped.
2. Identify an alternative to tzlocal

**References**
1. https://forum.qiime2.org/t/error-while-running-qiime2-valueerror-timezone-offset-does-not-match-system-offset-0-25200-please-check-your-config-files/11418
2. https://github.com/qiime2/qiime2/issues/493
3. https://github.com/nf-core/ampliseq/issues/114

thermokarst:
Turns out tzlocal already does fall back to UTC, but, only when _no timezone info is present_ on the system. Otherwise, the error reported above is raised specifically when the host config is in an inconsistent state.

ebolyen:
It should be really easy to just back out of the tzlocal logic if we see that ValueError. I kind of wish they had a custom baseclass for that exception though.

So I think that would be my vote for now, use it if it is consistent, otherwise ignore it entirely.

thermokarst:
Closed in #513.

rsettlage:
Hi,
I have a few users encountering this issue.  Is there a timeline for pushing 2020.2?  Or should I ask them to install from the git repo?
Thanks,
Bob

thermokarst:
Hi @rsettlage!

>  Is there a timeline for pushing 2020.2?

2020.2 release plans can be found here:

- https://forum.qiime2.org/t/preview-qiime-2-2020-2-development-changelog/12328/
- https://github.com/orgs/qiime2/projects/18

> I have a few users encountering this issue.

One immediate solution is that they can attempt to resolve the misconfigured environment. tzlocal is raising this warning because it has detected _multiple conflicting_ timezone configurations.

PR #513, opened 2019-12-19
Fall back for tzlocal failure

David-Rod:
This PR addresses the issues that some users encounter regarding timezone information. This fix bypasses the `ValueError` that is shown and simply sets the timezone to none, which produces date & time, without any timezone information.

See issue #510 for more details.

### DIFF
--- a/qiime2/core/archive/tests/test_provenance.py
+++ b/qiime2/core/archive/tests/test_provenance.py
@@ -8,6 +8,7 @@
 
 import unittest
 import re
+import unittest.mock as mock
 
 import pandas as pd
 import pandas.util.testing as pdt
@@ -15,6 +16,7 @@ import pandas.util.testing as pdt
 import qiime2
 from qiime2.plugins import dummy_plugin
 from qiime2.core.testing.type import IntSequence1, Mapping
+import qiime2.core.archive.provenance as provenance
 
 
 class TestProvenanceIntegration(unittest.TestCase):
@@ -248,6 +250,17 @@ class TestProvenanceIntegration(unittest.TestCase):
 
         self.assertIn('foo: 3', prov_yml)
         self.assertIn('bar: 2', prov_yml)
+
+    @mock.patch('qiime2.core.archive.provenance.tzlocal.get_localzone',
+                side_effect=ValueError())
+    def test_ts_to_date(self, mocked_link):
+        q2_paper_date = 1563984000
+
+        obs = str(provenance._ts_to_date(q2_paper_date))
+        exp = "2019-07-24 09:00:00"
+
+        self.assertEqual(obs, exp)
+        assert mocked_link.called
 
 
 if __name__ == '__main__':
--- a/qiime2/core/archive/tests/test_provenance.py
+++ b/qiime2/core/archive/tests/test_provenance.py
@@ -253,14 +253,14 @@ class TestProvenanceIntegration(unittest.TestCase):
 
     @mock.patch('qiime2.core.archive.provenance.tzlocal.get_localzone',
                 side_effect=ValueError())
-    def test_ts_to_date(self, mocked_link):
+    def test_ts_to_date(self, mocked_tzlocal):
         q2_paper_date = 1563984000
 
         obs = str(provenance._ts_to_date(q2_paper_date))
         exp = "2019-07-24 16:00:00+00:00"
 
         self.assertEqual(obs, exp)
-        assert mocked_link.called
+        self.assertTrue(mocked_tzlocal.called)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


David-Rod:
@thermokarst Yeah I will need to take a closer look at this one. All tests passed locally on my machine, but there seems to be a time difference of 7hrs for the test. That difference is the same as the difference between MST and UTC time. It might have to do with `tzlocal`. Thanks.

David-Rod:
@thermokarst and @ebolyen,

I would like to make a commit to test a change that I made to the code for this PR, but I received an error regarding an `Invalid header` when I ran `make lint`(Please see snippet below).

___
image
___

So, I replaced all 2019 references in the header of each affected file to 2020. The error went away after I made these changes. My concern is that the diff for the changes will include a lot of files for you guys to sift through. In the past, I know that header changes were made in individual PR's for each directory. Any suggestions for how to proceed?

thermokarst:
> Any suggestions on how I should proceed?

Please open a separate PR with the q2lint-related changes. FWIW, we will need to make similar changes across _all_ QIIME 2 repos, so if you're looking for something to do, that would be great. Thanks!

thermokarst:
Here is last year's q2lint PR for this repo, BTW: https://github.com/qiime2/qiime2/pull/426

David-Rod:
Thanks for the link to last year's changes. It looks like there are 98 changes from last year vs. 111 changes for my new PR. I will review to ensure that there were no unnecessary changes to the repo.

thermokarst:
You will need to update this branch now that #515 has been merged.

qiime2/core/archive/tests/test_provenance.py
```diff
@@ -249,6 +251,17 @@ def test_unioned_primitives(self):
...
+    @mock.patch('qiime2.core.archive.provenance.tzlocal.get_localzone',
+                side_effect=ValueError())
+    def test_ts_to_date(self, mocked_link):
+        q2_paper_date = 1563984000
+
+        obs = str(provenance._ts_to_date(q2_paper_date))
+        exp = "2019-07-24 09:00:00"
+
+        self.assertEqual(obs, exp)
+        assert mocked_link.called
```
thermokarst: Regarding the test failure on travis - I think you should move this assertion up to _before_ checking that obs == exp, that way you can verify that the patch is being applied correctly. Put another way, we should make sure that the code path [here](https://github.com/qiime2/qiime2/pull/513/files#diff-e183f9c47f8eeab74f2219b35f12921cR34) is actually being exercised in this test.

thermokarst: Okay scratch that, I think I know what is wrong. Check out the docs for [`datetime.fromtimestamp`](https://docs.python.org/3.6/library/datetime.html#datetime.datetime.fromtimestamp), in particular, this part:

> If optional argument tz is None or not specified, the timestamp is converted to the platform’s local date and time, and the returned datetime object is naive.

So here is what I think is happening: when [this bit](https://github.com/qiime2/qiime2/pull/513/files#diff-e183f9c47f8eeab74f2219b35f12921cR35) of code is being executed, the `None` value signals to `datetime` to check for the system's timezone info. Since the travis env apparently _does_ have timezone info, the date/time is altered to match the system timezone which isn't really what we want.

I think a more robust solution would be to explicitly set the fallback timezone to 'UTC', rather than `None`, I will leave that exercise (and research) up to you. Let me know if you need a hand!

qiime2/core/archive/tests/test_provenance.py
```diff
@@ -249,6 +251,17 @@ def test_unioned_primitives(self):
         self.assertIn('foo: 3', prov_yml)
         self.assertIn('bar: 2', prov_yml)
 
+    @mock.patch('qiime2.core.archive.provenance.tzlocal.get_localzone',
+                side_effect=ValueError())
+    def test_ts_to_date(self, mocked_link):
```
thermokarst:
```suggestion
    def test_ts_to_date(self, mocked_tzlocal):
```

qiime2/core/archive/tests/test_provenance.py
```diff
@@ -249,6 +251,17 @@ def test_unioned_primitives(self):
...
+        q2_paper_date = 1563984000
+
+        obs = str(provenance._ts_to_date(q2_paper_date))
+        exp = "2019-07-24 16:00:00+00:00"
+
+        self.assertEqual(obs, exp)
+        assert mocked_link.called
```
thermokarst:
```suggestion
        self.assertTrue(mocked_tzlocal.called)
```

thermokarst:
Ah bummer:

> You do not have permission to push to this repository.

PR #522, opened 2020-02-10
get_formats on PluginManager

ebolyen:
cc @David-Rod @Oddant1


qiime2/sdk/plugin_manager.py
```diff
@@ -9,8 +9,17 @@
 import collections
 import os
 import pkg_resources
-import qiime2.core.transform as transform
 import qiime2.core.type
+from qiime2.core.format import FormatBase
+from qiime2.plugin.model import SingleFileDirectoryFormatBase
+from qiime2.sdk.util import parse_type
+
+import enum
```
ebolyen: this one goes to the top (with the other stdlib imports)

qiime2/sdk/plugin_manager.py
```diff
@@ -101,43 +122,112 @@ def _integrate_plugin(self, plugin):
...
+                             "valid.", (filter))
+
+        if semantic_type is None:
+            formats = set(f.format for f in self.type_formats)
+
+        else:
+            formats = set()
+
+            if isinstance(semantic_type, str):
+                semantic_type = parse_type(semantic_type, "semantic")
```
ebolyen: you may want to verify that non-strings are in fact semantic types, otherwise we'll get a weird error with `not_type <= type_format.type_expression` downstream.

qiime2/sdk/plugin_manager.py
```diff
@@ -101,43 +122,112 @@ def _integrate_plugin(self, plugin):
...
 
-        A format is importable in a QIIME 2 deployment if it can be transformed
-        into at least one of the canonical semantic type formats.
+        semantic_type : TypeExpression | String
+            The semantic type is used to filter the formats associated with
+            that specific semantic type
 
+        This method will filter out the formats using the filter provided by
+        the user and the semantic type. The return is a set of filtered
+        formats.
```
ebolyen: update this to indicate dictionary

qiime2/sdk/plugin_manager.py
```diff
@@ -101,43 +122,112 @@ def _integrate_plugin(self, plugin):
             self.formats[name] = record
         self.type_formats.extend(plugin.type_formats)
 
+    def get_semantic_types(self):
+        types = set()
```
ebolyen: I would make this a dictionary to match `get_formats`

qiime2/sdk/plugin_manager.py
```diff
@@ -101,43 +122,112 @@ def _integrate_plugin(self, plugin):
...
-        importable_types = set()
-        for type_format in self.type_formats:
-            for type in type_format.type_expression:
-                importable_types.add(type)
-        return importable_types
+        _get_formats_helper(self, formats, transformer_dict)
+
+        formats : set(DirectoryFormat)
+            We are finding all formats that can be directly transformed to and
+            from formats in this set
```
ebolyen: drop language about to/from since that is contextual

qiime2/sdk/plugin_manager.py
```diff
@@ -101,43 +122,112 @@ def _integrate_plugin(self, plugin):
...
+    def _get_formats_helper(self, formats, transformer_dict):
         """
-        importable_types = set()
-        for type_format in self.type_formats:
-            for type in type_format.type_expression:
-                importable_types.add(type)
-        return importable_types
+        _get_formats_helper(self, formats, transformer_dict)
+
+        formats : set(DirectoryFormat)
```
ebolyen: `Set[DirectoryFormat]`

qiime2/sdk/plugin_manager.py
```diff
@@ -101,43 +122,112 @@ def _integrate_plugin(self, plugin):
...
+            We are finding all formats that can be directly transformed to and
+            from formats in this set
+
+        tranformer_dict : Class: Dict{ Class: TransformerRecord }
```
ebolyen: `Dict[str, Dict[str, TransformerRecord]]` (I think it's a str anyhow, update if not)

qiime2/sdk/plugin_manager.py
```diff
@@ -101,43 +122,112 @@ def _integrate_plugin(self, plugin):
...
+                if issubclass(transformed_format, FormatBase):
+                    result_formats.add(transformed_format)
+
+                    if issubclass(transformed_format,
+                                  SingleFileDirectoryFormatBase):
+                        result_formats.add(transformed_format.file.format)
+
+                    if transformed_format in self._ff_to_sfdf:
+                        result_formats.update(
+                            self._ff_to_sfdf[transformed_format])
```
ebolyen: excellent!

qiime2/sdk/tests/test_plugin_manager.py
```diff
@@ -10,20 +10,25 @@
 
 import qiime2.plugin
 import qiime2.sdk
+
```
ebolyen: drop line

qiime2/sdk/plugin_manager.py
```diff
@@ -101,43 +122,112 @@ def _integrate_plugin(self, plugin):
...
+                types.add(type_record)
+
+        return types
+
     # TODO: Should plugin loading be transactional? i.e. if there's
     # something wrong, the entire plugin fails to load any piece, like a
     # databases rollback/commit
 
-    @property
-    def importable_formats(self):
```
ebolyen: This will break q2cli, we can either block this PR until it's updated, or we can shim this property to use get_formats

qiime2/sdk/plugin_manager.py
```diff
@@ -101,43 +122,112 @@ def _integrate_plugin(self, plugin):
...
-                    type_format.format)
-                if from_type.has_transformation(to_type):
-                    importable_formats[name] = record
+                if semantic_type <= type_format.type_expression:
+                    formats.add(type_format.format)
                     break
-        return importable_formats
 
-    @property
-    def importable_types(self):
```
ebolyen: same deal here re: q2cli

ebolyen:
conflicts are from #511

qiime2/core/testing/format.py
```diff
@@ -38,7 +38,7 @@ def _validate_(self, level):
         self._validate_n_ints(record_map[level])
 
 
-class IntSequenceFormatV2(IntSequenceFormat):
```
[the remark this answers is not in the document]
Oddant1: I think I can just add the inheritance back, I honestly can't remember why I changed that in the first place. Whatever the reason was, changing it back to inheriting from `IntSequenceFormat` doesn't break the tests at all.

qiime2/core/testing/transformer.py
```diff
@@ -94,6 +97,18 @@ def _1000(ff: IntSequenceFormat) -> IntSequenceFormatV2:
     return new_ff
 
 
+@dummy_plugin.register_transformer
+def _1100(df: IntSequenceMultiFileDirectoryFormat) -> \
+        IntSequenceDirectoryFormat:
+    return IntSequenceDirectoryFormat()
```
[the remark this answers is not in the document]
Oddant1: The `IntSequenceMultiFileDirectoryFormat` doesn't actually contain any data :laughing:. The entire body of the class is just `pass`. Do we want it to have data in it? I just went with the path of least resistance when creating it because data wasn't needed to test this behavior.

qiime2/sdk/plugin_manager.py
Oddant1: Hmmm, I think it makes the most sense to basically just add the old functionality back here? Since the `get_semantic_types` method doesn't filter only importable types.

qiime2/sdk/plugin_manager.py
Oddant1: Now that this has been largely signed off on, I think it's time to implement the cli interface for this new functionality like we talked about. If you want, I can address the rest of this review then open a cli PR allowing for proper interfacing with the new `get_formats` which would allow for viewing all importable formats among other things obviously. We can also temporarily reinstate the old interface so this can be merged immediately, then that functionality can be removed once the new interface is ready.

qiime2/sdk/plugin_manager.py
ebolyen: There's no real difference, "importable_types" in this context just means not-a-type-fragment, which is exactly what `get_semantic_types` does. We just had worse language back then.

qiime2/sdk/plugin_manager.py
ebolyen: Since we are up against a release, I would rather just shim these properties, then we can address q2cli at a later time (whenever we want to).

qiime2/core/testing/transformer.py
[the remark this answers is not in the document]
ebolyen: fair enough, let's ignore then, but add a comment indicating this is for testing `get_formats` only

qiime2/plugin/plugin.py
```diff
@@ -67,9 +69,10 @@ def __init__(self, name, version, website, package, citation_text=None,
 
         self.formats = {}
         self.views = {}
-        self.types = {}
+        self.type_fragments = {}
         self.transformers = {}
         self.type_formats = []
+        self.types = set()
```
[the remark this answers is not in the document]
Oddant1: So this becomes an `@property` producing a dictionary of `str(semantic_type)` to `SemanticTypeRecord`?  Because that's what `get_semantic_types` does, so should we just make that an `@property`?

ebolyen:
I think this is ready! Thank you so much @David-Rod and @Oddant1 for your work on this!

PR #523, opened 2020-02-10
NEW: Usage API

thermokarst:
Replaces #458.

### DIFF
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -113,7 +113,7 @@ def typical_pipeline_complex(use):
     right2 = use.get_result('right2')
     right2.assert_has_line_matching(
         label='a nice label about this assertion',
-        path='.*/data/ints.txt',
+        path='ints.txt',
         expression='1',
     )
 
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -8,7 +8,7 @@
 
 import pandas as pd
 
-from qiime2 import Artifact, Metadata
+from qiime2 import Artifact, Metadata, CategoricalMetadataColumn
 
 from qiime2.plugin import UsageAction, UsageInputs, UsageOutputNames
 
@@ -41,6 +41,13 @@ def md2_factory():
     return Metadata(pd.DataFrame({'b': ['4', '5', '6']},
                                  index=pd.Index(['0', '1', '2'],
                                                 name='id')))
+
+
+def mdc1_factory():
+    return CategoricalMetadataColumn(pd.Series(['1', '2', '3'],
+                                     name='a',
+                                     index=pd.Index(['0', '1', '2'],
+                                                    name='id')))
 
 
 def concatenate_ints_simple(use):
@@ -146,5 +153,31 @@ def identity_with_metadata_merging(use):
         UsageAction(plugin_id='dummy_plugin',
                     action_id='identity_with_metadata'),
         UsageInputs(ints=ints, metadata=md3),
+        UsageOutputNames(out='out'),
+    )
+
+
+def identity_with_metadata_column_get_mdc(use):
+    ints = use.init_data('ints', ints1_factory)
+    md = use.init_data('md', md1_factory)
+
+    mdc = use.get_metadata_column('mdc', md, 'a')
+
+    use.action(
+        UsageAction(plugin_id='dummy_plugin',
+                    action_id='identity_with_metadata_column'),
+        UsageInputs(ints=ints, metadata=mdc),
+        UsageOutputNames(out='out'),
+    )
+
+
+def identity_with_metadata_column_from_factory(use):
+    ints = use.init_data('ints', ints1_factory)
+    mdc = use.init_data('mdc', mdc1_factory)
+
+    use.action(
+        UsageAction(plugin_id='dummy_plugin',
+                    action_id='identity_with_metadata_column'),
+        UsageInputs(ints=ints, metadata=mdc),
         UsageOutputNames(out='out'),
     )
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -6,7 +6,9 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
-from qiime2 import Artifact
+import pandas as pd
+
+from qiime2 import Artifact, Metadata
 
 from qiime2.plugin import UsageAction, UsageInputs, UsageOutputNames
 
@@ -27,6 +29,18 @@ def ints3_factory():
 
 def mapping1_factory():
     return Artifact.import_data(Mapping, {'a': 42})
+
+
+def md1_factory():
+    return Metadata(pd.DataFrame({'a': ['1', '2', '3']},
+                                 index=pd.Index(['0', '1', '2'],
+                                                name='id')))
+
+
+def md2_factory():
+    return Metadata(pd.DataFrame({'b': ['4', '5', '6']},
+                                 index=pd.Index(['0', '1', '2'],
+                                                name='id')))
 
 
 def concatenate_ints_simple(use):
@@ -107,3 +121,30 @@ def typical_pipeline_complex(use):
 def comments_only(use):
     use.comment('comment 1')
     use.comment('comment 2')
+
+
+def identity_with_metadata_simple(use):
+    ints = use.init_data('ints', ints1_factory)
+    md = use.init_data('md', md1_factory)
+
+    use.action(
+        UsageAction(plugin_id='dummy_plugin',
+                    action_id='identity_with_metadata'),
+        UsageInputs(ints=ints, metadata=md),
+        UsageOutputNames(out='out'),
+    )
+
+
+def identity_with_metadata_merging(use):
+    ints = use.init_data('ints', ints1_factory)
+    md1 = use.init_data('md1', md1_factory)
+    md2 = use.init_data('md2', md2_factory)
+
+    md3 = use.merge_metadata('md3', md1, md2)
+
+    use.action(
+        UsageAction(plugin_id='dummy_plugin',
+                    action_id='identity_with_metadata'),
+        UsageInputs(ints=ints, metadata=md3),
+        UsageOutputNames(out='out'),
+    )
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -36,7 +36,7 @@ def concatenate_ints_simple(use):
 
     use.comment('This example demonstrates basic usage.')
     use.action(
-        UsageAction(plugin_id='dummy_plugin', action_name='concatenate_ints'),
+        UsageAction(plugin_id='dummy_plugin', action_id='concatenate_ints'),
         UsageInputs(ints1=ints_a, ints2=ints_b, ints3=ints_c, int1=4, int2=2),
         UsageOutputNames(concatenated_ints='ints_d'),
     )
@@ -49,7 +49,7 @@ def concatenate_ints_complex(use):
 
     use.comment('This example demonstrates chained usage (pt 1).')
     use.action(
-        UsageAction(plugin_id='dummy_plugin', action_name='concatenate_ints'),
+        UsageAction(plugin_id='dummy_plugin', action_id='concatenate_ints'),
         UsageInputs(ints1=ints_a, ints2=ints_b, ints3=ints_c, int1=4, int2=2),
         UsageOutputNames(concatenated_ints='ints_d'),
     )
@@ -57,7 +57,7 @@ def concatenate_ints_complex(use):
     ints_d = use.get_result('ints_d')
     use.comment('This example demonstrates chained usage (pt 2).')
     use.action(
-        UsageAction(plugin_id='dummy_plugin', action_name='concatenate_ints'),
+        UsageAction(plugin_id='dummy_plugin', action_id='concatenate_ints'),
         UsageInputs(ints1=ints_d, ints2=ints_b, ints3=ints_c, int1=41, int2=0),
         UsageOutputNames(concatenated_ints='concatenated_ints'),
     )
@@ -68,7 +68,7 @@ def typical_pipeline_simple(use):
     mapper = use.init_data('mapper', mapping1_factory)
 
     use.action(
-        UsageAction(plugin_id='dummy_plugin', action_name='typical_pipeline'),
+        UsageAction(plugin_id='dummy_plugin', action_id='typical_pipeline'),
         UsageInputs(int_sequence=ints, mapping=mapper, do_extra_thing=True),
         UsageOutputNames(out_map='out_map', left='left', right='right',
                          left_viz='left_viz', right_viz='right_viz')
@@ -80,7 +80,7 @@ def typical_pipeline_complex(use):
     mapper1 = use.init_data('mapper1', mapping1_factory)
 
     use.action(
-        UsageAction(plugin_id='dummy_plugin', action_name='typical_pipeline'),
+        UsageAction(plugin_id='dummy_plugin', action_id='typical_pipeline'),
         UsageInputs(int_sequence=ints1, mapping=mapper1, do_extra_thing=True),
         UsageOutputNames(out_map='out_map1', left='left1', right='right1',
                          left_viz='left_viz1', right_viz='right_viz1')
@@ -90,9 +90,8 @@ def typical_pipeline_complex(use):
     mapper2 = use.get_result('out_map1')
 
     use.action(
-        UsageAction(plugin_id='dummy_plugin', action_name='typical_pipeline'),
+        UsageAction(plugin_id='dummy_plugin', action_id='typical_pipeline'),
         UsageInputs(int_sequence=ints2, mapping=mapper2, do_extra_thing=False),
-        # TODO with and without overwriting
         UsageOutputNames(out_map='out_map2', left='left2', right='right2',
                          left_viz='left_viz2', right_viz='right_viz2')
     )
@@ -106,5 +105,5 @@ def typical_pipeline_complex(use):
 
 
 def comments_only(use):
-    # TODO
-    pass
+    use.comment('comment 1')
+    use.comment('comment 2')
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -1,0 +1,110 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016-2020, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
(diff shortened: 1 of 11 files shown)


qiime2/sdk/action.py
```diff
@@ -342,6 +344,8 @@ def _build_numpydoc(self):
         parameters = self._build_section("Parameters", sig.signature_order)
         returns = self._build_section("Returns", sig.outputs)
 
+        # TODO: include Usage-rendered examples here
```
ebolyen: It seems like you have all you need to make this happen, it would be pretty cool to add.

qiime2/sdk/tests/test_usage.py
```diff
@@ -0,0 +1,241 @@
...
+from qiime2.core.type import signature
+from qiime2.core.testing.util import get_dummy_plugin
+from qiime2.core.testing.type import Mapping
+from qiime2.sdk import usage, action
+from qiime2 import plugin
+
+
+class TestCaseUsage(unittest.TestCase):
+    def setUp(self):
+        # TODO standardize temporary directories created by QIIME 2
```
ebolyen: lol

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,349 @@
...
+# TODO: docstrings
+
+
+class UsageAction:
+    def __init__(self, plugin_id: str, action_name: str):
+        if plugin_id == '':
+            raise ValueError('Must specify a value for plugin_id.')
+
+        if action_name == '':
+            raise ValueError('Must specify a value for action_name.')
```
ebolyen: is this an `action_id`?

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,349 @@
...
+        self._plugin_manager = sdk.PluginManager()
+
+    def __repr__(self):
+        return 'UsageAction(plugin_id=%r, action_name=%r)' %\
+            (self.plugin_id, self.action_name)
+
+    def get_action(self):
+        plugin = self._plugin_manager.get_plugin(id=self.plugin_id)
+        # TODO: should this validation be pushed up into
+        # plugin.py or action.py?
```
ebolyen: That's a good question, we have overridden those classes, so it would be pretty easy to super `__getitem__` with a nicer KeyError. I would be in favor I think (as long as it stays `KeyError` we won't have broken backwards compat, I think).

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,349 @@
...
+
+        self.plugin_id = plugin_id
+        self.action_name = action_name
+        self._plugin_manager = sdk.PluginManager()
+
+    def __repr__(self):
+        return 'UsageAction(plugin_id=%r, action_name=%r)' %\
+            (self.plugin_id, self.action_name)
+
+    def get_action(self):
```
ebolyen: I like having a little API on these objects, this is very nice.

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,349 @@
...
+        # TODO: should this validation be pushed up into
+        # plugin.py or action.py?
+        try:
+            action_f = plugin.actions[self.action_name]
+        except KeyError:
+            raise KeyError('No action currently registered with '
+                           'name: "%s".' % (self.action_name,))
+        return (action_f, action_f.signature)
+
+    def validate(self, inputs, outputs):
```
ebolyen: This is great!

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,349 @@
...
+        missing = exp_params - provided
+        if len(missing) > 0:
+            raise ValueError('Missing parameter(s): %r' % (missing, ))
+
+        extra = provided - exp_inputs - exp_params
+        if len(extra) > 0:
+            raise ValueError('Extra input(s) or parameter(s): %r' %
+                             (extra, ))
+
+    def prepare(self, signature, scope):
```
ebolyen: would `bind` be a better word for this?

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,349 @@
...
+
+class Usage(metaclass=abc.ABCMeta):
+    def __init__(self):
+        self._scope = Scope()
+        self._reverse_lookup = {}
+
+    def init_data(self, ref, factory):
+        override = self._factory_override(ref, factory)
+        record = self._push_record(factory=override)
+        self._reverse_lookup[ref] = record.id
```
ebolyen: Should this be the job of `Scope`?

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,349 @@
...
+
+    def _action_(self, action: UsageAction,
+                 input_opts: dict, output_opts: list):
+        return output_opts
+
+    def _comment_(self, text: str):
+        return None
+
+    def _assert_has_line_matching_(self, scope_id, label, path, expression):
+        return None
```
ebolyen: should this `raise NotImplementedError`?

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,349 @@
...
+
+class ExecutionUsage(Usage):
+    def _action_(self, action: UsageAction,
+                 input_opts: dict, output_opts: dict):
+        action_f, _ = action.get_action()
+        results = action_f(**input_opts)
+        return {k: getattr(results, v) for k, v in output_opts.items()}
+
+    def _assert_has_line_matching_(self, scope_id, label, path, expression):
+        data = self._get_record(scope_id).result
```
ebolyen: I think digging into the `Archiver` here is totally fine, vs saving+extracting. Plus, we have some utility methods/properties which can help.

qiime2/sdk/usage.py
thermokarst: How about `build_opts`?

qiime2/sdk/usage.py
thermokarst: I'm just going to drop the reverse lookup, it isn't strictly necessary, and might just be extra surface area to keep clean.

qiime2/sdk/usage.py
thermokarst: Good catch - this was part of the "stuff that changed between usage and #511" that I mentioned IRL this morning.

qiime2/core/testing/examples.py
```diff
@@ -0,0 +1,110 @@
...
+    right2 = use.get_result('right2')
+    right2.assert_has_line_matching(
+        label='a nice label about this assertion',
+        path='.*/data/ints.txt',
+        expression='1',
+    )
+
+
+def comments_only(use):
+    # TODO
```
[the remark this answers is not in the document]
thermokarst: I'll just add something simple here and test it.

qiime2/plugin/testing.py
```diff
@@ -242,3 +243,11 @@ def transform_format(self, source_format, target, filename=None,
             self.assertIsInstance(obs, target)
 
         return input, obs
+
+    def execute_examples(self):
+        if self.plugin is not None:
```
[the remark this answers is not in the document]
thermokarst: I implemented an explicit exception here (see https://github.com/qiime2/qiime2/pull/523/commits/a21363a92f84668df41406e1a3f6f0e08275e2a3 for more details).

thermokarst:
I am going to work on:

- metadata merging 
- metadata columns

qiime2/sdk/usage.py
thermokarst: Right now there is no real typechecking in the base `Usage` class - `ExecutionUsage` performs typechecking in a trial-by-fire approach, which is better than nothing, but as it stands, unless a usage example is invoked with `ExecutionUsage`, it is "unproven." I think that `build_opts` might be a good place to beef up typechecking, let's chat about this IRL.

thermokarst: Maybe the trial-by-fire isn't so bad. I need to stew on this, since the trade-off is the factory business becomes a little superfluous with the purposes of typechecking (its probably a lot easier to check types with materialized data, rather than factories).

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,368 @@
...
+        return self._result
+
+    @property
+    def is_pending(self):
+        return self.state == self.STATE_PENDING
+
+    def resolve(self):
+        if self.is_pending:
+            self._result = self._factory()
+            self.state = self.STATE_FINISHED
```
ebolyen: It seems like there's a bit of overlap in function between `_result`/`state` and `result`/`resolve()`. It seems like state could be a computed property of result (if you had a UNSET singleton), and then we could worry less about synchronizing `_result` and `is_pending` for `.result`, also, maybe `.result` and `resolve()` could just be merged, they are both R words which could almost do the same thing.  This might simplify the constructor as well.

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,368 @@
...
+    def __init__(self, *, plugin_id: str, action_id: str):
+        if plugin_id == '':
+            raise ValueError('Must specify a value for plugin_id.')
+
+        if action_id == '':
+            raise ValueError('Must specify a value for action_id.')
+
+        self.plugin_id = plugin_id
+        self.action_id = action_id
+        self._plugin_manager = sdk.PluginManager()
```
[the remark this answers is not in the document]
ebolyen: I see why this was done now.

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,368 @@
...
+        override = self._factory_override(ref, factory)
+        return self._push_record(ref=ref, factory=override)
+
+    def merge_metadata(self, ref, *records):
+        if len(records) < 2:
+            # TODO: seems like we should error for this case - thoughts?
+            raise ValueError('')
+
+        factory = self._merge_metadata_(ref, records)
+        return self.init_data(ref, factory)
```
ebolyen: Is there any risk of the `_override_factory` trampling on the intention of `_merge_metadata_`? Since we are returning a closure, I don't think I would anticipate it running through another factory method.

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,368 @@
...
+        except KeyError:
+            raise KeyError('No record with id: "%s" in scope.' % (id,))
+
+
+class Usage(metaclass=abc.ABCMeta):
+    def __init__(self):
+        self._scope = Scope()
+
+    def init_data(self, ref, factory):
+        override = self._factory_override(ref, factory)
```
ebolyen: I think we always want `factory` for input, but it would be nice if `_factory_override` could return a single value, then we can ditch the lambdas that people may forget.

ebolyen: This might also help Execution usages in that `_factory_override` could just run the factory, causing any failures to be tied to right the line of `init_data` in the stack trace, instead of being deferred.

qiime2/sdk/usage.py
ebolyen: I'm imagining a situation where `_merge_metadata_` returns a closure that has a tuple of references, but my `_override_factory` drops the factory and uses `ref`. So now I  have to figure out some kind of type I can check for inside the `_override_factory` to avoid discarding it. It seems conceptually easier if `_merge_metadata_` had the same rules as `_override_factory` (and maybe we call that `_init_data_` like I mentioned elsewhere).

ebolyen: I think this makes sense, because they are both really a de-reference operation (one is just more involved).

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,368 @@
...
+                    match = re.search(expression, target,
+                                      flags=re.MULTILINE)
+                    if match is None:
+                        raise ValueError(
+                            'Expression %r not found in %s.' %
+                            (expression, path))
+
+    def _merge_metadata_(self, ref, records):
+        def closure_fn():
+            # Note: resolved records should be guaranteed at this point
```
ebolyen: I don't think I understand why this is true, so I must be missing some part of the lifecycle. Wouldn't these records still unresolved? Nothing else has looked at them yet.

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,368 @@
...
+    def _comment_(self, text: str):
+        raise NotImplementedError
+
+    def _assert_has_line_matching_(self, scope_id, label, path, expression):
+        raise NotImplementedError
+
+    def _merge_metadata_(self, ref, records):
+        raise NotImplementedError
+
+    def _factory_override(self, ref, factory):
```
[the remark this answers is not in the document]
thermokarst: `_init_data_` sounds good to me.

qiime2/sdk/usage.py
thermokarst: I need to think a bit about this.

qiime2/sdk/usage.py
thermokarst: > It seems conceptually easier if `_merge_metadata_`

I actually had that laying around in my working changeset initially, then changed my mind. I think the change you suggested sounds good, and works really well with the `_init_data_` change above.

qiime2/sdk/usage.py
thermokarst: The closure returned here is deferred until here:

https://github.com/qiime2/qiime2/pull/523/files#diff-50eb2f5e696918079827fa682a7f9fd0R257-R258

The records are resolved in the order that they are put into the scope, so any dependencies are resolved prior.

qiime2/sdk/usage.py
thermokarst: Are you proposing that we construct the factory here? Or just push a "resolved" record right away?

ebolyen: The latter, so if `override` isn't callable, we just take it as resolved.

qiime2/core/testing/plugin.py
```diff
@@ -135,7 +144,10 @@
     name='Concatenate integers',
     description='This method concatenates integers into a single sequence in '
                 'the order they are provided.',
-    citations=[citations['baerheim1994effect']]
+    citations=[citations['baerheim1994effect']],
+    examples={'concatenate_ints_simple': concatenate_ints_simple,
```
wasade: Just curious, why have this as a `dict` rather than a `set` or `tuple`? It looks like the key names are the same as the value names, and (based on a *very* cursory static review) it doesn't seem like the keys relate back to (for example) descriptions as is the case with `parameters`?

qiime2/sdk/tests/test_usage.py
```diff
@@ -0,0 +1,259 @@
...
+        self.plugin = get_dummy_plugin()
+
+    def tearDown(self):
+        self.test_dir.cleanup()
+
+
+class TestUsage(TestCaseUsage):
+    def test_basic(self):
+        action = self.plugin.actions['concatenate_ints']
+        use = usage.DiagnosticUsage()
```
ebolyen: It's like a mock, but not because it's just as valid an implementation as anything else! :zap:

qiime2/core/testing/plugin.py
thermokarst: Hey @wasade!

It is so that you can pluck the example off of the action, like this:

```python
        action = self.plugin.actions['concatenate_ints']
        use = usage.DiagnosticUsage()
        action.examples['concatenate_ints_simple'](use)
```

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,344 @@
...
+import re
+import types
+
+from qiime2 import sdk
+
+# TODO: docstrings
+
+
+class UsageAction:
+    def __init__(self, *, plugin_id: str, action_id: str):
```
ebolyen: Do we need this to be keyword only?  I'm usually all for them, but this seems like a situation where it's pretty obvious which arg is which, and I'm not sure what extensions to this idea would ever come about.

qiime2/core/testing/plugin.py
wasade: ahhhhh! :) thanks!

wasade:
This is extremely cool!!!!

qiime2/sdk/usage.py
ebolyen: seems best to defer this for now.

qiime2/plugins.py
```diff
@@ -19,6 +21,75 @@ def available_plugins():
     return set('qiime2.plugins.' + s.replace('-', '_') for s in pm.plugins)
 
 
+class ArtifactAPIUsage(usage.Usage):
```
ebolyen: This implementation looks very clean, and it is easy to trace the logic.

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,359 @@
...
+        return self._scope.get_record(ref)
+
+    def _get_records(self):
+        return self._scope.records
+
+
+class DiagnosticUsage(Usage):
+    def __init__(self):
+        super().__init__()
+        self._recorder = []
```
ebolyen: Since the goal is to poke the recorded actions, it seems this should be public.

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,359 @@
...
+            'ref': ref,
+            'label': label,
+            'path': path,
+            'expression': expression,
+        })
+
+
+class ExecutionUsage(Usage):
+    def _init_data_(self, ref, factory):
+        return factory()
```
ebolyen: :100:

qiime2/sdk/usage.py
```diff
@@ -0,0 +1,359 @@
...
+        self._recorder.append({
+            'type': 'assert_has_line_matching',
+            'ref': ref,
+            'label': label,
+            'path': path,
+            'expression': expression,
+        })
+
+
+class ExecutionUsage(Usage):
```
ebolyen: This one is also really straight-forward. I think this iteration is working really well!

qiime2/core/testing/plugin.py
ebolyen: We could populate a dict behind the scenes (similar to citations... ish)

ebolyen: I'm good either way though

qiime2/sdk/usage.py
thermokarst: It forces a consistent style across all the `UsageAction/UsageInputs/UsageOutputNames` invocations, which I like:

```python
    use.action(
        UsageAction(plugin_id='dummy_plugin', action_id='typical_pipeline'),
        UsageInputs(int_sequence=ints2, mapping=mapper2, do_extra_thing=False),
        UsageOutputNames(out_map='out_map2', left='left2', right='right2',
                         left_viz='left_viz2', right_viz='right_viz2')
    )
```

vs 

```python
    use.action(
        UsageAction('dummy_plugin', typical_pipeline'),
        UsageInputs(int_sequence=ints2, mapping=mapper2, do_extra_thing=False),
        UsageOutputNames(out_map='out_map2', left='left2', right='right2',
                         left_viz='left_viz2', right_viz='right_viz2')
    )
```

I am open to dropping the keyword-only nature, though.

ebolyen: That seems like a good reason

thermokarst:
Its the end of an era...

PR #526, opened 2020-02-27
BUG: fix behavior of parse_primitive with TypeMap

ebolyen:
When a TypeMap generates a type such as Bool % Choices(False) | Bool,
parse_primitive will will attempt to strip the predicates to recieve a
base type. However this resulted in Bool | Bool which caused the logic
within parse_primitive to fail to recognize the coercion priority giving
us a floating point 0.0 type instead of False.

Normalizing the UnionExp after stripping predicates corrects this issue,
giving us a single Bool type that is recognized by the internal logic.


ebolyen:
I don't know why Travis isn't picking this up... The app is still configured for the repo as far as I can tell.

PR #528, opened 2020-03-26
FIX: Added condition for list and set in build opts

David-Rod:
PR adds checks to ensure that we are able to handle `ScopeRecord` instances within a list or set.

When adding a usage example to `q2-feature-table`, it was discovered that the framework handled lists and `ScopeRecord` types, but could not take in lists or sets with ScopeRecord types as values within that list/set.

This PR adds two conditions that update the `value` variable, using set or list comprehension (depending on the case), to handle such cases.

@thermokarst Is there anything you would recommend for testing these changes moving forward?

Thanks in advance.

### DIFF
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -12,7 +12,7 @@ from qiime2 import Artifact, Metadata, CategoricalMetadataColumn
 
 from qiime2.plugin import UsageAction, UsageInputs, UsageOutputNames
 
-from .type import IntSequence1, IntSequence2, Mapping
+from .type import IntSequence1, IntSequence2, Mapping, SingleInt
 
 
 def ints1_factory():
@@ -48,6 +48,26 @@ def mdc1_factory():
                                      name='a',
                                      index=pd.Index(['0', '1', '2'],
                                                     name='id')))
+
+
+def intSequence_list_factory():
+    int1 = ints1_factory()
+    int2 = ints2_factory()
+    return [int1, int2]
+
+
+def singleInt1_factory():
+    return Artifact.import_data(SingleInt, 10)
+
+
+def singleInt2_factory():
+    return Artifact.import_data(SingleInt, 11)
+
+
+def singleInt1_set_factory():
+    singleInt1 = singleInt1_factory()
+    singleInt2 = singleInt2_factory()
+    return {singleInt1, singleInt2}
 
 
 def concatenate_ints_simple(use):
@@ -180,4 +200,16 @@ def identity_with_metadata_column_from_factory(use):
                     action_id='identity_with_metadata_column'),
         UsageInputs(ints=ints, metadata=mdc),
         UsageOutputNames(out='out'),
+    )
+
+
+def feature_table_merge_example(use):
+    ints = use.init_data('int', intSequence_list_factory)
+    int_set = use.init_data('int_set', singleInt1_set_factory)
+
+    use.action(
+        UsageAction(plugin_id='dummy_plugin',
+                    action_id='variadic_input_method'),
+        UsageInputs(ints=ints, int_set=int_set, nums={7, 8, 9}),
+        UsageOutputNames(output='out'),
     )
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -50,24 +50,24 @@ def mdc1_factory():
                                                     name='id')))
 
 
-def intSequence_list_factory():
+def int_sequence_list_factory():
     int1 = ints1_factory()
     int2 = ints2_factory()
     return [int1, int2]
 
 
-def singleInt1_factory():
+def single_int1_factory():
     return Artifact.import_data(SingleInt, 10)
 
 
-def singleInt2_factory():
+def single_int2_factory():
     return Artifact.import_data(SingleInt, 11)
 
 
-def singleInt1_set_factory():
-    singleInt1 = singleInt1_factory()
-    singleInt2 = singleInt2_factory()
-    return {singleInt1, singleInt2}
+def single_int_set_factory():
+    single_int1 = single_int1_factory()
+    single_int2 = single_int2_factory()
+    return {single_int1, single_int2}
 
 
 def concatenate_ints_simple(use):
@@ -204,8 +204,8 @@ def identity_with_metadata_column_from_factory(use):
 
 
 def feature_table_merge_example(use):
-    ints = use.init_data('int', intSequence_list_factory)
-    int_set = use.init_data('int_set', singleInt1_set_factory)
+    ints = use.init_data('int', int_sequence_list_factory)
+    int_set = use.init_data('int_set', single_int_set_factory)
 
     use.action(
         UsageAction(plugin_id='dummy_plugin',
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -203,7 +203,7 @@ def identity_with_metadata_column_from_factory(use):
     )
 
 
-def feature_table_merge_example(use):
+def variadic_input_simple(use):
     ints = use.init_data('int', int_sequence_list_factory)
     int_set = use.init_data('int_set', single_int_set_factory)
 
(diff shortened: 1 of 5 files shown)


thermokarst:
Hey @David-Rod - thanks for this PR, and thanks for filling me in on the details!

## Dummy Plugin Usage Examples

For testing, you will need to create a few new usage examples in the dummy plugin, drop those here:

https://github.com/qiime2/qiime2/blob/master/qiime2/core/testing/examples.py

Don't forget to register those examples in the dummy plugin (like you did with the examples you made for q2-feature-table).

Looks like this method might let you address `Set` and `List` in one shot:

https://github.com/qiime2/qiime2/blob/f89eaadf693c85874dc9fae4017a22b5ddcf7ed0/qiime2/core/testing/plugin.py#L437-L463

## Unit Tests

Once you have some usage examples in place, you can write some unit tests:

https://github.com/qiime2/qiime2/blob/f89eaadf693c85874dc9fae4017a22b5ddcf7ed0/qiime2/sdk/tests/test_usage.py#L28

You are mostly just wanting to run your usage example(s), with the `DiagnosticUsage` driver (see the other tests in that suite) and make sure that everything looks good. Think about your q2-feature-table examples while doing this - keep in mind the things that prompted you to write this PR in the first place - those are good things to check for!

Also, please drop in a test or two in this section:

https://github.com/qiime2/qiime2/blob/f89eaadf693c85874dc9fae4017a22b5ddcf7ed0/qiime2/sdk/tests/test_usage.py#L191

You won't need any usage examples there, but basically you should add two tests like https://github.com/qiime2/qiime2/blob/f89eaadf693c85874dc9fae4017a22b5ddcf7ed0/qiime2/sdk/tests/test_usage.py#L205, the first should check for a list, the other for a set.

## Misc

Other miscellaneous things: I don't think any changes need to be made to the other Usage drivers in this repo, but it might be worth duplicating your test that you write for `DiagnosticUsage`:

https://github.com/qiime2/qiime2/blob/f89eaadf693c85874dc9fae4017a22b5ddcf7ed0/qiime2/sdk/tests/test_usage.py#L28

in `ArtifactAPIUsage`, just to cover our bases:

https://github.com/qiime2/qiime2/blob/f89eaadf693c85874dc9fae4017a22b5ddcf7ed0/qiime2/tests/test_artifact_api.py#L122

## Summary

1. Create some new usage examples in the dummy plugin
2. Add some unit tests that exercise your examples with `DiagnosticUsage`
3. Add some unit tests that verify the basic functionality you added to `UsageInputs` (no need to use your usage examples)
4. Duplicate tests in step 2 for `ArtifactAPIUsage`

Please let me know if you need clarification or have any questions! I am happy to help out, just let me know. Thanks!

qiime2/sdk/usage.py
```diff
@@ -88,6 +88,14 @@ def build_opts(self, signature, scope):
                 if isinstance(self.values[name], ScopeRecord) \
                         and self.values[name].ref in scope.records:
                     value = self.values[name].result
+                elif (isinstance(self.values[name], list)
+                      and self.values[name]
+                      and isinstance(self.values[name][0], ScopeRecord)):
```
thermokarst: Perhaps for a "belt-and-suspenders" approach here, we should consider checking that _all_ of the records in the collection are `ScopeRecord` instances. The same consideration applies to the `set` branch below.

David-Rod: Would something like the following work to check the records in the collections mentioned above:

`all(isinstance(n, ScopeRecord) for n in self.values[name]))`

I've tested this solution for both set and list types that contain string values, and they both seem to work.

thermokarst: Works for me!

David-Rod:
Hey @thermokarst,

I believe I was able to accomplish the items you listed above. I am slightly unsure about the tests that I created, but I think most of what you may be looking for is here. 

Thanks in advance for your review!

thermokarst:
Thanks @David-Rod - will take a look soon!

qiime2/core/testing/examples.py
```diff
@@ -50,6 +50,26 @@ def mdc1_factory():
                                                     name='id')))
 
 
+def intSequence_list_factory():
```
thermokarst, line 53: Please reformat this as a [PEP 8-compliant name](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)

```suggestion
def int_sequence_list_factory():
```

qiime2/core/testing/examples.py
```diff
@@ -50,6 +50,26 @@ def mdc1_factory():
...
+    int1 = ints1_factory()
+    int2 = ints2_factory()
+    return [int1, int2]
+
+
+def singleInt1_factory():
```
thermokarst, line 59: PEP 8 (see above)

qiime2/core/testing/examples.py
```diff
@@ -50,6 +50,26 @@ def mdc1_factory():
...
+    return Artifact.import_data(SingleInt, 10)
+
+
+def singleInt2_factory():
```
thermokarst, line 63: PEP 8 (see above)

qiime2/core/testing/examples.py
```diff
@@ -50,6 +50,26 @@ def mdc1_factory():
...
+    return Artifact.import_data(SingleInt, 11)
+
+
+def singleInt1_set_factory():
```
thermokarst, line 67: PEP 8 (see above)

qiime2/sdk/tests/test_usage.py
```diff
@@ -231,6 +248,13 @@ def test_validated_missing_optional_value(self):
         ui.validate(self.signature)
         self.assertTrue(True)
 
+    def test_type_of_input(self):
+        test_inputs = usage.UsageInputs(x=[1, 2, 3], z={7, 8, 9})
+        test_inputs.validate(self.signature)
+
+        self.assertEqual(list, type(test_inputs.values['x']))
+        self.assertEqual(set, type(test_inputs.values['z']))
```
thermokarst, line 256: Great! A slightly more idiomatic approach would use https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsInstance

```suggestion
        self.assertIsInstance(test_inputs.values['x'], list)
        self.assertIsInstance(test_inputs.values['z'], set)
```

qiime2/tests/test_artifact_api.py
```diff
@@ -252,6 +252,22 @@ def test_metadata_column_from_factory(self):
...
+    def test_use_merge_feature_table(self):
+        action = self.plugin.actions['variadic_input_method']
+        use = ArtifactAPIUsage()
+        action.examples['feature_table_merge_example'](use)
+
+        exp = """\
+from qiime2.plugins.dummy_plugin.methods import variadic_input_method
+
+out, = variadic_input_method(
+    ints=int,
```
thermokarst, line 267: Note to @thermokarst - this driver should template in the factory body to make this more clear.

David-Rod:
Regarding the changes for the latest commit, I made a slight change to the validation for inputs that are provided to accommodate the changes made in https://github.com/qiime2/q2-feature-table/pull/226. I know this change does not capture all possibilities when we are trying to provide more than one input, but it may be a step in the right direction. Hopefully your review will provide some more clarity. Thanks.

qiime2/sdk/usage.py
```diff
@@ -77,8 +77,10 @@ def validate(self, signature):
 
         extra = provided - exp_inputs - exp_params
         if len(extra) > 0:
-            raise ValueError('Extra input(s) or parameter(s): %r' %
-                             (extra, ))
+            for provided_value in extra:
```
thermokarst, line 80: @David-Rod - I suspect you won't need any of this once https://github.com/qiime2/qiime2/pull/530 is merged.

thermokarst:
This is blocked by #530

thermokarst:
Hi @David-Rod, this PR should be unblocked, but looks like you will need to merge in the latest master and deal with the merge conflicts. Please let me know if you need a hand with that, I am happy to help.

thermokarst:
Hey @David-Rod - don't worry about the merge conflicts - I have some free time and can handle it, stay tuned.

PR #529, opened 2020-03-30
IMP: Invert UsageOutputNames SDK API

thermokarst:
@andrewsanchez pointed out to me that there was some clumsy inversion of SDK API in `sdk.usage.UsageOutputNames` - this PR attempts to reconcile that. Now usage drivers can expect to receive a dict of action inputs (parameter name as key, value as, ahem, the value), and a dict of action outputs (parameter name as key, scope ref name as value). I think this makes it a bit simpler to reason about, as an SDK dev.

### DIFF
--- a/qiime2/sdk/usage.py
+++ b/qiime2/sdk/usage.py
@@ -123,9 +123,9 @@ class UsageOutputNames:
         if len(extra) > 0:
             raise ValueError('Extra output(s): %r' % (extra, ))
 
-    def validate_derived(self, derived_outputs):
-        provided = set(derived_outputs.keys())
-        exp_outputs = set(self.values.values())
+    def validate_computed(self, computed_outputs):
+        provided = set(computed_outputs.keys())
+        exp_outputs = set(self.values.keys())
 
         missing = exp_outputs - provided
         if len(missing) > 0:
@@ -141,7 +141,7 @@ class UsageOutputNames:
         opts = {}
 
         for output in action_signature.outputs.keys():
-            opts[self.get(output)] = output
+            opts[output] = self.get(output)
 
         return opts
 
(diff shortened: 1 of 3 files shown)


qiime2/sdk/usage.py
```diff
@@ -240,11 +240,11 @@ def action(self, action: UsageAction, inputs: UsageInputs,
         input_opts = inputs.build_opts(action_signature, self._scope)
         output_opts = outputs.build_opts(action_signature, self._scope)
 
-        derived_outputs = self._action_(action, input_opts, output_opts)
-        self._add_outputs_to_scope(outputs, derived_outputs)
+        computed_outputs = self._action_(action, input_opts, output_opts)
```
andrewsanchez: Does `_action_` still need to return `computed_outputs` here?  It doesn't look like `output_opts` is actually being modified by calls to `_action_ ` in `ArtifactAPIUsage` anymore.  It's not  being modified in `CLIUsage` either.

thermokarst: > Does `_action_` still need to return `computed_outputs` here?

Yep! The new schema for `computed_outputs` is `Dict[str, Any]` where the key is name out the registered output on the action, and the value is whatever should be stored in the scope - it can be a computed value, or just some kind of placeholder.

>  It doesn't look like `output_opts` is actually being modified by calls to `_action_ ` in `ArtifactAPIUsage` anymore.

It doesn't necessarily need to be modified, but it does need to be returned. What we are doing in `ArtifactAPIUsage` is returning a `Dict[str, str]` where the keys are the output names and the values are the scope ref names.

qiime2/sdk/usage.py
```diff
@@ -253,9 +253,10 @@ def _assert_has_line_matching_(self, ref, label, path, expression):
     def get_result(self, ref):
         return self._get_record(ref)
 
-    def _add_outputs_to_scope(self, outputs, derived_outputs):
-        outputs.validate_derived(derived_outputs)
-        for ref, result in derived_outputs.items():
+    def _add_outputs_to_scope(self, outputs, computed_outputs):
+        outputs.validate_computed(computed_outputs)
```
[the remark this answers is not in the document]
thermokarst: > Will this ever actually raise `ValueError`?

It could if an implementation of a Usage driver did it wrong.

The check is there for SDK/Interface developers - in this case folks like you, implementing a Usage driver.

If you mess up the return from `_action_`, this guard might help you get set on the right course.

PR #531, opened 2020-04-04
TESTING: Decorated actions exploration2

### LINKED ISSUE #524: decorated Actions with default params fail
**Bug Description**
Actions that have default parameters fail with `KeyError: 'name_of_first_default_param'` if those Actions have had decorators functions applied to them.

This is caused by a framework bug, where default arguments are unintentionally stripped (in `Action._rewrite_wrapper_signature()`?), causing `Action.bound_callable()` to pass truncated argument lists to `self.signature.check_types()`. `check_types()` fails while attempting to access kwargs that still exist in `self.signature.order_items` which have been stripped from the kwargs it received.

1. Check out the following branches in a dev environment: 
https://github.com/ChrisKeefe/q2-diversity/tree/divLibWiring
https://github.com/ChrisKeefe/q2-diversity-lib/tree/bDiv
2. `make dev` (it _may_ be necessary to do so in both repos?)
3. navigate to q2-diversity, and `make test`

This should wire `q2-diversity` to use decorated Actions that are implemented in `q2-diversity-lib`, and raise relevant errors during testing.

Alternately, you could write a simple decorator function, apply it to an existing Action in any repo _that has one or more default parameters_, and test. If results vary, please let us know below.

**Expected behavior**
`KeyError: 'name_of_first_default_param' will be raised when calling decorated Actions with default parameters.

**Computation Environment**
- OS: LinuxMint 19.2, likely all others.
- QIIME 2 2019.10

**Questions**
1. Is it best for us to generate a new signature, rather than continuing to abuse the decorator library in manipulating wrapper signatures?

**References**

1. [decorator](https://github.com/micheles/decorator/blob/master/src/decorator.py) library
2. [check_types](https://github.com/qiime2/qiime2/blob/90f7bbf7c9bce673c6c2809f6ddfcd56ca3eee4c/qiime2/core/type/signature.py#L323)
3. [bound_callable](https://github.com/qiime2/qiime2/blob/90f7bbf7c9bce673c6c2809f6ddfcd56ca3eee4c/qiime2/sdk/action.py#L185)
4. [rewrite_wrapper_signature](https://github.com/qiime2/qiime2/blob/90f7bbf7c9bce673c6c2809f6ddfcd56ca3eee4c/qiime2/sdk/action.py#L299)
5.  Branches where the bug first materialized: 
     - https://github.com/ChrisKeefe/q2-diversity/tree/divLibWiring
     - https://github.com/ChrisKeefe/q2-diversity-lib/tree/bDiv

ChrisKeefe:
This attempts to fix #524. 

Concern 1: Select integration tests are failing locally, but all of those are failing when biom receives BIOMV210Format objects instead of file paths. Unless this update interferes with the framework's type handling, I suspect that issue is not directly related to the decorated actions problem. 

Concern 2: Is it problematic that I'm _not_ using the parameter names from registration in plugin_setup.py, instead coupling directly to the function signatures themselves? 

@ebolyen, I'd love your eyes on this before I dive too deep into troubleshooting the biom issues it raised in other repos.

### DIFF
--- a/qiime2/sdk/action.py
+++ b/qiime2/sdk/action.py
@@ -19,6 +19,7 @@ import qiime2.sdk
 import qiime2.core.type as qtype
 import qiime2.core.archive as archive
 from qiime2.core.util import LateBindingAttribute, DropFirstParameter, tuplize
+from inspect import signature
 
 
 def _subprocess_apply(action, args, kwargs):
@@ -189,20 +190,8 @@ class Action(metaclass=abc.ABCMeta):
             # `decorator.decorator`. When the signature is rewritten,
             # args[0] is the function whose signature was used to rewrite
             # this function's signature.
-            # TODO: remove
-            print("********* function used to rewrite sig", args[0])
-            import inspect
-            from inspect import signature
-            TODO_valid_sig = signature(args[0])
-            print("********* this function's sig", TODO_valid_sig)
-            func = args[0]
-            # if hasattr(func, '__wrapped__'):
-            while hasattr(func, '__wrapped__'):
-                func = func.__wrapped__
-                print("************wrapped_func", func)
-                print("********* wrapped func sig",
-                      signature(func))
 
+            func = args[0]
             args = args[1:]
             ctx = context_factory()
             # Set up a scope under which we can track destructable references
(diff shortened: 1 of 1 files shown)


thermokarst:
Please see https://github.com/qiime2/qiime2/issues/524#issuecomment-626108126

PR #539, opened 2020-06-09
ENH: Method for initializing data collections

andrewsanchez:


qiime2/plugins.py
```diff
@@ -37,6 +37,15 @@ def _init_metadata_(self, ref, factory):
         self._init_data_refs[ref] = factory
         return ref
 
+    def _init_data_collection_(self, ref, records):
```
andrewsanchez, line 40: Does it make sense to simplify the signature like this?  You originally suggested

```diff
+    def init_data_collection(self, ref, container_type, *records):
+        if len(records) < 2:
+            raise ValueError('Must provide two or more Artifact inputs.')
+
+        value = self._init_data_collection_(ref, container_type, records)
+        return self._push_record(ref, value, 'init_data_collection')
```

thermokarst: Thanks for proposing! I don't think it'll get us what we need, though. I will comment directly on the method defined in the base class.

qiime2/sdk/usage.py
```diff
@@ -238,6 +238,16 @@ def init_metadata(self, ref, factory):
     def _init_metadata_(self, ref, factory):
         raise NotImplementedError
 
+    def init_data_collection(self, ref, records):
```
thermokarst, line 241: I don't think that this approach will work, since you are passing in the factories now, rather than `ScopeRecord`s. I think we need scope records, because those will have `ref`s associated with them, which will let us name the things in our various implementations. Without that info, I think you'll be left guessing _what_ to name them in the driver implementations.

andrewsanchez: I will push up another iteration rather than discussing here.  I think that will be easier.

qiime2/core/testing/method.py
```diff
@@ -117,7 +117,7 @@ def optional_artifacts_method(ints: list, num1: int, optional1: list = None,
     return result
 
 
-def variadic_input_method(ints: list, int_set: int, nums: int,
+def variadic_input_method(ints: list, int_set: set, nums: int,
```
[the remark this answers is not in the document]
thermokarst: The correct type annotation is `int`. We don't currently have a way to communicate the view's container type back to the framework, so instead we register the view type of the elements, rather than the container.

```suggestion
def variadic_input_method(ints: list, int_set: int, nums: int,
```

qiime2/tests/test_artifact_api.py
```diff
@@ -250,7 +250,7 @@ def test_use_merge_feature_table(self):
 from qiime2.plugins.dummy_plugin.methods import variadic_input_method
 
 out, = variadic_input_method(
-    ints=int,
+    ints=ints,
     int_set=int_set,
```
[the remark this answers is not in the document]
thermokarst: Unrelated, can you fix up the name of this test (`test_use_merge_feature_table`) - that's a copy-paste error. Thanks!

qiime2/sdk/tests/test_usage.py
```diff
@@ -149,19 +149,29 @@ def test_use_merge_feature_table(self):
         use = usage.DiagnosticUsage()
         action.examples['variadic_input_simple'](use)
 
-        self.assertEqual(len(use.recorder), 3)
+        self.assertEqual(len(use.recorder), 7)
```
thermokarst, line 152: Posting here because its the closest I can get to the line of interest, but can you rename this test, its a copy-paste error from ye olde days.

qiime2/sdk/usage.py
```diff
@@ -327,6 +337,13 @@ def _init_metadata_(self, ref, factory):
         })
         return ref
 
+    def _init_data_collection_(self, ref, container_type, *records):
+        self.recorder.append({
+            'source': 'init_data_collection',
+            'ref': ref,
+        })
+        return ref, [i.ref for i in records]
```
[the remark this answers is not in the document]
andrewsanchez: Good catch!

qiime2/sdk/usage.py
```diff
@@ -327,6 +337,13 @@ def _init_metadata_(self, ref, factory):
         })
         return ref
 
+    def _init_data_collection_(self, ref, container_type, *records):
```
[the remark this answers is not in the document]
andrewsanchez: I think `init_data_collection` is fine.  I don't see any pressing reasons to change it, but I like your previous suggestions better, i.e. `combine_collection_data` or maybe `merge_collection_data`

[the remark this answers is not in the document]
thermokarst: Thanks @andrewsanchez! What about the argument `container_type`? That's what I was asking about above (but I am glad you brought up the function name, too - I think sticking with `init_data_collection` makes sense to me).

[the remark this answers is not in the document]
andrewsanchez: Oh, I see!  Sorry, totally misunderstood. I think `collection_type` is n improvement since it matches up with the function name.  Happy to go with that!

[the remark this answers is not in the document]
thermokarst: Let's go with `collection_type`, then! Thanks!

qiime2/plugins.py
```diff
@@ -37,6 +37,13 @@ def _init_metadata_(self, ref, factory):
         self._init_data_refs[ref] = factory
         return ref
 
+    def _init_data_collection_(self, ref, collection_type, *records):
+        self._init_data_refs[ref] = ref
+        collection = sorted([i.ref for i in records])
+        collection = collection_type(collection)
+        t = str(collection).replace("'", "")
+        return t
```
andrewsanchez, line 45: Do you know, off the top of your head, how to consistently sort all possible inputs to `sorted` here?  I keep getting inconsistent test results, e.g. `{single_int1, single_int2}` and then a failing test case when collection ends up being `{single_int2, single_int1}`. [This](https://www.tutorialspoint.com/How-to-correctly-sort-a-string-with-a-number-inside-in-Python) didn't work, but I haven't dug any deeper.  Thought I'd check with you first.

qiime2/sdk/tests/test_usage.py
```diff
@@ -380,6 +390,10 @@ def test_init_data(self):
         with self.assertRaisesRegex(TypeError, 'expected Metadata'):
             use.init_metadata('name', lambda: object)
 
+        with self.assertRaisesRegex(ValueError, 'expected a ScopeRecord.'):
+            use.init_data_collection('', list, object)
+            use.init_data_collection('', list, object, object)
+
```
andrewsanchez, line 396: Will this do the trick, @thermokarst?

qiime2/plugins.py
thermokarst: I think the issue is you're going from a sorted Iterable (list, L42), no an unsorted Iterable (set, L43). Set's don't have an order (https://docs.python.org/3.6/library/stdtypes.html#set-types-set-frozenset):

> A set object is an unordered collection of distinct hashable objects.

---

I think you might have to get a little more involved in the templating step here, maybe repr out a sorted list, then do a string substitution to replace square brackets with curlies?

Alternatively, you could template out something like this:

```python
x=set(['a', 'b', 'c', 'd'])
```

that uses the explicit set/list constructor.

qiime2/sdk/tests/test_usage.py
thermokarst: For checking the type validation of the inputs? Yes I think so! Although, you'll need to tweak a bit. First, you can't put both failing calls to `init_data_collection` in one `self.assertRaisesRegex` context: the second call to `init_data_collection` will never run (because the first raises an exception. Second, the multiple-object example might be more useful if it has a ScopeRecord first, then an object, to test that we are checking all the values:

```suggestion
        with self.assertRaisesRegex(ValueError, 'expected a ScopeRecord.'):
            use.init_data_collection('', list, object)
        with self.assertRaisesRegex(ValueError, 'expected a ScopeRecord.'):
            use.init_data_collection('', list, ScopeRecord(...), object)
```

qiime2/plugins.py
andrewsanchez: Oh, duhhh,  I don't even need the actual container type here 🤦‍♂️

qiime2/plugins.py
```diff
@@ -37,6 +37,15 @@ def _init_metadata_(self, ref, factory):
         self._init_data_refs[ref] = factory
         return ref
 
+    def _init_data_collection_(self, ref, collection_type, *records):
+        self._init_data_refs[ref] = ref
+        collection = sorted([i.ref for i in records])
+        t = str(collection).replace("'", "")
```
thermokarst, line 43: Nice! How about this modification?

```suggestion
        t = ', '.join(sorted([i.ref for i in records])
```

You'll have to tweak how you template in the set brackets, though...

qiime2/plugins.py
```diff
@@ -37,6 +37,15 @@ def _init_metadata_(self, ref, factory):
...
+        if collection_type is set:
+            translation = t.maketrans("[]", "{}")
```
thermokarst, line 45: I learned something new today! Thanks so much for sharing this!

qiime2/sdk/usage.py
```diff
@@ -396,6 +413,17 @@ def _init_metadata_(self, ref, factory):
 
         return result
 
+    def _init_data_collection_(self, ref, collection_type, *records):
+        collection = []
+        for record in records:
+            result_type = type(record)
+            if not isinstance(record, ScopeRecord):
+                raise ValueError('Record (%r) returned a %s, expected a '
+                                 'ScopeRecord.' % (record, result_type))
```
thermokarst, line 422: What do you think about putting this check up in the base class?

andrewsanchez: Would you also want to move the other similar checks on the init data methods of `ExecutionUsage` into the base class?

thermokarst: What checks did you have in mind? I'm not seeing any others, but please drop a link or two in to set my straight.

qiime2/sdk/usage.py
```diff
@@ -396,6 +413,17 @@ def _init_metadata_(self, ref, factory):
 
         return result
 
+    def _init_data_collection_(self, ref, collection_type, *records):
```
thermokarst, line 416: Please remove the guard on https://github.com/qiime2/qiime2/pull/539/files#diff-50eb2f5e696918079827fa682a7f9fd0R398-R402

qiime2/plugins.py
```diff
@@ -37,6 +37,13 @@ def _init_metadata_(self, ref, factory):
         self._init_data_refs[ref] = factory
         return ref
 
+    def _init_data_collection_(self, ref, collection_type, *records):
+        self._init_data_refs[ref] = ref
+        refs = sorted([i.ref for i in records])
+        t = ', '.join(refs).strip(',')
```
thermokarst, line 43: The `strip` should be unnecessary.

```suggestion
        t = ', '.join(refs)
```

qiime2/plugins.py
```diff
@@ -37,6 +37,13 @@ def _init_metadata_(self, ref, factory):
...
+        t = f"[{t}]" if collection_type is list else f"{{{t}}}"
```
thermokarst, line 44: Please convert this to c-style `%` strings, to match the rest of the module.

qiime2/plugins.py
```diff
@@ -37,6 +37,13 @@ def _init_metadata_(self, ref, factory):
         self._init_data_refs[ref] = factory
         return ref
 
+    def _init_data_collection_(self, ref, collection_type, *records):
+        self._init_data_refs[ref] = ref
```
thermokarst, line 41: I don't think we need this line, since the individual records in the collection have already been registered. The `_init_data_refs` attribute is for tracking data that we might need to materialize, in case the user asks for it (via the `get_example_data()` method).

qiime2/sdk/usage.py
andrewsanchez: You can disregard my comment.  I was confused.

thermokarst: > I was confused.

story of my life.

qiime2/plugins.py
andrewsanchez: Hmmm, weird.  I had a test case failing that showed me the `strip` was necessary, but I'm sure what happened.  Fixed!

qiime2/plugins.py
andrewsanchez: right-o! 🌭

qiime2/sdk/tests/test_usage.py
```diff
@@ -380,9 +390,28 @@ def test_init_data(self):
...
         md1 = use.init_metadata('md1', examples.md1_factory)
         md2 = use.init_metadata('md2', examples.md2_factory)
         merged = use.merge_metadata('md3', md1, md2)
         self.assertIsInstance(merged.result, Metadata)
+
+    def test_variadic_input_simple(self):
+        use = usage.ExecutionUsage()
+        action = self.plugin.actions['variadic_input_method']
+        action.examples['variadic_input_simple'](use)
+        for record in use._get_records().values():
```
thermokarst, line 411: Rather than iterating through, can you specifically look up the inputs and the outputs?

```python
int_list_record = use.get_result('int')
int_set_record = use.get_result('int_set')
out_record = use.get_result('out')

...
```

My concern here is that the loop doesn't explicitly check for our _known_ records. This is a testing pitfall that I have gotten sucked into before...

andrewsanchez: Not sure why, but `Usage.get_result` only works if `record.source == 'action'`.  How does 23ef3a5 this look?

thermokarst: That's right, I forgot that `get_result` is "protected".

The tests in 23ef3a5 look good, but you're still effectively looping over the scope records, rather than explicitly evaluating them one by one.

```python
int_list_record = use._get_record('int')
...
```

Make sense?

thermokarst: Alternatively, you could stick with the unpacking, and add in additional assertions about the scoperecord's refs:

```python
self.assertEqual(ints_a.ref, 'ints_a')
self.assertIsInstance(ints_a.result, Artifact)
...
```

qiime2/sdk/usage.py
```diff
@@ -238,6 +238,21 @@ def init_metadata(self, ref, factory):
...
         raise NotImplementedError
 
+    def init_data_collection(self, ref, collection_type, *records):
+        if len(records) < 1:
+            raise ValueError('Must provide at least one ScopeRecord input.')
+        for record in records:
+            result_type = type(record)
+            if not isinstance(record, ScopeRecord):
+                raise ValueError('Record (%r) returned a %s, expected a '
+                                 'ScopeRecord.' % (record, result_type))
```
thermokarst, line 248:
```suggestion
            if not isinstance(record, ScopeRecord):
                raise ValueError('Record (%r) returned a %s, expected a '
                                 'ScopeRecord.' % (record, type(result)))
```

qiime2/sdk/usage.py
```diff
@@ -238,6 +238,20 @@ def init_metadata(self, ref, factory):
     def _init_metadata_(self, ref, factory):
         raise NotImplementedError
 
+    def init_data_collection(self, ref, collection_type, *records):
+        if len(records) < 1:
+            raise ValueError('Must provide at least one ScopeRecord input.')
+        for record in records:
+            if not isinstance(record, ScopeRecord):
+                raise ValueError('Record (%r) returned a %s, expected a '
+                                 'ScopeRecord.' % (record, type(result)))
```
thermokarst, line 247:
```suggestion
                                 'ScopeRecord.' % (record, type(record)))
```

PR #542, opened 2020-06-23
MAINT: Update error message to point out common mistake

### LINKED ISSUE #537: tabulate fails inscrutably when passed a csv
**Bug Description**
`qiime metadata tabulate` produces an `unrecognized ID column name` error when passed a .csv, because it is disregarding the commas as separators and gluing all columns together. This is confusing to newer users.

1. Download and unzip this guy: https://data.qiime2.org/2020.2/tutorials/read-joining/fj-joined.zip
2. In a QIIME 2 environment, run:
```
qiime metadata tabulate --m-input-file fj-joined/manifest --o-visualization metadata
```
3. Error, apparently caused by metadata combining all comma-separated columns into one.

**Expected behavior**
Either
1. Error, saying metadata tabulate only works with `.tsv`
2. Gracefully handle `.csv` tabulation, silently, or by adding an additional parameter that defaults to `.tsv`

**Screenshots**
image

**Computation Environment**
- OS: LinuxMint 19.3
- QIIME 2 2020.2

**Questions**
1. Are old (V1) `SingleEndFastqManifestPhred33` files are the only thing blowing this up? If so, a good error message might be sufficient. 

**References**
[forum xref](https://forum.qiime2.org/t/problems-about-analyzing-joined-reads-using-qiime2/14950)

Oddant1:
Closes #537. It seemed to me like the simplest way to resolve this issue would be adding a short note that metadata files must be tab and not comma separated. Hopefully people will be able to see this note and quickly realize what the issue is if they attempt to use a csv as metadata.

### DIFF
--- a/qiime2/metadata/io.py
+++ b/qiime2/metadata/io.py
@@ -144,7 +144,9 @@ class MetadataReader:
                 raise MetadataFileError(
                     "Found unrecognized ID column name %r while searching for "
                     "header. The first column name in the header defines the "
-                    "ID column, and must be one of these values:\n\n%s" %
+                    "ID column, and must be one of these values:\n\n%s\n\n"
+                    "NOTE: Metadata files must be tab separated not comma "
+                    "separated." %
                     (row[0], FORMATTED_ID_HEADERS))
 
         if header is None:
--- a/qiime2/metadata/io.py
+++ b/qiime2/metadata/io.py
@@ -145,8 +145,7 @@ class MetadataReader:
                     "Found unrecognized ID column name %r while searching for "
                     "header. The first column name in the header defines the "
                     "ID column, and must be one of these values:\n\n%s\n\n"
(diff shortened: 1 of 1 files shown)


qiime2/metadata/io.py
```diff
@@ -144,7 +144,9 @@ def _read_header(self):
                 raise MetadataFileError(
                     "Found unrecognized ID column name %r while searching for "
                     "header. The first column name in the header defines the "
-                    "ID column, and must be one of these values:\n\n%s" %
+                    "ID column, and must be one of these values:\n\n%s\n\n"
+                    "NOTE: Metadata files must be tab separated not comma "
+                    "separated." %
```
thermokarst, line 149:
```suggestion
                    "NOTE: Metadata files must contain tab-separated values." %
```

PR #536, opened 2020-05-12
IMP: add source-tracking to Usage API

thermokarst:

PR #546, opened 2020-07-20
Clean up the usage module

### LINKED ISSUE #544: Clean up usage module
**Bug Description**
Some usage methods have type annotations, some don't. Nothing has docstrings. `_init_data_collection_` appears to have an errant `*` on the records attr (although I might need to look more closely at this). In general this module could use a nice tidying-up.

andrewsanchez:
- [x] documentation
- [x] type hints
- [x] misc clean up
- [x] **TODO: Remove changes to importing logic before merging**

Closes #544.

Initial commit is a WIP, chipping away at docstrings and type hints.


thermokarst:
I'm putting this on the 2020.8 project board, but no rush @andrewsanchez - if it doesn't make the release, no big deal.

andrewsanchez:
Including changes proposed in #549 for Travis.  I can re-write the history if you prefer to keep these change sets discreet.  Still a WIP, but it would be nice to get a quick review, @thermokarst, just to make sure I'm on the right track.  This has been incredibly useful for understanding the Usage API and the framework a little better. 😄

qiime2/sdk/usage.py
```diff
@@ -8,60 +8,194 @@
...
+    def __init__(self, ref: str, value: ActionData, source: str,
+                 assert_has_line_matching: Optional[Callable] = None):
+        """
+        An object for recording information needed by Usage drivers to render
+        Usage examples for QIIME 2 interfaces.
+
+        Parameters
+        ----------
+        ref: str
+            A name for referring to `value`.
```
thermokarst, line 35:
```suggestion
            A unique name for referring to `value`.
```

qiime2/sdk/usage.py
```diff
@@ -8,60 +8,194 @@
 
 import abc
 import re
-import types
+from types import MappingProxyType
+from typing import Tuple, Callable, Optional, Union, List, Set, Dict
```
thermokarst, line 12: Stylistically, I prefer `import types`, then using `types.MappingProxyType` - I like this approach (esp with external deps) because it shows where the object comes from, inline.

qiime2/sdk/usage.py
```diff
@@ -8,60 +8,194 @@
...
 
 from qiime2 import sdk, metadata
+from qiime2.core.type import MethodSignature, PipelineSignature
 
-# TODO: docstrings
+# Is Metadata sufficient for typing, or should we specify MetadataColumn
```
thermokarst, line 17: Metadata is sufficient, since we don't allow MetadataColumn as a primitive in usage API.

qiime2/sdk/usage.py
```diff
@@ -8,60 +8,194 @@
...
+                not callable(assert_has_line_matching):
+            raise TypeError('Value for `assert_has_line_matching` should be a '
+                            '`callable`.')
 
-        self.plugin_id = plugin_id
-        self.action_id = action_id
-        self._plugin_manager = sdk.PluginManager()
+        self.ref = ref
+        self._result = value
+        # TODO: Should we put a guard somewhere for acceptable sources?
```
thermokarst, line 53: Wouldn't hurt! Good idea.

qiime2/sdk/usage.py
```diff
@@ -8,60 +8,194 @@
...
+from types import MappingProxyType
+from typing import Tuple, Callable, Optional, Union, List, Set, Dict
 
 from qiime2 import sdk, metadata
+from qiime2.core.type import MethodSignature, PipelineSignature
 
-# TODO: docstrings
+# Is Metadata sufficient for typing, or should we specify MetadataColumn
+# as well?
+Factory = Callable[..., Union[metadata.Metadata, sdk.Artifact]]
```
thermokarst, line 19: Does python not blow up on this ellipsis??!?!

qiime2/sdk/usage.py
```diff
@@ -8,60 +8,194 @@
...
+        Raises
+        ______
+        AssertionError
+            If `expression` is not found in `path`
+
+        See Also
+        --------
+        See `ExecutionUsage` for an example implementation
+        """
+        # TODO: mypy reports "None" not callable. Should we refactor?
```
thermokarst, line 106: Schwat?

qiime2/sdk/usage.py
andrewsanchez: Lolz, neato, eh?

>`Callable[..., ReturnType]` (literal ellipsis) can be used to type hint a callable taking any number of arguments and returning ReturnType.

https://docs.python.org/3.7/library/typing.html#typing.Callable

qiime2/sdk/usage.py
andrewsanchez: Fair enough!  Do you prefer this for both `types` and `typing`?

qiime2/sdk/usage.py
andrewsanchez: Sorry, that was not at all clear without additional context.

There's a guard in the `ScopeRecord` constructor which makes sure that `assert_has_line_matching` is either a callable or `None`.  But if `assert_has_line_matching` *was* `None`, the line below would break since `None` is not callable.  Which mypy doesn't like.

```python
return self._assert_has_line_matching_(self.ref, label, path, expression)
````

So my question is, does the mypy error indicate that a slight refactor is desirable?  How much do we care about what mypy thinks?

File: qiime2/sdk/usage.py (hunk 2 above)
Comment (thermokarst) [reply qiime2/sdk/usage.py:12]: Yes (the comment was on both lines, although it isn't very clear here - maybe GH only formats that specially when suggesting a code change?)

Comment (andrewsanchez) [reply qiime2/sdk/usage.py:12]: I saw that the comment was on both lines but just wanted to make sure :)

File: qiime2/sdk/usage.py (hunk 6 above)
Comment (thermokarst) [reply qiime2/sdk/usage.py:114]: Right now its fine, because if an interface chooses not to implement `assert_has_line_matching`, those statements in the usage examples are skipped.

File: qiime2/sdk/usage.py [hunk 7]
```diff
@@ -9,59 +9,224 @@
...
 from qiime2 import sdk, metadata
+from qiime2.core.type import MethodSignature, PipelineSignature
 
-# TODO: docstrings
+Factory = typing.Callable[..., typing.Union[metadata.Metadata, sdk.Artifact]]
+ExampleInputs = typing.Union[
+    str, int, typing.List[int], typing.Set[int], None, Factory
+]
+ActionData = typing.Union[sdk.Artifact, sdk.Visualization, metadata.Metadata]
+Signature = typing.Union[MethodSignature, PipelineSignature]
```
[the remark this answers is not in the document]
andrewsanchez: That's fine with me.  Let me know if you like it better now without the aliases.

qiime2/sdk/usage.py
```diff
@@ -9,59 +9,224 @@
...
+class ScopeRecord:
+    """
+    Tracks information needed by Usage drivers to render usage examples.
+    """
+
+    def __init__(
+            self, ref: str, value: ActionData, source: str,
+            assert_has_line_matching: typing.Optional[typing.Callable] = None
+    ):
+        """
```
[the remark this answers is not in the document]
andrewsanchez: Same with `Scope`, eh?

andrewsanchez:
@thermokarst, we still have issues, even with moving `usage` to `plugin`, and moving to `core` won't help either.

Are you familiar with [this PEP (563)](https://www.python.org/dev/peps/pep-0563/#abstract) that solves the problem of circular imports in the context of typing?  Seems like it is the solution to my problems, but only in 3.6+.  😕
```
  File "/Users/aas/Projects/qiime2/qiime2/qiime2/__init__.py", line 9, in <module>
    from qiime2.sdk import Artifact, Visualization
  File "/Users/aas/Projects/qiime2/qiime2/qiime2/sdk/__init__.py", line 11, in <module>
    from .plugin_manager import PluginManager
  File "/Users/aas/Projects/qiime2/qiime2/qiime2/sdk/plugin_manager.py", line 16, in <module>
    from qiime2.plugin.model import SingleFileDirectoryFormatBase
  File "/Users/aas/Projects/qiime2/qiime2/qiime2/plugin/__init__.py", line 17, in <module>
    from .usage import UsageAction, UsageInputs, UsageOutputNames
  File "/Users/aas/Projects/qiime2/qiime2/qiime2/plugin/usage.py", line 19, in <module>
    class ScopeRecord:
  File "/Users/aas/Projects/qiime2/qiime2/qiime2/plugin/usage.py", line 70, in ScopeRecord
    ) -> typing.Union[sdk.Artifact, sdk.Visualization, metadata.Metadata]:
AttributeError: module 'qiime2.sdk' has no attribute 'Artifact'
```

So if I'm understanding all of this correctly, when our type hints depend on accessing module attributes which are unavailable at "definition time," those hints will need to be string literals for now.  Does that sound right?

thermokarst:
Since you no longer need to worry about circular imports, can you please revert your changes that moved `usage.py` out of `sdk` and into `plugin`?

qiime2/sdk/usage.py
```diff
@@ -9,59 +9,236 @@
...
+    @property
+    def source(self) -> str:
+        """
+        Usage method called to initialize example data
+        """
+        return self._source
+
+    def assert_has_line_matching(
+        self, label: str, path: str, expression: str
+    ) -> None:
```
thermokarst, line 87:
```suggestion
    def assert_has_line_matching(self, label: str, path: str, expression: str) -> None:
```

qiime2/sdk/usage.py
```diff
@@ -9,59 +9,236 @@
...
+        """
+        Verify that the file at `path` contains a line matching `expression`
+        within an Artifact.
```
thermokarst, line 90:
```suggestion
        Verify that the file at `path` contains a line matching `expression`.
```

qiime2/sdk/usage.py
```diff
@@ -9,59 +9,236 @@
...
+
+        Parameters
+        ----------
+        label : str
+            A label for describing this assertion
```
thermokarst, line 95:
```suggestion
            A label for describing this assertion. Interface drivers may
            choose to omit this information in the rendered output.
```

qiime2/sdk/usage.py
```diff
@@ -9,59 +9,236 @@
...
+        --------
+        See `ExecutionUsage` for an example implementation.
+        """
+        return self._assert_has_line_matching_(self.ref, label, path,
+                                               expression)
+
+
+class Scope:
+    """
+    Track all ScopeRecords for a Usage example.
```
thermokarst, line 116:
```suggestion
    Track all ScopeRecords for a Usage example, sequentially.
```

qiime2/sdk/usage.py
```diff
@@ -9,59 +9,236 @@
...
-        if not isinstance(inputs, UsageInputs):
-            raise TypeError('Must provide an instance of UsageInputs.')
-        if not isinstance(outputs, UsageOutputNames):
-            raise TypeError('Must provide an instance of UsageOutputNames.')
+    def push_record(
+        self,
+        ref: str,
+        value: typing.Union[
+            "sdk.Artifact", "sdk.Visualization", "metadata.Metadata"
+        ],
```
thermokarst, line 142:
```suggestion
        value: typing.Union['Artifact', 'Visualization', 'Metadata'],
```

qiime2/sdk/usage.py
```diff
@@ -9,59 +9,236 @@
...
+        source: str,
+        assert_has_line_matching: typing.Callable = None,
+    ) -> ScopeRecord:
+        """
+        Update `self._records` with an entry for this record where `ref` is
+        the key and `ScopeRecord` is the value.
```
thermokarst, line 148:
```suggestion
        Appends a new `ScopeRecord`.
```

qiime2/sdk/usage.py
```diff
@@ -9,59 +9,236 @@
...
+        """
+        record = ScopeRecord(ref=ref, value=value, source=source,
+                             assert_has_line_matching=assert_has_line_matching)
+        self._records[ref] = record
+        return record
 
-        _, sig = self.get_action()
+    def get_record(self, ref: str) -> ScopeRecord:
+        """
+        Get a `ScopeRecord` from the current scope.
```
thermokarst, line 174:
```suggestion
        Look up a `ScopeRecord` from the current scope by identifier.
```

qiime2/sdk/usage.py
```diff
@@ -9,59 +9,236 @@
...
+        """
         self.values = kwargs
 
     def __repr__(self):
         return 'UsageInputs(**%r)' % (self.values,)
 
-    def validate(self, signature):
+    def validate(
+        self, signature: typing.Union[MethodSignature, PipelineSignature]
+    ) -> None:
```
thermokarst, line 226:
```suggestion
    def validate(self, signature: 'Signature') -> None:
```

qiime2/sdk/usage.py
```diff
@@ -81,19 +258,41 @@ def validate(self, signature):
...
-                             (extra, ))
-
-    def build_opts(self, signature, scope):
+                             (extra,))
+
+    def build_opts(
+        self,
+        signature: typing.Union[MethodSignature, PipelineSignature],
+        scope: Scope,
+    ) -> dict:
```
thermokarst, line 277:
```suggestion
    def build_opts(self, signature: 'Signature', scope: Scope) -> dict:
```

qiime2/sdk/usage.py
```diff
@@ -109,7 +308,17 @@ def build_opts(self, signature, scope):
 
 
 class UsageOutputNames:
-    def __init__(self, **kwargs):
+    """Define the outputs for a Usage example.
+    """
```
thermokarst, line 312:
```suggestion
```

qiime2/sdk/usage.py
```diff
@@ -119,38 +328,101 @@ def __init__(self, **kwargs):
...
+        -------
+        str
+            The name of an example output
+        """
         return self.values[key]
 
-    def validate(self, signature):
+    def validate(
+        self, signature: typing.Union[MethodSignature, PipelineSignature]
+    ) -> None:
```
thermokarst, line 345:
```suggestion
    def validate(self, signature: 'Signature') -> None:
```

qiime2/sdk/usage.py
```diff
@@ -119,38 +328,101 @@ def __init__(self, **kwargs):
...
+
+    def validate_computed(
+        self,
+        computed_outputs: typing.Dict[
+            str,
+            typing.Union[
+                "sdk.Artifact", "sdk.Visualization", "metadata.Metadata"
+            ],
+        ],
+    ) -> None:
```
thermokarst, line 379:
```suggestion
    def validate_computed(
        self,
        computed_outputs: typing.Dict[
            str, typing.Union['Artifact', 'Visualization', 'Metadata'],
        ]) -> None:
```

qiime2/sdk/usage.py
```diff
@@ -119,38 +328,101 @@ def __init__(self, **kwargs):
...
+            If there are missing or extra outputs as per the action signature
+        """
         provided = set(computed_outputs.keys())
         exp_outputs = set(self.values.keys())
 
         missing = exp_outputs - provided
         if len(missing) > 0:
             raise ValueError('SDK implementation is missing output(s): %r' %
-                             (missing, ))
+                             (missing,))
```
thermokarst, line 399:
```suggestion
                             (missing, ))
```

qiime2/sdk/usage.py
```diff
@@ -119,38 +328,101 @@ def __init__(self, **kwargs):
...
-                             'output(s): %r' % (extra, ))
-
-    def build_opts(self, action_signature, scope):
+                             'output(s): %r' % (extra,))
+
+    def build_opts(
+        self,
+        action_signature: typing.Union[MethodSignature, PipelineSignature],
+        scope: Scope,
+    ) -> dict:
```
thermokarst, line 410:
```suggestion
    def build_opts(self, action_signature: 'Signature', scope: Scope) -> dict:
```

qiime2/sdk/usage.py
```diff
@@ -159,80 +431,160 @@ def build_opts(self, action_signature, scope):
...
-class ScopeRecord:
-    def __init__(self, ref: str, value: object, source: str,
-                 assert_has_line_matching: callable = None):
-        if assert_has_line_matching is not None and \
-                not callable(assert_has_line_matching):
-            raise TypeError('Value for `assert_has_line_matching` should be a '
-                            '`callable`.')
+class UsageAction:
+    """Provide an action for a Usage example.
+    """
```
thermokarst, line 436:
```suggestion
```

qiime2/sdk/usage.py
```diff
@@ -159,80 +431,160 @@ def build_opts(self, action_signature, scope):
...
+    def __repr__(self):
+        return 'UsageAction(plugin_id=%r, action_id=%r)' % \
+               (self.plugin_id, self.action_id)
+
+    def get_action(
+        self,
+    ) -> typing.Tuple[
+        typing.Union[sdk.Method, sdk.Pipeline],
+        typing.Union[MethodSignature, PipelineSignature],
+    ]:
```
thermokarst, line 469:
```suggestion
    def get_action(self) -> typing.Tuple['Action', 'Signature']:
```

qiime2/sdk/usage.py
```diff
@@ -159,80 +431,160 @@ def build_opts(self, action_signature, scope):
...
-            raise KeyError('No record with ref id: "%s" in scope.' % (ref, ))
+        _, sig = self.get_action()
+
+        inputs.validate(sig)
+        outputs.validate(sig)
 
 
 class Usage(metaclass=abc.ABCMeta):
+    """`Usage` is the base class for Usage driver implementations.
+    """
```
thermokarst, line 514:
```suggestion
    """
    `Usage` is the base class for Usage driver implementations.
    """
```

qiime2/sdk/usage.py
```diff
@@ -159,80 +431,160 @@ def build_opts(self, action_signature, scope):
...
+
     def __init__(self):
         self._scope = Scope()
 
-    def init_data(self, ref, factory):
+    def init_data(
+        self, ref: str, factory: typing.Callable[[], "sdk.Artifact"]
+    ) -> ScopeRecord:
```
thermokarst, line 521:
```suggestion
    def init_data(self, ref: str,
                  factory: typing.Callable[[], 'Artifact']) -> ScopeRecord:
```

qiime2/sdk/usage.py
```diff
@@ -159,80 +431,160 @@ def build_opts(self, action_signature, scope):
...
+        """Initialize example data from a factory.
```
thermokarst, line 522:
```suggestion
        """
        Initialize example data from a factory.
```

qiime2/sdk/usage.py
```diff
@@ -159,80 +431,160 @@ def build_opts(self, action_signature, scope):
...
         return self._push_record(ref, value, 'init_data')
 
     def _init_data_(self, ref, factory):
         raise NotImplementedError
 
-    def init_metadata(self, ref, factory):
+    def init_metadata(
+        self, ref: str, factory: typing.Callable[[], "metadata.Metadata"]
+    ) -> ScopeRecord:
+        """Initialize metadata for a Usage example.
```
thermokarst, line 545:
```suggestion
    def init_metadata(self, ref: str,
                      factory: typing.Callable[[], 'Metadata']) -> ScopeRecord:
        """
        Initialize metadata for a Usage example.
```

qiime2/sdk/usage.py
```diff
@@ -159,80 +431,160 @@ def build_opts(self, action_signature, scope):
...
         raise NotImplementedError
 
-    def init_data_collection(self, ref, collection_type, *records):
+    def init_data_collection(
+        self,
+        ref: str,
+        collection_type: typing.Union[list, set],
+        *records: ScopeRecord
+    ) -> ScopeRecord:
+        """Initialize a collection of data for a Usage example.
```
thermokarst, line 571:
```suggestion
    def init_data_collection(self, ref: str,
                             collection_type: typing.Union[list, set],
                             *records: typing.List[ScopeRecord]) \
                             -> ScopeRecord:
        """
        Initialize a collection of data for a Usage example.
```

qiime2/sdk/usage.py
```diff
@@ -243,10 +595,26 @@ def init_data_collection(self, ref, collection_type, *records):
         value = self._init_data_collection_(ref, collection_type, *records)
         return self._push_record(ref, value, 'init_data_collection')
 
-    def _init_data_collection_(self, ref, collection_type, *records):
+    def _init_data_collection_(self, ref, collection_type, records):
         raise NotImplementedError
 
-    def merge_metadata(self, ref, *records):
+    def merge_metadata(self, ref: str, *records: ScopeRecord) -> ScopeRecord:
```
thermokarst, line 601:
```suggestion
    def merge_metadata(self, ref: str,
                       *records: typing.List[ScopeRecord]) -> ScopeRecord:
```

qiime2/sdk/usage.py
```diff
@@ -256,7 +624,24 @@ def merge_metadata(self, ref, *records):
     def _merge_metadata_(self, ref, records):
         raise NotImplementedError
 
-    def get_metadata_column(self, column_name, record):
+    def get_metadata_column(
+        self, column_name: str, record: ScopeRecord
+    ) -> ScopeRecord:
```
thermokarst, line 629:
```suggestion
    def get_metadata_column(self, column_name: str, record: ScopeRecord) \
            -> ScopeRecord:
```

qiime2/sdk/usage.py
```diff
@@ -269,8 +654,30 @@ def comment(self, text: str):
...
         raise NotImplementedError
 
-    def action(self, action: UsageAction, inputs: UsageInputs,
-               outputs: UsageOutputNames):
+    def action(
+        self,
+        action: UsageAction,
+        inputs: UsageInputs,
+        outputs: UsageOutputNames,
+    ) -> None:
```
thermokarst, line 662:
```suggestion
    def action(self, action: UsageAction, inputs: UsageInputs,
               outputs: UsageOutputNames) -> None:
```

qiime2/sdk/usage.py
```diff
@@ -269,8 +654,30 @@ def comment(self, text: str):
...
+        """
+        This is the primary entry point for the Usage API.  This method is
+        where Usage example developers pass in information necessary for
+        drivers to render Usage examples.
```
thermokarst, line 666:
```suggestion
        This method invokes (or simulates invokation) of a QIIME 2 Action.
```

qiime2/sdk/usage.py
```diff
@@ -284,30 +691,50 @@ def action(self, action: UsageAction, inputs: UsageInputs,
         computed_outputs = self._action_(action, input_opts, output_opts)
         self._add_outputs_to_scope(outputs, computed_outputs)
 
-    def _action_(self, action: UsageAction,
-                 input_opts: dict, output_opts: dict):
+    def _action_(
+        self, action: UsageAction, input_opts: dict, output_opts: dict
+    ) -> dict:
```
thermokarst, line 696:
```suggestion
    def _action_(self, action: UsageAction, input_opts: dict,
                 output_opts: dict) -> dict:
```

qiime2/sdk/usage.py
```diff
@@ -284,30 +691,50 @@ def action(self, action: UsageAction, inputs: UsageInputs,
...
+        method used to access records generated after running `Usage.action`.
+
+        Parameters
+        ----------
+        ref : str
+            Output name
+
+        Raises
+        ------
+            If `ref` is not associated with a record generated by
```
thermokarst, line 714: Missing the exception type (probably a `KeyError`)

qiime2/sdk/usage.py
```diff
@@ -284,30 +691,50 @@ def action(self, action: UsageAction, inputs: UsageInputs,
...
             raise TypeError('source == %s but must be "action"' % source)
         return record
 
-    def _add_outputs_to_scope(self, outputs, computed_outputs):
+    def _add_outputs_to_scope(
+        self, outputs: UsageOutputNames, computed_outputs
+    ):
         outputs.validate_computed(computed_outputs)
-        for output, result in computed_outputs.items():
+        for output, value in computed_outputs.items():
```
thermokarst, line 727:
```suggestion
        for output, result in computed_outputs.items():
```

`Result` is the base-class for `Artifact` and `Visualization`, so `result` here is deliberate, to indicate that this is some kind of `Result` object.

qiime2/sdk/usage.py
```diff
@@ -284,30 +691,50 @@ def action(self, action: UsageAction, inputs: UsageInputs,
...
             ref = outputs.get(output)
-            self._push_record(ref, result, 'action')
+            self._push_record(ref, value, 'action')
```
thermokarst, line 729:
```suggestion
            self._push_record(ref, result, 'action')
```

qiime2/sdk/usage.py
```diff
@@ -284,30 +691,50 @@ def action(self, action: UsageAction, inputs: UsageInputs,
...
         record = self._get_record(ref)
         source = record.source
         if source != 'action':
             raise TypeError('source == %s but must be "action"' % source)
         return record
 
-    def _add_outputs_to_scope(self, outputs, computed_outputs):
+    def _add_outputs_to_scope(
+        self, outputs: UsageOutputNames, computed_outputs
+    ):
```
thermokarst, line 725:
```suggestion
    def _add_outputs_to_scope(self, outputs: UsageOutputNames,
                              computed_outputs):
```

qiime2/sdk/usage.py
```diff
@@ -317,6 +744,12 @@ def _get_records(self):
 
 
 class DiagnosticUsage(Usage):
+    """Generate information for testing the Usage API.
```
thermokarst, line 747:
```suggestion
    """
    Reference implementation of a Usage driver. Used to generate information
    for testing the Usage API.
```

qiime2/sdk/usage.py
```diff
@@ -385,6 +818,13 @@ def _assert_has_line_matching_(self, ref, label, path, expression):
 
 
 class ExecutionUsage(Usage):
+    """Execute and test rendered examples.
```
thermokarst, line 821:
```suggestion
    """
    Execute and test rendered examples.
```

qiime2/sdk/usage.py
```diff
@@ -428,8 +868,9 @@ def _get_metadata_column_(self, column_name, record):
     def _comment_(self, text):
         pass
 
-    def _action_(self, action: UsageAction,
-                 input_opts: dict, output_opts: dict):
+    def _action_(
+        self, action: UsageAction, input_opts: dict, output_opts: dict
+    ):
```
thermokarst, line 873:
```suggestion
    def _action_(self, action: UsageAction,
                 input_opts: dict, output_opts: dict):
```

thermokarst (changes requested):
Looks great! This is going to be so useful to have - thanks! I have included a bunch of in-line suggestions. A few of those suggestions I only called out once or twice, and have asked you to find-and-replace elsewhere in the diff. As well, here are a few broader suggestions:

-  Please replace `"` with `'` - to keep the prevailing style intact.
- Please ensure that all statements are converted to sentences (ending with periods)

A minor quibble - the diff here was pretty painful to review because the _order_ of the code was shuffled in addition to the content. In the future it would be helpful to stage the commits in a logical order to assist with review.

Thanks! Let me know if you have any questions, comments, or concerns!

qiime2/sdk/usage.py
andrewsanchez: Assuming that we will eventually run something like mypy, wouldn't it be better to use actual types in these type hints, even though they are just strings atm?  If there is no actual `Signature` class or an alias, might this type hint potentially cause confusion?  Or is there a `Signature` object?  I couldn't find one though I see it (erroneously?) mentioned [here](https://github.com/qiime2/qiime2/blob/e3def226e64340d9a474f283863edb8fad57cd7f/qiime2/sdk/action.py#L82).

qiime2/sdk/usage.py
andrewsanchez: Is there a practical reason to prefer this suggestion over my original commit?  I don't see the preference recommended anywhere, including PEP 257, https://numpydoc.readthedocs.io/en/latest/format.html.  I also see both variations in the framework source code as well as NumPy source code.  I don't really care, I'm just curious!  😄 .  I think I actually prefer your suggestion.  I was just mimicking the examples in PEP 257 (which does mention either variation is fine).  Just curious if there is a deeper practical reason I'm unaware of.

qiime2/sdk/usage.py
andrewsanchez: Good catch!  We went with `TypeError` in #536, here and in other places as well.  You good with sticking with that?

qiime2/sdk/usage.py
andrewsanchez: I think `*records: ScopeRecord` is the correct type hint here (see https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values).

andrewsanchez:
I'm surprised to learn that string literal type hints [confused flake8](https://travis-ci.com/github/qiime2/qiime2/builds/196822971#L1288):  `./qiime2/sdk/usage.py:31:29: F821 undefined name 'Artifact'`.  We will need to import these objects to satisfy flake8 or decide on another solution.

qiime2/sdk/usage.py
thermokarst: Good catch, that should be `signature: 'PipelineSignature'` on line 198 above, since `PipelineSignature` is the base class for all the signature classes.

qiime2/sdk/usage.py
thermokarst: > Is there a practical reason to prefer this suggestion over my original commit?

Nope, you can ignore it - it was part of a broad set of linting rules I applied locally. I have no preference, but personally I like the `"""\nfoo\n"""` style, I think the `"""` delimits the comments well.

> Just curious if there is a deeper practical reason I'm unaware of.

I think the only other thing to keep in mind with these kinds of decisions is project/repository consistency. I quick spin around this repo shows examples of both docstring styles, so no real pattern here, but something to keep in mind, in general.

qiime2/sdk/usage.py
thermokarst: Thanks, I agree with your finding!

qiime2/sdk/usage.py
thermokarst: This method raises both a `KeyError` (by way of `_get_record`) and a `TypeError`, depending on nature of the failure, you will need to account for both in the docs.

thermokarst:
> We will need to import these objects to satisfy flake8 or decide on another solution.

You have already imported most of these objects, you'll need to use the namespace in the annotations:

Instead of `'Artifact', 'Visualization', 'Metadata'` you'll need to use `'sdk.Artifact', 'sdk.Visualization', 'sdk.Metadata'`.

I'll leave the import and namespace for the `PipelineSignature` objects as an exercise to you, but as a hint, it _can_ be done (I just verified locally).

qiime2/sdk/usage.py
andrewsanchez: >Nope, you can ignore it - it was part of a broad set of linting rules I applied locally.

Out of curiosity, do you apply those programmatically somehow (YAPF, IDE functionality, etc.)?

andrewsanchez:
> A minor quibble - the diff here was pretty painful to review because the _order_ of the code was shuffled in addition to the content. In the future it would be helpful to stage the commits in a logical order to assist with review.

I raised this concern (it feels like _ages ago_ at this point!) during a call when I realized this changeset would require shuffling the order. I thought I remember you telling me that I shouldn't worry about it because you thought GitHub's UI would display the diff intelligently.  But maybe I misinterpreted your response.  Sorry about that!

The latest changes address your suggestions as well as what was reported by `python -m numpydoc`.

qiime2/sdk/usage.py
```diff
@@ -28,10 +50,24 @@ def __init__(self, *, plugin_id: str, action_id: str):
...
 
     def __repr__(self):
-        return 'UsageAction(plugin_id=%r, action_id=%r)' %\
-            (self.plugin_id, self.action_id)
+        return 'UsageAction(plugin_id=%r, action_id=%r)' % \
+               (self.plugin_id, self.action_id)
+
+    def get_action(
+        self,
+    ) -> typing.Tuple[typing.Union['sdk.Method', 'sdk.Pipeline'],
```
thermokarst, line 58: This is minor, and is filed under "stylistic concerns," but splitting `self` out onto its own line seems unnecessary to me and just adds noise to the PR diff. Changing the prevailing style of pre-existing code just seems like extra work. Take that comment with a grain of salt, though: I like to play diff-golf.

qiime2/sdk/usage.py
```diff
@@ -6,16 +6,38 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
+"""
+The Usage module enables generating examples for multiple QIIME 2 interfaces.
```
thermokarst, line 10: When you put this back in the dev-docs please drop this sentence.

qiime2/sdk/usage.py
```diff
@@ -6,16 +6,38 @@
...
+
+The Usage API provides an interface-agnostic way for QIIME 2 plugin developers
+to define examples of how to use their plugin’s actions. This enables the
+programmatic generation of examples for all QIIME 2 interfaces, eliminating the
+need to maintain specific examples for multiple interfaces.
+
+Examples
+--------
+Please see `qiime2.core.testing.examples` for examples.
```
thermokarst, line 19: Putting a note here, but please update _everywhere_ in this document:

These docstrings are restructuredtext, not markdown, so you'll need to tweak the backticks to work as expected: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

qiime2/sdk/usage.py
```diff
@@ -269,8 +645,28 @@ def comment(self, text: str):
...
-    def action(self, action: UsageAction, inputs: UsageInputs,
-               outputs: UsageOutputNames):
+    def action(
+        self,
+        action: UsageAction,
+        inputs: UsageInputs,
+        outputs: UsageOutputNames,
+    ) -> None:
+        """
+        This method invokes (or simulates invokation) of a QIIME 2 Action.
```
thermokarst, line 655:
```suggestion
        This method is a proxy for invoking a QIIME 2 Action. Whether or not the Action is actually invoked is dependent on the driver executing the example. 
```

qiime2/sdk/usage.py
```diff
@@ -269,8 +645,28 @@ def comment(self, text: str):
...
+
+        Parameters
+        ----------
+        action : UsageAction
+            Example action.
+        inputs : UsageInputs
+            Example inputs.
+        outputs : UsageOutputNames
+            Example outputs.
```
thermokarst, line 664: I think we should make these a bit more descriptive.

qiime2/sdk/usage.py
```diff
@@ -178,61 +370,211 @@ def __repr__(self):
...
+
     def __init__(self):
         self._scope = Scope()
 
-    def init_data(self, ref, factory):
+    def init_data(
+        self, ref: str, factory: typing.Callable[[], 'sdk.Artifact']
+    ) -> 'ScopeRecord':
+        """
+        Initialize example data from a factory.
```
thermokarst, line 510:
```suggestion
        Initialize example data from a factory. Whether or not the example data is actually created is dependent on the driver executing the example. 
```

qiime2/sdk/usage.py
```diff
@@ -178,61 +370,211 @@ def __repr__(self):
...
 
     def _init_data_(self, ref, factory):
         raise NotImplementedError
 
-    def init_metadata(self, ref, factory):
+    def init_metadata(
+        self, ref: str, factory: typing.Callable[[], 'metadata.Metadata']
+    ) -> 'ScopeRecord':
+        """
+        Initialize metadata for a Usage example.
```
thermokarst, line 534:
```suggestion
        Initialize metadata for a Usage example. Whether or not the example metadata is actually created is dependent on the driver executing the example. 
```

qiime2/sdk/usage.py
```diff
@@ -178,61 +370,211 @@ def __repr__(self):
...
+        collection_type : list or set
+            The type of collection required by an action.
+        records : ScopeRecords belonging to the collection
+            The record associated with data to be initialized in the
+            collection.
+
+        Returns
+        -------
+        record : ScopeRecord
+            A record with information about the example Metadata.
```
thermokarst, line 576: This looks like a copy-and-paste error.

qiime2/sdk/usage.py
```diff
@@ -178,61 +370,211 @@ def __repr__(self):
...
+        Initialize a collection of data for a Usage example.
+
+        Parameters
+        ----------
+        ref : str
+            Unique identifier for example data collection.
+        collection_type : list or set
+            The type of collection required by an action.
+        records : ScopeRecords belonging to the collection
+            The record associated with data to be initialized in the
```
thermokarst, line 570: record -> records

qiime2/sdk/usage.py
```diff
@@ -243,10 +585,27 @@ def init_data_collection(self, ref, collection_type, *records):
...
 
-    def _init_data_collection_(self, ref, collection_type, *records):
+    def _init_data_collection_(self, ref, collection_type, records):
         raise NotImplementedError
 
-    def merge_metadata(self, ref, *records):
+    def merge_metadata(self, ref: str,
+                       *records: typing.List['ScopeRecord']) -> 'ScopeRecord':
+        """
+        Create a `ScopeRecord` for merged Metadata.
```
thermokarst, line 594: Technically all of these plugin-dev-facing methods create `ScopeRecord`s. I prefer the language you have adopted above, omitting the mention of the `ScopeRecord`.

qiime2/sdk/usage.py
```diff
@@ -256,7 +615,24 @@ def merge_metadata(self, ref, *records):
     def _merge_metadata_(self, ref, records):
         raise NotImplementedError
 
-    def get_metadata_column(self, column_name, record):
+    def get_metadata_column(
+        self, column_name: str, record: ScopeRecord
+    ) -> 'ScopeRecord':
+        """
+        Create a new record for a specific column in example Metadata.
```
thermokarst, line 622: Same comment as above - please adopt language consistent with `init_data` et al.

qiime2/sdk/usage.py
thermokarst: This is still missing the `KeyError` I mentioned in this thread. Please see https://github.com/qiime2/qiime2/blob/3953ce918683e7af7012a8d8570ca73f3a486970/qiime2/sdk/usage.py#L214 for more details.

thermokarst (changes requested):

qiime2/sdk/usage.py
thermokarst: I forgot to mention, but there are a ton of other examples of this sprinkled throughout this PR, I'm not too sure if that is really a necessary change to the module.

qiime2/sdk/usage.py
```diff
@@ -6,16 +6,38 @@
...
+
+The Usage API provides an interface-agnostic way for QIIME 2 plugin developers
+to define examples of how to use their plugin’s actions. This enables the
+programmatic generation of examples for all QIIME 2 interfaces, eliminating the
+need to maintain specific examples for multiple interfaces.
+
+Examples
+--------
+Please see `qiime2.core.testing.examples` for examples.
+"""
```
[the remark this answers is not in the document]
andrewsanchez: This docstring still ends up in the rendered dev docs because of the `.. automodule:: qiime2.sdk.usage` directive.

Moving it was prompted by running `numpydoc --validate` which reminded me about [numpy's advice for documenting modules](https://numpydoc.readthedocs.io/en/latest/format.html#documenting-modules).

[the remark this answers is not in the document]
thermokarst: > This docstring still ends up in the rendered dev docs because of the .. automodule:: qiime2.sdk.usage directive.

Yep, I saw that change. My point is that this prose here is more appropriate as high-level plugin-developer-facing documentation, and less appropriate as a general-purpose module-level docstring.

I'm not saying we can't have a module-level docstring, I'm just saying that _this_ docstring is better for plugin-dev docs. As the docs grow, the differences between the plugin-dev docs vs the sdk docs will become more cumbersome, so let's split it from the outset.

qiime2/sdk/usage.py
```diff
@@ -6,16 +6,38 @@
...
+    """
+    Parameters
+    ----------
+    plugin_id : str
+        Plugin ID.
+    action_id : str
+        Action ID.
+    """
+
+    # TODO If *arg here is necessary, create an example
```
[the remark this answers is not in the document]
andrewsanchez: I wasn't sure what the type annotation for `*` should be in the constructor here since I couldn't find any examples of it being used.  I only see `UsageAction` being instantiated with the keyword arguments.  I have a feeling like this is a stupid question, but why is `*` a part of the signature for `UsageAction`'s constructor? Also, why isn't it `*args` instead?

qiime2/sdk/usage.py
andrewsanchez: I've tried to follow the numpydoc style guide.  Please see the note [here]( https://numpydoc.readthedocs.io/en/latest/format.html#common-rest-concepts):

> Variable, module, function, and class names should be written between single back-ticks (\`numpy\`).

thermokarst: Fair enough, but have you looked at the rendered dev docs? The single-backticks don't render correctly.

Screen Shot 2020-11-09 at 3 44 15 PM

You'll notice that other modules in qiime2 also use the double-backtick syntax, so I don't think you'll be able to change a config option on the dev-docs (at least not without breaking everything else). For an example, take a look at [`qiime2.metadata`](https://github.com/qiime2/qiime2/tree/master/qiime2/metadata).

qiime2/sdk/usage.py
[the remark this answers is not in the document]
andrewsanchez: That makes sense. Thanks for clarifying.

qiime2/sdk/usage.py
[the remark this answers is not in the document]
thermokarst: Great question! Check out [PEP3102](https://www.python.org/dev/peps/pep-3102/) for details, also:
https://python-3-for-scientists.readthedocs.io/en/latest/python3_advanced.html#keyword-only-arguments

andrewsanchez:
Not sure why I couldn't reply directly to your comment above, but after reviewing [numpydoc Sections](https://numpydoc.readthedocs.io/en/latest/format.html#sections) again, what do you think about dropping the Raises section here entirely?  Or keeping just the `TypeError`?

>This section should be used judiciously, i.e., only for errors that are non-obvious or have a large chance of getting raised.

thermokarst:
> Not sure why I couldn't reply directly to your comment above

Because the comment is on a _previous_ review - scroll up and unhide the collapse comments to find it. Alternatively you can reply in the "Files changed" tab.

> what do you think about dropping the Raises section here entirely? Or keeping just the `TypeError`?

Why do you think the `TypeError` should be kept over the `KeyError`?  I think following the guidance you quoted above, both types of errors "have a large chance of getting raised."

qiime2/sdk/usage.py
andrewsanchez: Is 4e326c5 better?  I think my languaging faculties are failing me at the end of the day.

qiime2/sdk/usage.py
andrewsanchez: I think I just mistakenly called it a `ValueError`. 

>Why do you think the TypeError should be kept over the KeyError?

The `KeyError` seems more obvious to me.  My (unjustified) interpretation of "errors that are non-obvious or have a large chance of getting raised," was that, if it's obvious, leave it out.

qiime2/sdk/usage.py
```diff
@@ -317,6 +728,15 @@ def _get_records(self):
...
 
 class DiagnosticUsage(Usage):
+    """
+    Reference implementation of a Usage driver.
+
+    Used to generate information for testing the Usage API.
+
+    See Also
+    --------
+    qiime2.plugin.tests.TestUsage : Unit tests using this driver.
```
thermokarst, line 738: Looks like `TestUsage` isn't a suite in this module (although that probably should be its name, can you rename in this PR?).

qiime2/sdk/usage.py
```diff
@@ -385,6 +805,14 @@ def _assert_has_line_matching_(self, ref, label, path, expression):
 
 
 class ExecutionUsage(Usage):
+    """
+    Execute and test rendered examples.
+
+    See Also
+    --------
+    qiime2.plugin.tests.TestUsage : Unit tests using this driver.
```
thermokarst, line 813: I don't see a suite by this name at this path - can you either update the name here, or fix the name in the tests?

thermokarst (changes requested):

qiime2/sdk/usage.py
```diff
@@ -243,10 +580,27 @@ def init_data_collection(self, ref, collection_type, *records):
         value = self._init_data_collection_(ref, collection_type, *records)
```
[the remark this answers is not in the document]
andrewsanchez: The `*` here is intentional:

```
ERROR: test_use_init_collection_data (qiime2.tests.test_artifact_api.TestArtifactAPIUsage)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/admin/Projects/q2/qiime2/qiime2/tests/test_artifact_api.py", line 247, in test_use_init_collection_data
    action.examples['variadic_input_simple'](use)
  File "/Users/admin/Projects/q2/qiime2/qiime2/core/testing/examples.py", line 177, in variadic_input_simple
    ints = use.init_data_collection('ints', list, ints_a, ints_b)
  File "/Users/admin/Projects/q2/qiime2/qiime2/sdk/usage.py", line 585, in init_data_collection
    value = self._init_data_collection_(ref, collection_type, records)
  File "/Users/admin/Projects/q2/qiime2/qiime2/plugins.py", line 41, in _init_data_collection_
    t = ', '.join(sorted([i.ref for i in records]))
  File "/Users/admin/Projects/q2/qiime2/qiime2/plugins.py", line 41, in <listcomp>
    t = ', '.join(sorted([i.ref for i in records]))
AttributeError: 'tuple' object has no attribute 'ref'
```

 IIRC, you didn't want the `*` in the abstract method signature, and I removed it in 17cca13.  I can't remember why...  🤔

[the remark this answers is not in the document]
thermokarst: I'll leave the rest of the exercise to you, but ping me if you need help. This * should be removed, because it was an oversight on the initial draft of the api.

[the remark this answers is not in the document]
andrewsanchez: I think I squarshed it 🐛 .

thermokarst (approved):

PR #560, opened 2021-02-03
API: usage, inject the action helpers, too

thermokarst:
This is a breaking API change. I think it will be helpful for a few
reasons:

1) Importing the UsageAction/Inputs/OutputNames from qiime2.plugin was
   weird and a little confusing, IMO.
2) This feels more consistent to me with other elements of Usage
3) This will simplify writing sphinx-ext-qiime2's usage extensions

---

I am open to other ideas, and also am interested in discussing names.

cc @andrewsanchez && @ebolyen

### DIFF
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -10,8 +10,6 @@ import pandas as pd
 
 from qiime2 import Artifact, Metadata
 
-from qiime2.plugin import UsageAction, UsageInputs, UsageOutputNames
-
 from .type import IntSequence1, IntSequence2, Mapping, SingleInt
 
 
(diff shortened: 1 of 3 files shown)


andrewsanchez:
This sounds good to me.  The only thought I have to contribute after a first pass is maybe a preference for `use.get_action` over `use.lookup_action`.

ebolyen:
I see how this would be annoying when writing a tutorial, what about `use.OriginalNames` and we keep the old imports available?

I do think there is some value in having these objects obviously be objects.

ebolyen:
Actually maybe getting rid of the original imports is better, since it is possible to override these now that they are on the usage driver.

So if for some reason you had a weird situation where you did:

```python
if isinstance(arg, UsageInput):
    ...
```

You would really want it to be `use.UsageInput` on the off-chance it had been overridden (in this situation, it would probably be the very same code doing so, but still)

thermokarst:
> since it is possible to override these now that they are on the usage driver.

This sounds like a pretty useful thing to have, as well!

RE the original names - I'm open to that, it just sounded awkward with `use.UsageAction`, `use.UsageInputs`, and `use.UsageOutputNames`. We could chop `Usage` off:

- `use.Action`
- `use.Inputs`
- `use.OutputNames`,

but, then `use.Action` is a little too similar to `use.action`, which does something else entirely. 🤯

ebolyen:
Definitely, what about something like `DeclareAction`, `DeclareInputs`, `DeclareOutputNames`? Or something in that vein.

Alternatively, I am not too concerned about the redundancy, since the Python stdlib is full of it and few complain. In my mind `use` is closer to how I would treat a module object, so I just see it as a namespace and don't mind too much.

thermokarst:
I'm good with the redundancy if y'all are, too. @andrewsanchez - thoughts on `UsageAction` vs `DeclareAction`?

andrewsanchez:
I like the idea of using the original names.  But I don't feel too strongly about it, one way or the other.  I think `use.Action` _and_ `use.action` would be unfortunate.

thermokarst:
Thanks! Let's just keep it with the original names, then.

andrewsanchez (approved):
👌 🧠

PR #566, opened 2021-03-12
bug: networked filesystem race conditions

thermokarst:
This changeset attempts to address some longstanding issues we've seen crop up on the forum from time to time:

- https://forum.qiime2.org/t/qiime-tools-import-on-cluster/17531
- https://forum.qiime2.org/t/error-while-runing-dada2/16600
- https://forum.qiime2.org/t/qiime2-installation-and-run-error/4766
- https://forum.qiime2.org/t/out-of-disk-space-error/5736

This has been a tricky one to debug though, since it seems to be specifically tied to the [BeeGFS filesystem](https://en.wikipedia.org/wiki/BeeGFS). I finally figured out how to reproduce this error though, using the following repo (plus patch; I also had to manually rebuild the beegfs client and launch the system service):

https://github.com/jeremysj/beegfs-sandbox/

```patch
diff --git a/roles/beegfs_client/tasks/main.yml b/roles/beegfs_client/tasks/main.yml
index 5ea86fa..6ed0ec9 100644
--- a/roles/beegfs_client/tasks/main.yml
+++ b/roles/beegfs_client/tasks/main.yml
@@ -6,7 +6,8 @@
 - name: "BeeGFS client packages"
   yum: name="{{ item }}" state=present
   with_items:
-  - "kernel-devel-{{ ansible_kernel }}"
+  - kernel
+  - kernel-devel
   - gcc
   - beegfs-client
   - beegfs-helperd

```

So, after poking around a bit, I can't find anything obviously wrong with the framework's handling of these temp paths - no double-creation of dirs/files, etc. It really only seems to be an issue when setting your machine's `TMPDIR` to a beegfs mount. One thing I learned while poking around at things is that [`pathlib.Path.rename`](https://docs.python.org/3.6/library/pathlib.html#pathlib.Path.rename) is a thin wrapper around [`os.rename`](https://docs.python.org/3.6/library/os.html#os.rename). Check out this `os.rename` note in the docs:

> The operation may fail on some Unix flavors if src and dst are on different filesystems.

I suppose we have been warned. Anyway, other solutions I looked at included replacing `rename` with `replace`, but I saw identical failures.

I haven't written any tests for this yet, I wanted to put this out there into the world (cc @ebolyen) to see what folks think. This changeset takes the approach of attempting to perform the lower-cost rename operations, and if it fails due to this `FileExistsError`, falls back to a slightly more expensive copy operation. I am open to other suggestions, just let me know. PS - thought about trying to use `qiime2.util.duplicate` in here, but steered clear for now.

### DIFF
--- a/qiime2/core/archive/provenance.py
+++ b/qiime2/core/archive/provenance.py
@@ -317,7 +317,12 @@ class ProvenanceCapture:
         self.write_action_yaml()
         self.write_citations_bib()
 
-        self.path.rename(final_path)
+        # Certain networked filesystems will experience a race
+        # condition on `rename`, so fall back to copying.
+        try:
+            self.path.rename(final_path)
+        except FileExistsError:
+            distutils.dir_util.copy_tree(str(self.path), str(final_path))
 
     def fork(self):
         forked = copy.copy(self)
--- a/qiime2/core/archive/provenance.py
+++ b/qiime2/core/archive/provenance.py
@@ -324,8 +324,6 @@ class ProvenanceCapture:
         except FileExistsError:
             distutils.dir_util.copy_tree(str(self.path), str(final_path))
             distutils.dir_util.remove_tree(str(self.path))
-            # TODO: update self.path? This doesn't seem to work as hoped
-            # self.path = qiime2.core.path.ProvenancePath(str(final_path))
 
     def fork(self):
         forked = copy.copy(self)
--- a/qiime2/core/archive/provenance.py
+++ b/qiime2/core/archive/provenance.py
@@ -323,6 +323,9 @@ class ProvenanceCapture:
             self.path.rename(final_path)
         except FileExistsError:
             distutils.dir_util.copy_tree(str(self.path), str(final_path))
+            distutils.dir_util.remove_tree(str(self.path))
+            # TODO: update self.path? This doesn't seem to work as hoped
+            # self.path = qiime2.core.path.ProvenancePath(str(final_path))
 
     def fork(self):
         forked = copy.copy(self)
--- a/qiime2/core/archive/tests/test_provenance.py
+++ b/qiime2/core/archive/tests/test_provenance.py
@@ -262,6 +262,20 @@ class TestProvenanceIntegration(unittest.TestCase):
         self.assertEqual(obs, exp)
         self.assertTrue(mocked_tzlocal.called)
 
+    def test_prov_rename(self):
+        viz, = dummy_plugin.actions.no_input_viz()
+
+        viz_p_dir = viz._archiver.provenance_dir
+        self.assertTrue(viz_p_dir.exists())
+
+    @mock.patch('qiime2.core.path.ProvenancePath.rename',
+                side_effect=FileExistsError)
+    def test_prov_rename_file_exists(self, _):
+        viz, = dummy_plugin.actions.no_input_viz()
+
+        viz_p_dir = viz._archiver.provenance_dir
+        self.assertTrue(viz_p_dir.exists())
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/qiime2/core/archive/tests/test_provenance.py
+++ b/qiime2/core/archive/tests/test_provenance.py
@@ -274,7 +274,9 @@ class TestProvenanceIntegration(unittest.TestCase):
         viz, = dummy_plugin.actions.no_input_viz()
 
         viz_p_dir = viz._archiver.provenance_dir
-        self.assertTrue(viz_p_dir.exists())
+
+        with (viz_p_dir / 'action' / 'action.yaml').open() as fh:
+            self.assertIn('output-name: visualization', fh.read())
 
 
 if __name__ == '__main__':
--- a/qiime2/core/path.py
+++ b/qiime2/core/path.py
@@ -33,6 +33,12 @@ class OwnedPath(_ConcretePath):
         else:
             return shutil.copy(str(self), str(other))
 
+    def _destruct(self):
+        if self.is_dir():
+            distutils.dir_util.remove_tree(str(self))
+        else:
+            self.unlink()
+
     def _move_or_copy(self, other):
         if self._user_owned:
             return self._copy_dir_or_file(other)
@@ -42,7 +48,9 @@ class OwnedPath(_ConcretePath):
             try:
                 return _ConcretePath.rename(self, other)
             except FileExistsError:
-                return self._copy_dir_or_file(other)
+                copied = self._copy_dir_or_file(other)
+                self._destruct()
+                return copied
 
 
 class InPath(OwnedPath):
--- a/qiime2/core/path.py
+++ b/qiime2/core/path.py
@@ -27,14 +27,22 @@ class OwnedPath(_ConcretePath):
         self._user_owned = True
         return self
 
+    def _copy_dir_or_file(self, other):
+        if self.is_dir():
+            return distutils.dir_util.copy_tree(str(self), str(other))
+        else:
+            return shutil.copy(str(self), str(other))
+
     def _move_or_copy(self, other):
         if self._user_owned:
-            if self.is_dir():
-                return distutils.dir_util.copy_tree(str(self), str(other))
-            else:
-                return shutil.copy(str(self), str(other))
+            return self._copy_dir_or_file(other)
         else:
-            return _ConcretePath.rename(self, other)
+            # Certain networked filesystems will experience a race
+            # condition on `rename`, so fall back to copying.
+            try:
+                return _ConcretePath.rename(self, other)
+            except FileExistsError:
+                return self._copy_dir_or_file(other)
 
 
 class InPath(OwnedPath):
--- a/qiime2/core/tests/test_path.py
+++ b/qiime2/core/tests/test_path.py
@@ -62,11 +62,14 @@ class TestOwnedPath(unittest.TestCase):
         d._move_or_copy(self.to_dir)
 
         # since from_dir is not owned, but the network fs race condition crops
-        # up, _move_or_copy should copy, not move
-        self.assertTrue(os.path.exists(os.path.join(self.from_dir, 'foo.txt')))
+        # up, _move_or_copy should copy, not move, but then we still ensure
+        # that the original path has been cleaned up
+        self.assertFalse(os.path.exists(os.path.join(self.from_dir,
+                                                     'foo.txt')))
         self.assertTrue(os.path.exists(os.path.join(self.to_dir, 'foo.txt')))
 
-        shutil.rmtree(self.from_dir)
+        with self.assertRaises(FileNotFoundError):
+            shutil.rmtree(self.from_dir)
         shutil.rmtree(self.to_dir)
 
 
--- a/qiime2/core/tests/test_path.py
+++ b/qiime2/core/tests/test_path.py
@@ -7,9 +7,67 @@
 # ----------------------------------------------------------------------------
 
 import os
+import pathlib
+import shutil
+import tempfile
 import unittest
 
-from qiime2.core.path import OutPath
+from qiime2.core.path import OwnedPath, OutPath
+
+
+class TestOwnedPath(unittest.TestCase):
+    def setUp(self):
+        self.from_dir = tempfile.mkdtemp()
+        (pathlib.Path(self.from_dir) / 'foo.txt').touch()
+
+        self.to_dir = tempfile.mkdtemp()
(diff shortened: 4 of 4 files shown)


ebolyen:
This is fantastic! What great detective work, I remember trying to figure this out once and giving up, I had no idea about os.rename, but I suppose it makes sense.

Re: Duplicate, is this worth addressing further? I feel like we could really use our own cheap copy-tree utility that recursively calls duplicate (or even something smarter). This seems like a good motivation to introduce it and fix any other renames in the framework. This shouldn't be construed as a problem you need to solve (i.e. I'm good with merging), but I think we might be staring at some reasonably low-hanging fruit.

ebolyen:
Maybe duplicate isn't quite it, but something that let's us generally "move" things, where you could choose to "destroy" the original (meaning we will try rename, then hard-link+delete, then copy+delete, else error), or force a copy skipping rename.

`clone_tree(destructive=False)/clone_file(destructive=False)`

philippfalk:
Hi, my name is Philipp and I am part of the BeeGFS product support team. I was made aware of this issue by one of our customers.

I would like to learn more about the process on the filesystem that triggers this error. I read through the discussions mentioned above and as far as I understood, having TMPDIR set to a path on a BeeGFS filesystem seems to trigger this, but then some of the reports mention, that the location of the TMPDIR doesn't make a difference. Is it really the TMPDIR or is it the working directory that triggers the error?

I also didn't fully understand the error message that is raised by Python. For example:
```
[Errno 17] File exists: ‘/scratch/qiime2-provenance-difz5eeh’ -> ‘/scratch/qiime2-archive-isgo3hiy/b7f5c095-1f9f-480d-837b-aba392d161c4/provenance’
```
This mentions two filenames. What exactly happens at that step and what does that arrow mean? Is it a copy or a move? Is that related to the rename you mention?

thermokarst:
Hey @philippfalk, thanks for stopping by!

> I would like to learn more about the process on the filesystem that triggers this error.

Me too! I think it is safe to say we are still in the early stages of trying to get a grip on what is going on here. One thing I am still unsure of, is it seems to me like we should be seeing this error on other networked filesystems, but, I can't find any reports on our support forum for a non beegfs mount experiencing these issues. That has me a bit stumped, because I think a significant number of our users utilize QIIME 2 in HPC environments, which almost always seem to have some kind of network fs.

One thing to keep in mind is that we still haven't ruled out definitively any bad behavior on our part - it is still entirely possible that we have some file operations that are running multiple times, but, its worth noting again that for some reason we only see this become a problem in very specific cases.

> Is it really the TMPDIR or is it the working directory that triggers the error?

With respect to QIIME 2 (and the changes introduced in this PR), I have only observed these issues when writing to the system TMPDIR. As far as QIIME 2 goes - we use the tmp dir extensively during operation, to keep track of input and output data, as well as some other bookkeeping activities.

> I also didn't fully understand the error message that is raised by Python

This is a pretty confusing python error message in my opinion. The operation that is raising this is `os.rename`, which is a move (or maybe a link?). I believe the format of the error message is `src -> dst`, and the message is claiming that the dst path already exists.

I'm still looking into this, but any help or insight would be really helpful! Thanks so much, @phillippfalk!

philippfalk:
> I can't find any reports on our support forum for a non beegfs mount experiencing these issues. 

I found this issue, that looks very similar and there seems to be no BeeGFS involved, only NFS:

https://forum.qiime2.org/t/errno-17-file-exists-during-classify/6554

To analyze this a little better, it would be great to have some strace output of the specific code path that triggers this. I will work together with our customer to get that.
 
Regarding the rename: There is a limitation in BeeGFS, which makes it impossible to move open files to a different directory. The error that is triggered should be different (EBUSY), but that is the only limitation I can think of that seems somewhat related to what we are seeing here.

thermokarst:
Great, thanks so tracking that post down, @philippfalk! I'm still surprised that there aren't more reports like that floating around. Anyway, I'm going to work on doing a more thorough review of the current handling of these tmp files from within the QIIME 2 framework. I'll report back what I find sometime in the next few days.

thermokarst:
Okay, after taking a closer look at the python stdlib docs, it really does look to me like the change I have proposed here in this PR probably should've been implemented all along (or something like it, as @ebolyen proposed above). `os.rename` is just not guaranteed to work across different filesystems, although it sounds like in general we've been pretty lucky, and in most cases, it has seemed to work for us.

I think in general we haven't used `shutil.move` for these operations because a) it isn't exposed in `pathlib` and b) it isn't atomic.

The framework falls over on its use of `os.rename` in at least two cases, where the following directories _might_ exist: 1) [the archive's data directory](https://github.com/qiime2/qiime2/blob/6d8932eda130d4a9356f977fece2e252c135d0b9/qiime2/core/archive/format/v0.py#L60) and 2) [the archive's provenance directory](https://github.com/qiime2/qiime2/blob/6d8932eda130d4a9356f977fece2e252c135d0b9/qiime2/core/archive/format/v1.py#L23).

I don't think we can just drop the preemptive creation of those directories, at the very least it would break the contract for things like `InPath` and `OutPath` and would make for some test-churn to accommodate (that's not a real concern though, at the end of the day - I imagine there must be some other reason why we preemptively create those dirs -  @ebolyen, maybe you have some thoughts?). Otherwise, I think its safe to say that we just need to handle the os errors more gracefully in QIIME 2 (although it is a bummer that there isn't better tooling for this in the python stdlib...).

ebolyen:
I suspect any reason for the directories to exist (or not exist) was just an artifact of how I was thinking about these objects. I wouldn't put too much significance in them, the most likely reason I can think of is it would have made ancestral provenance tracking barely simpler as you can just take the contents of X/ and put it in artifacts/UUID/.

I think directory handling in general is pretty much a mess, and I hate reaching for `distutils`, so maybe we categorize some situations and figure out a general API (perhaps after merging this)?

thermokarst:
> so maybe we categorize some situations and figure out a general API (perhaps after merging this)?

Definitely. I'll stick a few test cases in with this PR, then we can merge. Later, we can circle back on planning an overhaul on file/dir move/copy ops.

thermokarst:
Okay @ebolyen, this is ready for review.

qiime2/core/tests/test_path.py
```diff
@@ -7,9 +7,67 @@
...
+    def test_move_or_copy_not_owned_copy(self, _):
+        d = OwnedPath(self.from_dir)
+        # ensure that we are not owned
+        d._user_owned = False
+
+        d._move_or_copy(self.to_dir)
+
+        # since from_dir is not owned, but the network fs race condition crops
+        # up, _move_or_copy should copy, not move
+        self.assertTrue(os.path.exists(os.path.join(self.from_dir, 'foo.txt')))
```
ebolyen, line 66: I think we want this to be false still.

qiime2/core/path.py
```diff
@@ -27,14 +27,22 @@ def __new__(cls, *args, **kwargs):
...
-                return shutil.copy(str(self), str(other))
+            return self._copy_dir_or_file(other)
         else:
-            return _ConcretePath.rename(self, other)
+            # Certain networked filesystems will experience a race
+            # condition on `rename`, so fall back to copying.
+            try:
+                return _ConcretePath.rename(self, other)
+            except FileExistsError:
+                return self._copy_dir_or_file(other)
```
ebolyen, line 45: I think this will leave dangling files, as `self` will not be managing the lifetime of `other` as it would have originally.

qiime2/core/archive/provenance.py
```diff
@@ -317,7 +317,12 @@ def finalize(self, final_path, node_members):
         self.write_action_yaml()
         self.write_citations_bib()
 
-        self.path.rename(final_path)
+        # Certain networked filesystems will experience a race
+        # condition on `rename`, so fall back to copying.
+        try:
+            self.path.rename(final_path)
+        except FileExistsError:
+            distutils.dir_util.copy_tree(str(self.path), str(final_path))
```
ebolyen, line 325: I think `self.path` needs to be updated to be the `final_path` and the original needs to be destroyed.

ebolyen: Well, it's possible `self.path` isn't used after this, but I think we're still going to have a dangling directory in /tmp/

qiime2/core/archive/tests/test_provenance.py
```diff
@@ -262,6 +262,20 @@ def test_ts_to_date(self, mocked_tzlocal):
...
+
+        viz_p_dir = viz._archiver.provenance_dir
+        self.assertTrue(viz_p_dir.exists())
+
+    @mock.patch('qiime2.core.path.ProvenancePath.rename',
+                side_effect=FileExistsError)
+    def test_prov_rename_file_exists(self, _):
+        viz, = dummy_plugin.actions.no_input_viz()
+
+        viz_p_dir = viz._archiver.provenance_dir
```
ebolyen, line 276: I'm not sure this test is enough, as that directory may exist (while being empty). Perhaps we check for the existence of `action.yaml` which should be in there?

ebolyen (changes requested):
Thanks @thermokarst, I think we have some dangling files since we aren't destroying the originals like a `rename` would.

qiime2/core/path.py
thermokarst: Are you specifically worried about the copy operation in this branch (not owned) - or in general (in both the owned and not owned branches)?

qiime2/core/archive/provenance.py
thermokarst: Yeah I thought we already had dangling dirs hanging around in some other cases, so I wasn't too worried about cleaning them up, but I can do that here if you want. I was kinda thinking that would be punted for v3 of this code, though.

qiime2/core/path.py
ebolyen: only the un-owned branch. Basically, if a rename was good enough before, then we weren't planning on keeping the original path around (and so nothing will expect to go back to clean it up).

qiime2/core/archive/provenance.py
ebolyen: I suppose I just remember the horror of working out the lifetimes for all of these objects and so I don't want to lose whatever logic was implied by the `rename`

ebolyen: under normal conditions, the framework won't leave behind any dangling files. Now I've definitely seen this fail with certain errors, but never looked to closely

qiime2/core/path.py
thermokarst: Perfect, thanks for confirming!

qiime2/core/archive/provenance.py
thermokarst: > Now I've definitely seen this fail with certain errors, but never looked to closely

Lol, yeah, me too - I think the race condition implied in this PR is one of those cases!

I'll add some code to clean up! Also, good catch on the `self.path` update - I completely spaced that.

Thanks for the review!

thermokarst: I ran into some issues with updating `self.path`, but ran out of time this morning to look into it further. If you get a second to peek at it, please do.

ebolyen: Ok, looking at the semantics of `rename`, it never actually changed state, so `self.path` can remain stale (it is clearly never used again). Sorry about that, the docs weren't super clear so I tested it on Py3.6 and a `rename` only seems to side-effect the system, but not the object.

qiime2/core/archive/provenance.py
```diff
@@ -317,7 +317,15 @@ def finalize(self, final_path, node_members):
...
+            distutils.dir_util.remove_tree(str(self.path))
+            # TODO: update self.path? This doesn't seem to work as hoped
+            # self.path = qiime2.core.path.ProvenancePath(str(final_path))
```
ebolyen, line 328: Drop the TODO for now and this is good to go

ebolyen (approved):
LGTM, thanks @thermokarst!

qiime2/core/path.py
```diff
@@ -27,14 +27,30 @@ def __new__(cls, *args, **kwargs):
...
+            return self._copy_dir_or_file(other)
         else:
-            return _ConcretePath.rename(self, other)
+            # Certain networked filesystems will experience a race
+            # condition on `rename`, so fall back to copying.
+            try:
+                return _ConcretePath.rename(self, other)
+            except FileExistsError:
+                copied = self._copy_dir_or_file(other)
+                self._destruct()
```
thermokarst, line 52: https://www.youtube.com/watch?v=92T6y2auxAQ

PR #569, opened 2021-03-22
ENH: allow typing.Union in transformers

### LINKED ISSUE #229: Support `typing` package in transformers
**Current Behavior**
Presently `typing` expressions do not pickle well, but it's a small list that can be enumerated manually.

What we are starting to need is the ability for transformers to solve to a `typing` type like [`Union`](https://docs.python.org/3/library/typing.html#typing.Union) which would make paired-end visualizations simpler, as they could request either transformation. 

This gets into some of the bigger things that are wrong with transformers right now and should probably be handled during a refactor to permit transitivity and restrictions on semantic type.

**References**
[`Union`](https://docs.python.org/3/library/typing.html#typing.Union)

misialq:
Closes #229.

Introduces support for `typing.Union` in view types by selecting the first available transformer. Doesn't work on asynchronous calls (as pointed out in the issue description), but that should be fixed when upgrading to Python >= 3.7.


thermokarst:
@Keegan-Evans and I are working on migrating the framework's minimum version of Python to 3.8 this week. We should know more soon!

misialq:
Hey @thermokarst, all fixed and ready for review.

ebolyen (approved):
This looks fantastic, sorry it took so long to get merged!

PR #584, opened 2021-04-30
NEW: async UsageExecutor and dill'ed callables

ebolyen:
Changes to support testing async via mystery stew. Was originally messing around with this to explore paralellism, but realized mystery-stew was broken for async calls.

This change uses dill during getstate and setstate which creates a byte string. This is then pickled by Python, but it can do that correctly, so it let's us effectively use dill for multiprocessing (which wouldn't have otherwise).

By using dill, we can now handle closures which is handy for mystery-stew.


ebolyen:
see https://github.com/qiime2/q2-mystery-stew/pull/23

thermokarst (approved):
Good catch, this LGTM. I like `dill`, it seems like a pretty safe bet for our needs, and the extension of `ExecutionUsage` is clear and makes sense. Any thoughts on tests?

ebolyen:
Should be able to make some examples in a closure (mimicking mystery-stew). And then repeat whatever is testing the ExecutorUsage with a new arg.

thermokarst:
Yeah that sounds good to me - do you want to do that, or would you like me to?

ebolyen:
I can do it, I was just avoiding creating work for myself before the API made sense.

thermokarst:
Makes sense! Yeah I think the action changes make sense. For the usage changes, I like that its scoped to just the execution driver, since most other template-style drivers just won't care about the distinction.

ebolyen:
Verified I get a failing test case without `dill`

thermokarst:
@Oddant1 you're next! This PR LGTM - if it works for you, feel free to merge when ready.

Oddant1 (approved):

PR #586, opened 2021-05-13
IMP: save() updates: adding optional file extension argument

lizgehret:
.save() method now includes the extension argument, which takes string values as input.
This argument has been added for Metadata, Artifact, and Visualization classes.
Default values (if left blank) for each class are as follows:

Metadata: blank
Artifact: .qza
Visualization: .qzv

Additionally, Metadata.save() will now return the filepath (including optional extension), just as Artifact and Visualization currently do.

Unit tests have been implemented for the above changes, as well as docstring additions.

### DIFF
--- a/qiime2/metadata/metadata.py
+++ b/qiime2/metadata/metadata.py
@@ -473,7 +473,7 @@ class Metadata(_MetadataBase):
         """
         return not (self == other)
 
-    def save(self, filepath):
+    def save(self, filepath, extension=None):
         """Save a TSV metadata file.
 
         The TSV metadata file format is described at https://docs.qiime2.org in
@@ -493,7 +493,12 @@ class Metadata(_MetadataBase):
 
         """
         from .io import MetadataWriter
+        
+        if extension is not None and not filepath.endswith(extension):
+            filepath += extension
+
         MetadataWriter(self).write(filepath)
+        return filepath
 
     def to_dataframe(self):
         """Create a pandas dataframe from the metadata.
--- a/qiime2/metadata/metadata.py
+++ b/qiime2/metadata/metadata.py
@@ -487,13 +487,19 @@ class Metadata(_MetadataBase):
         filepath : str
             Path to save TSV metadata file at.
 
-        extension : str
+        ext : str
             Preferred file extension (.tsv, .txt, etc).
             Will be left blank if no extension is included.
             Including a period in the extension is
             optional, and any additional periods delimiting
             the filepath and the extension will be reduced
             to a single period.
+
+        Returns
+        -------
+        str
+            Filepath and extension (if provided) that the
+            file was saved to.
 
         See Also
         --------
@@ -502,18 +508,7 @@ class Metadata(_MetadataBase):
         """
         from .io import MetadataWriter
 
-        if ext is None:
-            ext = ''
-        else:
-            ext = '.' + ext.lstrip('.')
-
-        filepath = filepath.rstrip('.')
-
-        if not filepath.endswith(ext):
-            filepath += ext
-
-        MetadataWriter(self).write(filepath)
-        return filepath
+        return MetadataWriter(self).write(filepath, ext)
 
     def to_dataframe(self):
         """Create a pandas dataframe from the metadata.
@@ -958,7 +953,7 @@ class MetadataColumn(_MetadataBase, metaclass=abc.ABCMeta):
         """
         return not (self == other)
 
-    def save(self, filepath):
+    def save(self, filepath, ext=None):
         """Save a TSV metadata file containing this metadata column.
 
         The TSV metadata file format is described at https://docs.qiime2.org in
@@ -972,9 +967,24 @@ class MetadataColumn(_MetadataBase, metaclass=abc.ABCMeta):
         filepath : str
             Path to save TSV metadata file at.
 
+        ext : str
+            Preferred file extension (.tsv, .txt, etc).
+            Will be left blank if no extension is included.
+            Including a period in the extension is
+            optional, and any additional periods delimiting
+            the filepath and the extension will be reduced
+            to a single period.
+
+        Returns
+        -------
+        str
+            Filepath and extension (if provided) that the
+            file was saved to.
+
         """
         from .io import MetadataWriter
-        MetadataWriter(self).write(filepath)
+
+        return MetadataWriter(self).write(filepath, ext)
 
     def to_series(self):
         """Create a pandas series from the metadata column.
--- a/qiime2/metadata/metadata.py
+++ b/qiime2/metadata/metadata.py
@@ -237,7 +237,7 @@ class _MetadataBase:
 
         """
         from .io import MetadataWriter
-        
+
         if ext is None:
             ext = ''
         else:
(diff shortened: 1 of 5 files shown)


lizgehret:
@thermokarst hitting some flake8 errors due to line length - these are for the commented headers within the unit tests. Happy to modify the names if needed, otherwise will disregard. q2lint ran locally without any issues.

ebolyen:
Exciting! Could I make two requests on the API? @lizgehret and @thermokarst let me know what you think of these:

1. Could the `extension` kwarg be `ext` instead? In this context it isn't ambiguous and I think makes it more convenient.
2. Could we make it so that the leading `'.'` in `ext` is optional (i.e. drop a leading period if it's there then join)? That way you could say `ext='zip'` or `ext='.zip'` and you don't have to go look up which one is "correct".

thermokarst:
I think those suggestions sound great, @ebolyen! @lizgehret, can you work on amending this PR to include those changes? You and I can chat a bit more about the request in item 2 during our mtg tomorrow. I'm glad @ebolyen brought that one up, I had been thinking about ensuring that `.txt` vs `txt` both result in `foo.txt` (instead of `foo.txt` and `footxt`), but totally forgot about it during our discussion today!

lizgehret:
@ebolyen I agree with @thermokarst, these are great suggestions! Extension kwarg has been updated to `ext` - I'll address the `'.'` updates first thing tomorrow and get that pushed through as well!

qiime2/metadata/metadata.py
```diff
@@ -473,10 +473,10 @@ def __ne__(self, other):
         """
         return not (self == other)
 
-    def save(self, filepath):
-        """Save a TSV metadata file.
+    def save(self, filepath, ext=None):
```
thermokarst, line 476: Guess what I just found? There's a related class that needs the same "treatment": `MetadataColumn` - you can find it elsewhere in this file. It might be beneficial to move the business logic into `MetadataWriter`.

qiime2/sdk/result.py
```diff
@@ -185,7 +217,7 @@ def validate(self, level=NotImplemented):
 
 
 class Artifact(Result):
-    extension = '.qza'
+    extension = 'qza'
```
thermokarst, line 220: I think this should be rolled back - the business logic implemented in `Result.save()` handles this case just fine.

```suggestion
    extension = '.qza'
```

qiime2/sdk/result.py
```diff
@@ -322,7 +354,7 @@ def validate(self, level='max'):
 
 
 class Visualization(Result):
-    extension = '.qzv'
+    extension = 'qzv'
```
thermokarst, line 357: Same comment as above:

```suggestion
    extension = '.qzv'
```

qiime2/sdk/result.py
```diff
@@ -143,9 +143,41 @@ def export_data(self, output_dir):
...
+        See Also
+        --------
+        load
+
+        extension : str
+            Preferred file extension (.qza, .qzv, .txt, etc).
+            If no preferred extension input is included,
+            Artifact extension will default to .qza and
+            Visualization extension will default to .qzv.
+            Note that including a period in the extension is
```
thermokarst, line 163: Same suggestions as above, in `Metadata.save()` (prose && location).

qiime2/sdk/result.py
```diff
@@ -143,9 +143,41 @@ def export_data(self, output_dir):
     def _destructor(self):
         return self._archiver._destructor
 
-    def save(self, filepath):
-        if not filepath.endswith(self.extension):
-            filepath += self.extension
+    def save(self, filepath, ext=None):
+        """Save an artifact (.qza) or visualization (.qzv) file.
```
thermokarst, line 147: Let's make this more generic, I think it might be a little confusing to read about Visualizations when saving an Artifact and vice versa. Maybe something like "Save to a file." is enough?

qiime2/sdk/result.py
```diff
@@ -143,9 +143,41 @@ def export_data(self, output_dir):
...
+
+        Parameters
+        ----------
+        filepath : str
+            Path to save artifact or visualization file at.
```
thermokarst, line 152: Same as above, maybe "Path to save file at." ?

thermokarst (changes requested):
This is lookin' great @lizgehret! I have a handful of comments/suggestions (inline) - please let me know if you have any questions or comments! 🦖

qiime2/metadata/metadata.py
lizgehret: @thermokarst so just to clarify, you want me to move the MetadataColumn class to be housed under the MetadataWriter class within metadata > io.py?

thermokarst: Close - I don't think `MetadataColumn` needs to go anywhere, though. You know how over in `results.py` you worked on `Artifact` and `Visualization` by making changes to their parent class, `Result`? I think we actually have something similar here: `Metadata` and `MetadataColumn` are like `Artifact` and `Visualization`. The difference though is that we aren't using direct inheritance - `Metadata` and `MetadataColumn` both just call the `MetadataWriter.write()` method manually. I'm thinking that it might be beneficial to move the "business logic" you have implemented in `Metadata` up into `MetadataWriter.write` - this would be a good use of DRY, IMO. We can chat on a call if you want - you know where to find me.

lizgehret: @thermokarst got it, thank you!! Will push up those changes for review right now!

qiime2/metadata/metadata.py
```diff
@@ -203,6 +203,54 @@ def _filter_ids_helper(cls, df_or_series, ids, ids_to_keep):
...
+
+        Returns
+        -------
+        str
+            Filepath and extension (if provided) that the
+            file was saved to.
+
+        See Also
+        --------
+        load
```
thermokarst, line 236: Now that this docstring is shared via this parent class, you'll need to update this to let all the children know _which_ `load` method to use:

```suggestion
        Metadata.load
```

qiime2/sdk/result.py
```diff
@@ -143,9 +143,40 @@ def export_data(self, output_dir):
...
+        extension : str
+            Preferred file extension (.qza, .qzv, .txt, etc).
+            If no preferred extension input is included,
+            Artifact extension will default to .qza and
+            Visualization extension will default to .qzv.
+            Including a period in the extension is
+            optional, and any additional periods delimiting
+            the filepath and the extension will be reduced
+            to a single period.
+
```
thermokarst, line 163: This is missing the 'Returns' section:

```suggestion

        Returns
        -------
        str
            Filepath and extension (if provided) that the
            file was saved to.
```

thermokarst (changes requested):
This is looking great, @lizgehret! Two very minor adjustments requested inline, then this should be ready to merge. Thanks so much, and if you have any questions, you know where to find me!

lizgehret (commented):
All requested changes updated during pair programming session on 5/17.

qiime2/sdk/result.py
```diff
@@ -143,9 +143,45 @@ def export_data(self, output_dir):
...
+        Returns
+        -------
+        str
+            Filepath and extension (if provided) that the
+            file was saved to.
+        See Also
```
thermokarst, line 169:
```suggestion
            file was saved to.

        See Also
```

thermokarst (approved):
🎉

PR #591, opened 2021-06-09
Update test_mapped_actions.py

costjose:
Tests updated to include only one assert statement each

### DIFF
--- a/qiime2/core/testing/tests/test_mapped_actions.py
+++ b/qiime2/core/testing/tests/test_mapped_actions.py
@@ -36,7 +36,7 @@ class ActionTester(unittest.TestCase):
 class TestConstrainedInputVisualization(ActionTester):
     ACTION = 'constrained_input_visualization'
 
-    def test_match_foo(self):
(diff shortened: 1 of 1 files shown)


ebolyen:
Hi @costjose!

I want to point out that this changeset introduces a lot of duplicated code and I believe conflates the idea of asserting a single "thing" with a single "assert call".

What we are generally interested in (especially in AAA style tests, such as these) is that something is arranged, an action taken, and then a single _conceptual_ assertion is made. In these tests the thing we want to ensure, is the fact that both inputs are appropriately handled. By duplicating the tests, this actually compromises (though not by very much) our ability to validate the logical AND, and adds a lot of extra code to understand.

Whereas, by having multiple assert calls in this particular pattern, ideally the reader would infer that these assertions are conceptually related and we are verifying some joined aspect of the behavior, rather than independent aspects (which should indeed be their own test).

That said, thank you for your interest in QIIME 2! I hope I haven't discouraged you in any way from continuing to poke around and seek improvements (as there are always things to improve)!

PR #595, opened 2021-07-29
FEAT: Semantic Validation

lizgehret:
FEAT: Add validation and test suite for Semantic Types within the QIIME 2 framework.


qiime2/plugin/plugin.py
```diff
@@ -135,6 +138,22 @@ def register_views(self, *views, citations=None):
...
 
+    def register_validator(self, semantic_expression):
+        if not is_semantic_type(semantic_expression):
+            raise TypeError('%s is not a Semantic Type' % semantic_expression)
+
+        def decorator(validator):
+            for semantic_type in semantic_expression:
+                if semantic_type not in self.validators:
+                    self.validators[semantic_type] = []
+                self.validators[semantic_type].append(ValidatorRecord(
```
ebolyen, line 149: I would probably construct a single ValidatorRecord, just so that their object IDs are shared between partitions. This isn't likely to be all that useful implementation wise, but I expect it would simplify testing and debug.

ebolyen: (instead of making a new one in each iteration of the loop)

qiime2/core/validate.py
```diff
@@ -0,0 +1,30 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016-2021, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+class SemanticValidation:
```
ebolyen, line 9: What if this object was per-partition?

Then the sorting of the entire system could be amortized, and we don't create this objects just to discard them moments later.

qiime2/core/testing/format.py
```diff
@@ -148,3 +148,14 @@ def _validate_(self, level):
...
+
+class Cephalapod(TextFileFormat):
+    """
+    Class that inherits from text file format.
+    Used for testing validator sorting.
+    """
+
+
+CephalapodDirectoryFormat = model.SingleFileDirectoryFormat(
+    'CephalapodDirectoryFormat', 'squids.tsv', Cephalapod)
```
ebolyen, line 161: :octopus:

qiime2/core/testing/plugin.py
```diff
@@ -149,8 +164,8 @@
         ('concatenated_ints', IntSequence1)
     ],
     name='Concatenate integers',
-    description='This method concatenates integers into a single sequence in '
-                'the order they are provided.',
+    description='This method concatenates integers into'
+                'a single sequence in the order they are provided.',
```
ebolyen, line 168: missing space here between 'into' and 'a'

qiime2/core/tests/test_validate.py
```diff
@@ -0,0 +1,279 @@
...
+               'validator_sort_last']
+
+        exp2 = ['validator_sort_first',
+                'validator_sort_middle_b',
+                'validator_sort_middle',
+                'validator_sort_last']
+
+        obs = [record.validator.__name__ for record in test_object.validators]
+
+        self.assertIn(obs, [exp, exp2])
```
ebolyen, line 180: nice approach to the ambiguity here! (which is worth preserving to keep the tests from being brittle to the implementation)

qiime2/core/validate.py
```diff
@@ -0,0 +1,105 @@
...
+        A partial order sort that will return a sorted list of validators. The
+        runtime for this sort is \theta(n^2). This is not a concern, as the
+        number of validators present for any particular type is expected to
+        remain trivially low. The validators are sorted from general to
+        specific.
+        """
+        self._validators = sorted_poset(
+            iterable=self._validators,
+            key=lambda record: record.context,
+            reverse=True)
```
ebolyen, line 63: :+1:

qiime2/core/validate.py
```diff
@@ -0,0 +1,105 @@
...
+            except Exception as e:
+                raise ImplementationError("An unexpected error occured when %r"
+                                          " from %r attempted to validate %r"
+                                          % (record.validator.__name__,
+                                             record.plugin,
+                                             data)) from e
+
+    def assert_transformation_available(self, dir_fmt):
+        """
+        Called by `qiime2.sdk.PluginManager._consistency_check` to ensure that
```
ebolyen, line 92: This is probably better as a comment, as it doesn't exactly explain how to use this public method (but the context is probably handy for the maintainer).

qiime2/core/validate.py
```diff
@@ -0,0 +1,105 @@
...
+    def __init__(self, concrete_type):
+        self._validators = []
+        self.concrete_type = concrete_type
+        self._is_sorted = False
+
+    def add_validator(self, validator_record):
+        """
+        Used by Plugin to add a `ValidatorRecord` for a new validator to a
+        plugin. See plugin/plugin for definition of `ValidatorRecord`. Usually
+        called through the `register_validtor` decorator.
```
ebolyen, line 24: These aren't quite numpydoc docstrings. I would check out scikit-bio or qiime2/sdk/usage.py for some good examples.

qiime2/sdk/plugin_manager.py
```diff
@@ -51,15 +52,27 @@ def iter_entry_points(cls):
...
+                cls.__instance = None
+                raise
         else:
             if add_plugins is False:
                 raise ValueError(
                     'PluginManager singleton already exists, cannot change '
                     'default value for `add_plugins`.')
         return cls.__instance
 
+    def destroy_singleton(self):
```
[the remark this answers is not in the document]
ebolyen: (bumping this into the current review) (I like forget_singleton I think)

ebolyen (changes requested):
Really awesome work @Keegan-Evans and @lizgehret!

I think this is a really solid implementation. Few nitpicks on some documentation, and a last-minute change to the public method on PluginManager (changing it later would be more annoying).

qiime2/sdk/result.py
```diff
@@ -275,11 +275,12 @@ def import_data(cls, type, view, view_type=None):
...
         provenance_capture = archive.ImportProvenanceCapture(format_, md5sums)
         return cls._from_view(type_, view, view_type, provenance_capture,
-                              validate_level='max')
+                              level='max')
 
     @classmethod
     def _from_view(cls, type, view, view_type, provenance_capture,
-                   validate_level='min'):
+                   level='min'):
+        type_raw = type
```
ebolyen, line 283: This isn't needed I think.

qiime2/sdk/result.py
```diff
@@ -275,11 +275,12 @@ def import_data(cls, type, view, view_type=None):
 
         provenance_capture = archive.ImportProvenanceCapture(format_, md5sums)
         return cls._from_view(type_, view, view_type, provenance_capture,
-                              validate_level='max')
+                              level='max')
```
ebolyen, line 278: I would revert this particular parameter back to `validate_level` (it's not public, and clarifies things at this call-site).

qiime2/sdk/result.py
```diff
@@ -275,11 +275,12 @@ def import_data(cls, type, view, view_type=None):
...
 
     @classmethod
     def _from_view(cls, type, view, view_type, provenance_capture,
-                   validate_level='min'):
+                   level='min'):
```
ebolyen, line 282: see above for reversion on `level`

qiime2/core/transform.py
```diff
@@ -63,9 +63,9 @@ def make_transformation(self, other, recorder=None):
                      input_record=self._record, output_name=other._view_name,
                      output_record=other._record)
 
-        def transformation(view, validate_level='min'):
+        def transformation(view, level='min'):
```
ebolyen, line 66: Revert these back as well (there's no "level" of transformation, rather we run the validation system, so the call-site would be ambiguous if called via key-word).

ebolyen (changes requested):
Sorry forgot to check if these had flipped back to the original yet.

qiime2/core/testing/format.py
lizgehret: q2-squid-squad

qiime2/core/tests/test_validate.py
Keegan-Evans: Thanks!

qiime2/core/transform.py
```diff
@@ -135,12 +135,12 @@ def coerce_view(self, view):
 
         return view
 
-    def validate(self, view, level='min'):
+    def validate(self, view, validate_level='min'):
```
ebolyen, line 138: Grep has overshot the replacement I think.

The general rule here was:
`validate(..., level=...)`

anywhere else that calls validate as a side-effect (rather than express purpose):
 `unrelated(validate_level=...)`

qiime2/core/exceptions.py
```diff
@@ -9,3 +9,7 @@
 
 class ValidationError(Exception):
     pass
+
+
+class ImplementationError(Exception):
```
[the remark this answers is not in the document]
Keegan-Evans: Do you want ValidationError added as well?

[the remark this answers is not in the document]
Keegan-Evans: Do you want ValidationError exported as well?

qiime2/core/validate.py
```diff
@@ -0,0 +1,141 @@
...
+        *others: Any number of validation objects of the same concrete type.
+
+        Notes
+        -----
+        Used to combine validation objects from different plugins. This is
+        done non-heirarchically by `PluginManager` by creating a new, blank
+        object for each `concrete_type` that it encounters, then adds the
+        objects from each plugin.
+
+        """
```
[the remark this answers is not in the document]
Keegan-Evans: Good call! Right now this is only "enforced" by only calling it when the concrete types match, but as far as the actual method goes there is no safeguard.

qiime2/core/exceptions.py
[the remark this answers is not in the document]
ebolyen: That's probably a good idea if it isn't already there.

qiime2/core/validate.py
```diff
@@ -0,0 +1,192 @@
...
+            The single, complete semantic type that the validators are to be
+            associated with.
+
+        Attributes
+        ----------
+        _validators: list
+            A list of ValidatorRecords
+
+        _is_sorted: Bool
+            Tracks whether or not `_validators` has been sorted or not.
```
ebolyen, line 40: Turn these attributes (`_validators` and `_is_sorted`) into a comment instead of public docs and this is good to go!

ebolyen (approved):
Thanks for all the work!

PR #597, opened 2021-08-20
Tests from_checksum_format() on filepaths with leading \'s

ChrisKeefe:
Confirms `from_checksum_format()`'s handling of leading-backslash filepaths, like those in the last couple lines here:

image


ebolyen (approved):
Thanks @ChrisKeefe!

ChrisKeefe:
Does this actually merit a changelog entry?

PR #614, opened 2022-01-06
Usage tweaks

### LINKED ISSUE #612: Inconsistent `import_from_format` parameter naming between Usage and its implementations
**Bug Description**
[`usage.Usage.import_from_format`](https://github.com/qiime2/qiime2/blob/39ac17da01e22057ff38197eb23ad6cca48f4c2e/qiime2/sdk/usage.py#L925) takes a parameter named `variable` in the third position.

[`ArtifactAPIUsage.import_from_format`](https://github.com/qiime2/qiime2/blob/6ef6df712f2f14be1baa5551368a41c3e9f8e340/qiime2/plugins.py#L190) and [`CLIUsage.import_from_format`](https://github.com/qiime2/q2cli/blob/bee9b88bc5a9a8beb116f5252d68a4b85361cbec/q2cli/core/usage.py#L136) both use `fmt_variable` in the same position.

Users who approach usage examples from the [docs](https://dev.qiime2.org/latest/api-reference/usage/authors/#importing) are likely to pass keyword arguments to this method, given that the rendered example does so. (I certainly did). 

**Proposed solutions**
Rename the variable variable-name to `variable` in all child Usages, so it aligns with the parent. This keeps the published API stable, even if the name is a little unspecific. We could also make these arguments positional-only, which is restrictive and a little gross but more explicitly indicates that implementations may not share variable names, and means that users get a more explicit error message: 

image

**Screenshots**
This error message left me wondering if I'd misspelled something:
image

Happy to ship a PR for this if that's useful.

### LINKED ISSUE #613: file extension handling fragile in `init_format`
**Bug Description**
The `ext` argument to `init_format` is rendered into "<name>.<ext>" by both the ArtifactAPIUsage and the CLIUsage.

If, as in the [dev-docs example](https://dev.qiime2.org/latest/api-reference/usage/authors/#qiime2.sdk.usage.Usage.init_format), a file extension string is passed with a leading `.` the rendering looks like `some_name..extension` (note two `.`s).

**proposed solution**
lstrip `.` from arguments to `ext`

**Screenshots**
image

ChrisKeefe:
A collection of minor tweaks impacting Usage
- closes #613 
- closes #612 
- some typos

### DIFF
--- a/qiime2/plugins.py
+++ b/qiime2/plugins.py
@@ -178,7 +178,7 @@ class ArtifactAPIUsage(usage.Usage):
 
     def init_format(self, name, factory, ext=None):
         if ext is not None:
-            name = '%s.%s' % (name, ext)
+            name = '%s.%s' % (name, ext.lstrip('.'))
 
         variable = super().init_format(name, factory, ext=ext)
 
--- a/qiime2/plugins.py
+++ b/qiime2/plugins.py
@@ -188,12 +188,12 @@ class ArtifactAPIUsage(usage.Usage):
         return variable
 
     def import_from_format(self, name, semantic_type,
-                           fmt_variable, view_type=None):
-        variable = super().import_from_format(
-            name, semantic_type, fmt_variable, view_type=view_type)
+                           variable, view_type=None):
+        imported_var = super().import_from_format(
+            name, semantic_type, variable, view_type=view_type)
 
-        interface_name = variable.to_interface_name()
-        import_fp = fmt_variable.to_interface_name()
+        interface_name = imported_var.to_interface_name()
+        import_fp = variable.to_interface_name()
 
         lines = [
             '%s = Artifact.import_data(' % (interface_name,),
@@ -209,7 +209,7 @@ class ArtifactAPIUsage(usage.Usage):
         self._update_imports(from_='qiime2', import_='Artifact')
         self._add(lines)
 
-        return variable
+        return imported_var
 
     def merge_metadata(self, name, *variables):
         variable = super().merge_metadata(name, *variables)
--- a/qiime2/sdk/usage.py
+++ b/qiime2/sdk/usage.py
@@ -820,7 +820,7 @@ class Usage:
         Parameters
         ----------
         name : str
-            The cannonical name of the variable to be returned.
+            The canonical name of the variable to be returned.
         factory : Callable which returns :class:`qiime2.sdk.Artifact`
             A function which takes no parameters, and returns an artifact.
             This function may do anything internally to create the artifact.
@@ -856,7 +856,7 @@ class Usage:
         Parameters
         ----------
         name : str
-            The cannonical name of the variable to be returned.
+            The canonical name of the variable to be returned.
         factory : Callable which returns :class:`qiime2.Metadata`
             A function which takes no parameters, and returns metadata.
             This function may do anything internally to create the metadata.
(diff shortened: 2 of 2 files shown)


ebolyen:
TIL how to spell canonical, lol

qiime2/plugins.py
```diff
@@ -188,12 +188,12 @@ def init_format(self, name, factory, ext=None):
...
     def import_from_format(self, name, semantic_type,
-                           fmt_variable, view_type=None):
+                           variable, view_type=None):
         variable = super().import_from_format(
-            name, semantic_type, fmt_variable, view_type=view_type)
+            name, semantic_type, variable, view_type=view_type)
 
         interface_name = variable.to_interface_name()
-        import_fp = fmt_variable.to_interface_name()
+        import_fp = variable.to_interface_name()
```
ebolyen, line 196: Namespace collision here same as in q2cli. (Use `var` for the super'ed return result to avoid)

thermokarst: This is why the sig name was `fmt_variable`, fwiw.

thermokarst: Not that that's a good reason or anything, this is just for context. I could've just as easily picked a different name for the `super()` result, but such is life.

qiime2/plugins.py
```diff
@@ -188,12 +188,12 @@ def init_format(self, name, factory, ext=None):
...
-                           fmt_variable, view_type=None):
-        variable = super().import_from_format(
-            name, semantic_type, fmt_variable, view_type=view_type)
+                           variable, view_type=None):
+        var = super().import_from_format(
+            name, semantic_type, variable, view_type=view_type)
 
         interface_name = variable.to_interface_name()
-        import_fp = fmt_variable.to_interface_name()
+        import_fp = variable.to_interface_name()
```
ebolyen, line 196: `variable` to `var` for this line (same as in q2cli)

ebolyen (changes requested):

qiime2/plugins.py
ebolyen: sorry wrong line, the one right after.

qiime2/plugins.py
```diff
@@ -188,12 +188,12 @@ def init_format(self, name, factory, ext=None):
...
 
     def import_from_format(self, name, semantic_type,
-                           fmt_variable, view_type=None):
-        variable = super().import_from_format(
-            name, semantic_type, fmt_variable, view_type=view_type)
+                           variable, view_type=None):
+        var = super().import_from_format(
+            name, semantic_type, variable, view_type=view_type)
 
         interface_name = variable.to_interface_name()
```
ebolyen, line 195: This line needs to become `var` now, because it's the resulting artifact variable (same as in q2cli).

qiime2/sdk/usage.py
```diff
@@ -1202,7 +1202,7 @@ def help(self, action: 'qiime2.sdk.usage.UsageAction'):
 
         Examples
         --------
-        >>> use.help(use.UsageAction('dummpy_plugin', 'split_ints'))
```
thermokarst, line 1205: dummpy ❤️ darwub

ebolyen: I love this

ebolyen: I may just be an under-parameterized RNN...

thermokarst: i just lolled at that.

ebolyen (approved):
LGTM, thanks @ChrisKeefe!

PR #623, opened 2022-03-07
Artifact cache

Oddant1:


qiime2/core/cache.py
```diff
@@ -0,0 +1,162 @@
...
+        # be in the cache. Not sure how we would really enforce that, but we
+        # can just... Heavily encourage it I guess
+        # os.mkdir('tmp')
+
+        self.version.write_text(
+            _VERSION_TEMPLATE % (self.CURRENT_FORMAT_VERSION,
+                                 qiime2.__version__))
+
+    # Tell us if the path is a cache or not
+    def is_cache(self):
```
ebolyen, line 89: You'll probably want this to be a class method, which can be invoked on the Class object, or an instance.

```python
@classmethod
def is_cache(cls, path):
    pass
```

Instead of recieving `self` you will receive your class back (or the relevant subclass).

Oddant1: Good idea

Oddant1:
I believe the test failures (outside of the garbage collection test which just needs to be reworked to understand the cache works now) are due to the fact that the process pool only contains the alias_of artifacts and not the original artifacts. All of the failures seem to be tests looking in the process pool for the original non-alias artifact and not finding it. @ebolyen do you have any ideas on how we want to handle this?

ebolyen (changes requested):
This is fantastic @Oddant1! Just pulled this down and have been playing around with it. It's working really well.

I have a few points that we probably need to address, and then some general nice-to-haves/ideas for post-release

**lint**

0. Checkout `qiime2/core/testing/util.py` from `master`; some docstrings lost their lines and there's no changes at this point.

**important**

1. `Cache.load()` needs to induce a symlink in the relevant process pool:
    ```python
    c1 = Cache('foo')
    c2 = Cache('bar')
    with c1:
        a = Artifact.load('whatever.qza')  # has a symlink, all good
    b = c2.save(a, 'whatever')  # does not have one, and only exists because the key does
    c2.remove('whatever')  # key no longer exists, but `b` still does
    b.view(SomeType)  -> raises FileNotFound as it's been GC'ed

2. We should set the read-only flag on `/data/<uuid>` recursively after it's been reasonably finalized (which may mean calling a new Cache method inside Archiver to indicate as such). The reason for this is that we've never had artifacts live longer than a plugin, and so it's possible we have misbehaved plugins which will mutate the directory/file they're given. If that happens, we'd like to know about them. This also means you will need to recursively mark files as writable when performing a GC, otherwise GC won't work.

3. We need to extend the process-pool duration considerably, to like 45 days. It's not important they get cleaned up when something that catastrophic happens, but we should get around to it eventually.

**nice to have**

4. Process pools may get reaped by a cache that has a different duration, so we might want to parameterize the process pool directory with a days-to-live argument or similar so that all cache instances can equally respect that, no matter how they are individually parameterized.

5. In the event there is a process we can view on the system with the same start/user/PID combo, then there's no reason to reap it even if it has expired. This one isn't super important but it would be nice to not be rude.

6. An Archiver might want to store the cache (or potentially process pool) used when it was initialized, as there is a very slight chance that something may change the cache between initialization and some failure resulting in key destruction (e.g. our try-except blocks in archiver).

7. Related to the above, if would be good to have some kind of limited destructor on Archiver which will clean up the process-pool's symlink on the self-stored cache. That way, when you run the unit-tests, or any code that creates thousands of artifacts that get disposed of, the process pool doesn't grow without bound. This is just cleaning up the refs, not the data itself, that's still GC's job.

8. Re-entrancy of Cache and Pool context managers. We should prevent where possible entering a Cache within a Cache and entering a Pool inconsistent with the currently entered Cache. We may also just enter the Cache of a Pool if no cache is current. I think these can be accomplished with some simple flags on the instance like `self._is_entered = True` and threadlocal `__cache__.is_entered = True`

**notes about directory structure, since I stare at it because it's neat**

9. It seems like the keys exist in mutually exclusive states, it might be cool to make them symlinks also, but when it's a symlink to a pool, we could prefix the pool with `pool-` so that we knew it was a pool, and if it didn't have that prefix, we would know it was data. Then you can basically know the state of a Cache from just `tree -a` which would be really cool.

10. `process/` should probably be something plural like `processes` to match the other directories.

11. I'm pretty stoked to put other temporary stuff in here someday soon. It's so much nicer than what we used to do and I'm really glad to be rid of `qiime2-archive-asdjhasdkjahf` finally.

---

Only items 0-3 need to be addressed prior to merge, and I hope that they are easy to do. The rest are as you see fit, although I might put re-entrancy towards the top of the nice-to-haves since it's more user-facing.

ebolyen:
Point 1. is probably worth a test and would be a pretty useful integration test as it touches on so many concepts by the end of it without being very hard to instigate.

qiime2/core/cache.py
```diff
@@ -0,0 +1,511 @@
...
+        # and remove it leaving us with a dangling reference and no data
+        with self.lock:
+            self._register_key(key, str(ref.uuid))
+
+        _copy_to_data(self, ref, self.data)
+
+        # We want this ref to also be backed by the proces pool so for the
+        # duration of this process we still have access to it even if we delete
+        # the key
+        self.process_pool.save(ref)
```
ebolyen, line 350: @Oddant1, we probably want this to happen in `load()`, since you could end up with the same issue if you loaded a key, and then deleted it. (Also, since save falls through load, you get two birds with one stone).

Oddant1: Good idea

ebolyen (approved):
This is super exciting @Oddant1! Thanks for your hard work on this, I think this is really going to advance what's possible with QIIME 2.

PR #625, opened 2022-05-06
BUG: semantic validation's transformations

ebolyen:
Discovered in q2-fmt. When multiple validators exist, the "source" `data` was overwritten within a loop, causing the next iteration to fail as `data` would have the wrong type for the next selected transformer.

### DIFF
--- a/qiime2/core/tests/test_validate.py
+++ b/qiime2/core/tests/test_validate.py
@@ -133,6 +133,7 @@ class TestValidationObject(unittest.TestCase):
         validator_object = ValidationObject(IntSequence1)
 
         has_run = False
+        has_also_run = False
 
         def test_validator_method(data: list, level):
             nonlocal has_run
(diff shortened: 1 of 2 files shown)


ebolyen:
TODO:

 - [ ] add a test that fails without this change

ebolyen:
Updated test such that it fails in this way (same kind of failure as q2-fmt) without [420e860](https://github.com/qiime2/qiime2/pull/625/commits/420e860029d40701d2bc753c5ed38835526dd82d):

```
qiime2/core/tests/test_validate.py:160: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
qiime2/core/validate.py:151: in __call__
    data = transformation(data)
qiime2/core/transform.py:68: in transformation
    self.validate(view, level=validate_level)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <qiime2.core.transform.FormatType object at 0x7f541c10ef10>
view = [0, 1, 2], level = 'min'

    def validate(self, view, level='min'):
        if not isinstance(view, self._view_type):
>           raise TypeError("%r is not an instance of %r."
                            % (view, self._view_type))
E           TypeError: [0, 1, 2] is not an instance of <class 'qiime2.core.testing.format.IntSequenceFormat'>.

qiime2/core/transform.py:140: TypeError
=========================== short test summary info ============================
FAILED qiime2/core/tests/test_validate.py::TestValidationObject::test_run_validators
```

lizgehret (approved):
changes look great, thanks for catching this @ebolyen!

thermokarst (approved):
Good catch, what a silly little bug (with big consequences).

Keegan-Evans (approved):
Looks great to me, good catch!

PR #632, opened 2022-08-08
NEW/MISC: allow dict for output registrations and error/sort on sets

ebolyen:
Fixes a long-standing annoyance for output registration.
Also, prevents the plugin author from accidentally creating a `set` of tuples as the output registration.

Sorts `set`s provided to `Choices` so that interfaces which cache the types will maintain a consistent ordering.

### DIFF
--- a/qiime2/core/testing/plugin.py
+++ b/qiime2/core/testing/plugin.py
@@ -160,9 +160,9 @@ dummy_plugin.methods.register_function(
         'int1': Int,
         'int2': Int
     },
-    outputs=[
-        ('concatenated_ints', IntSequence1)
-    ],
+    outputs={
+        'concatenated_ints': IntSequence1
+    },
     name='Concatenate integers',
     description='This method concatenates integers into'
                 ' a single sequence in the order they are provided.',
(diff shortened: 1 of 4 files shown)


qiime2/core/testing/plugin.py
```diff
@@ -160,9 +160,9 @@
         'int1': Int,
         'int2': Int
     },
-    outputs=[
-        ('concatenated_ints', IntSequence1)
-    ],
+    outputs={
+        'concatenated_ints': IntSequence1
+    },
```
ebolyen, line 165: These are only a few, but commonly used dummy plugin actions to demonstrate that this works. There isn't a corresponding `test_signature.py` and most of that setup code is exercised heavily in other tests (such as these).

qiime2/core/type/signature.py
```diff
@@ -109,6 +109,14 @@ def __init__(self, callable, inputs, parameters, outputs,
             Output name to description string.
 
         """
+        # update type of outputs if needed
+        if type(outputs) is list:
+            outputs = dict(outputs)
+        elif type(outputs) is set:
+            raise ValueError("Plugin registration for %r cannot use a set()"
+                             " to define the outputs, as the order is random."
+                             % callable.__name__)
```
ebolyen, line 118: This was surprisingly easy to have happen without noticing:
```python
plugin.register_function(
...
    outputs={
        ('foo', FooType),
        ('bar', BarType)
    },
...
)
```

does not look particularly wrong, but would result in mis-ordered outputs which randomly shuffled themselves.

Keegan-Evans (approved):
Looks good as far as I can tell!

PR #635, opened 2022-08-22
Parsl

Oddant1:
New PR basing parsl off of the master that includes the initial artifact cache


Oddant1:
This branch now contains all of the 2022.8 patches

Oddant1:
blocking qiime2/q2cli#294

qiime2/core/testing/pipeline.py
```diff
@@ -114,30 +116,108 @@ def resumable_collection_pipeline(ctx, int_list, int_dict, fail=False):
...
+        uuids.append([str(result.uuid) for result in dict_ret.values()])
+
+        uuids.append(str(identity_ret.uuid))
+
+        uuids.append(str(viz_ret.uuid))
+
+        raise PipelineError(uuids)
+
+    return (ints1_ret, ints2_ret, int1_ret, list_ret, dict_ret, identity_ret,
+            viz_ret)
```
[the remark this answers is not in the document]
Oddant1: Sort of resolved by https://github.com/qiime2/qiime2/pull/635/commits/93c7f6b37a3548cdd96ad4ad6ae0b965b7f8a320. This commit just added a `result` method to `Results` to mirror the one on `ProxyResults`. This method should be called by pipeline developers inside of try-except blocks and context managers and all that to ensure things work as expected when their pipeline is executed with parsl.

ci/recipe/meta.yaml
```diff
@@ -13,6 +13,8 @@ build:
 
 requirements:
   host:
+    # TODO: Pin version of openssl that supports @rpath/libssl.1.1.dylib
+    - openssl =1.1.1q
```
ebolyen, line 17: Is this still needed? Was it OS X where this happened?

Oddant1: I don't remember honestly. I think we are good?

Oddant1: https://github.com/qiime2/qiime2/commit/d7cc9becd534f11b43c88d47b622015b2ccc128d Looks like that was pinned for Mac, and things are working now, so I think we are good?

qiime2/sdk/__init__.py
```diff
@@ -9,14 +9,14 @@
 from .context import Context
 from .action import Action, Method, Visualizer, Pipeline
 from .plugin_manager import PluginManager, UninitializedPluginManagerError
-from .result import Result, Artifact, Visualization
```
ebolyen, line 12: We should hoist `ResultCollection` up to the root `__init__` as it's perfectly reasonable for a Python API user to instantiate one of these, although kind of more work than they need to do. They might also want to do `isinstance` checks on `Results<>` which would be fine as well.

qiime2/sdk/action.py
```diff
@@ -14,30 +14,62 @@
...
 import qiime2.sdk
 import qiime2.core.type as qtype
 import qiime2.core.archive as archive
 from qiime2.core.util import LateBindingAttribute, DropFirstParameter, tuplize
+from qiime2.sdk.parsl_config import PARSL_CONFIG, setup_parsl
+from qiime2.sdk.proxy import Proxy
 
 
-def _subprocess_apply(action, ctx, args, kwargs):
+def _subprocess_apply(action, ctx, execution_ctx, args, kwargs):
```
ebolyen, line 28: I don't think we need to accept `execution_ctx`, because `foo.asynchronous()` is the only thing that calls this helper, so it's fixed as `{type: asynchronous}` for the purposes of `action._bind()`

qiime2/sdk/results.py
```diff
@@ -110,3 +110,10 @@ def __repr__(self):
 
     def _asdict(self):
         return dict(zip(self._fields, self))
+
+    def result(self):
+        """ This exists to provide a standardized interface with ProxyResults.
+            Check the 'result' method on ProxyResults for a full
+            explanation.
+        """
+        return self
```
ebolyen, line 119: :a: :heavy_plus_sign:

qiime2/sdk/action.py
```diff
@@ -14,30 +14,62 @@
...
 
 
+def run_parsl_action(action, ctx, execution_ctx, args, kwargs, inputs=[]):
+    """This is what the parsl app itself actually runs. It's basically just a
+    wrapper around our QIIME 2 action
+    """
+    remapped_kwargs = {}
+    for key, value in kwargs.items():
+        if isinstance(value, Proxy):
+            remapped_kwargs[key] = value.get_element(inputs[value._future_])
```
ebolyen, line 46: We might want to use a trivial variable here to communicate what we're working with, for example:
```
resolved_results = inputs[value._future_] # should this be _future_index_ instead?
... = value.get_element(resolved_results)
```

qiime2/sdk/action.py
```diff
@@ -219,8 +255,16 @@ def bound_callable(*args, **kwargs):
...
-                return qiime2.sdk.Results(self.signature.outputs.keys(),
-                                          outputs)
+                results = qiime2.sdk.Results(
+                    self.signature.outputs.keys(), outputs)
+
+                # If we are running a pipeline through parsl, we need to create
+                # a future here because the parsl join app the pipeline was
+                # running in is expected to return a future
+                if isinstance(self, Pipeline) and ctx.parsl:
+                    return _create_future(results)
```
ebolyen, line 265: Let's move this `_create_future` up to `run_parsl_action` just before returning, since this only matters to parsl.

qiime2/sdk/action.py
```diff
@@ -512,6 +668,15 @@ def _init(cls, callable, inputs, parameters, outputs, plugin_id, name,
...
 
+@python_app
+def _create_future(results):
+    """ This is a bit of a dumb hack. It's just a way for us to make pipelines
+    return a future which is what Parsl wants a join_app to return even though
+    we will have real results at this point.
+    """
+    return results
+
+
```
ebolyen, line 679: Bookmarking my progress so far.

qiime2/core/archive/provenance.py
```diff
@@ -428,24 +436,24 @@ def add_parameter(self, name, type_expr, parameter):
...
             self.inputs[name] = None
+        elif isinstance(input, qiime2.sdk.result.ResultCollection):
+            # If we took a Collection input, we will have a ResultCollection,
+            # and we want the keys to line up with the processed values we were
+            # given, so we can maintain the order of the artifacts
+            for artifact in input.values():
+                record = self.add_ancestor(artifact)
+                values.append(record)
+            self.inputs[name] = \
+                [{k: v} for k, v in zip(input.keys(), values)]
```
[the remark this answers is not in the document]
Oddant1: I feel like I had some reason to not do that. Don't remember what it was or if it's still valid.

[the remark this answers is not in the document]
Oddant1: I can use input.items() it just looks a little weird, but it's fine

qiime2/core/type/collection.py
```diff
@@ -116,12 +120,14 @@ def is_element_expr(self, self_expr, value):
         """Since this is a dictionary, we often need to make sure to use its
         values not its keys.
         """
+        from qiime2.sdk import ResultCollection
+
         contained_expr = self_expr.fields[0]
-        if isinstance(value, self._view) and len(value) > 0:
+
+        if (isinstance(value, ResultCollection) or
+                isinstance(value, self._view)) and len(value) > 0:
```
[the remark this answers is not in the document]
Oddant1: I'm not sure I have strong feelings either way about this. I did it this kinda way initially for obvious reasons (ResultCollection didn't exist when I started writing this), and for the other Collections _view is a Python built-in, honestly the ideal would kinda be for it to have dict and ResultCollection both as its views since it treats them identically. OR I guess in a perfect world, it would only ever see ResultCollection not dict, but given where this stuff gets called, that isn't feasible.

qiime2/sdk/results.py
Oddant1: 🤓

qiime2/sdk/parsl_config.py
```diff
@@ -0,0 +1,193 @@
...
+
+def _process_key(key, value):
+    """Takes a key given in the parsl config file and turns its value into the
+    correct data type or class instance to be used in instantiating a
+    parsl.Config object.
+    """
+    # Our key needs to point to some object.
+    if key in module_paths:
+        module = importlib.import_module(module_paths[key])
+        class_ = getattr(module, value.pop('class'))
```
ebolyen, line 154: not important... like at all. But people generally get around this by using either `cls` or `klass` depending on context/tradition.

ebolyen: (the toml key itself is obviously fine)

qiime2/sdk/proxy.py
```diff
@@ -0,0 +1,193 @@
...
+        return self._archiver.format
+
+    @property
+    def citations(self):
+        return self._archiver.citations
+
+    def result(self):
+        return self.get_element(self._future_.result())
+
+    def get_element(self, results):
```
ebolyen, line 50: Should this become `_get_element_` to discourage use outside of the framework?

qiime2/sdk/proxy.py
```diff
@@ -0,0 +1,193 @@
...
+
+class ProxyArtifact(ProxyResult):
+    """This represents a future Artifact that is being returned by a Parsl app
+    """
+    def __repr__(self):
+        if self._signature_ is None:
+            return f'<artifact: Unknown Type {object.__repr__(self)}>'
+        else:
+            return \
+                f'<artifact: {self._signature_[self._selector_].qiime_type}>'
```
ebolyen, line 69: Technically, we want the type that `signature.solve_outputs()` comes up with for this selector.

qiime2/sdk/results.py
```diff
@@ -110,3 +110,10 @@ def __repr__(self):
 
     def _asdict(self):
         return dict(zip(self._fields, self))
+
+    def result(self):
```
ebolyen, line 114: Looking at this again, it actually prevents someone from making an output named "result", as they would want to be able to say:

```
results = do_action()
results.result  # suddenly a function instead of an artifact
```

The workaround is to make it a `_result`. This is the same kind of concession that `namedtuple` makes for it's methods, so while it's strange, it's basically the same contract as the other instance methods.

qiime2/sdk/proxy.py
```diff
@@ -0,0 +1,193 @@
...
+    @property
+    def format(self):
+        return self._archiver.format
+
+    @property
+    def citations(self):
+        return self._archiver.citations
+
+    def result(self):
+        return self.get_element(self._future_.result())
```
ebolyen, line 48: There's not a matching no-op method for this on `Result`.

qiime2/sdk/proxy.py
```diff
@@ -0,0 +1,193 @@
...
+        self._selector_ = selector
+        self._signature_ = signature
+
+    @property
+    def _archiver(self):
+        return self.result()._archiver
+
+    @property
+    def type(self):
+        return self._archiver.type
```
ebolyen, line 33: Strictly speaking, this doesn't need to block, as the type can be resolved from `signature.solve_outputs` + selector. And I _believe_ enough information exists to make that happen before anything gets scheduled by the parsl app.

qiime2/sdk/proxy.py
```diff
@@ -0,0 +1,193 @@
...
+    def keys(self):
+        return self.result().collection.keys()
+
+    def values(self):
+        return self.result().collection.values()
+
+    def items(self):
+        return self.result().collection.items()
+
+    def get_element(self, results):
```
ebolyen, line 105: Same comment here re: `_get_element_`

qiime2/sdk/proxy.py
```diff
@@ -0,0 +1,193 @@
...
+        """Get the result we want off of the future we have
+        """
+        return getattr(results, self._selector_)
+
+    def result(self):
```
ebolyen, line 110: I believe this one is missing a no-op over on the real class as well.

qiime2/sdk/proxy.py
```diff
@@ -0,0 +1,193 @@
...
+
+        if is_collection_type(qiime_type):
+            return ProxyCollection(self._future_, selector, self._signature_)
+        elif is_visualization_type(qiime_type):
+            return ProxyVisualization(
+                self._future_, selector, self._signature_)
+
+        return ProxyArtifact(self._future_, selector, self._signature_)
+
+    def result(self):
```
ebolyen, line 176: This probably needs to be `_result()` to avoid namespace collisions with `__getattr__`/registered outputs

qiime2/sdk/proxy.py
```diff
@@ -0,0 +1,193 @@
...
+# Copyright (c) 2016-2023, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+from qiime2.core.type.util import is_visualization_type, is_collection_type
+
+
+class Proxy:
```
ebolyen, line 11: We may need to implement things like `__eq__` and friends from the base class, since we would want to block.

qiime2/sdk/proxy.py
```diff
@@ -0,0 +1,193 @@
...
+    def __repr__(self):
+        if self._signature_ is None:
+            return f'<visualization: Unknown Type {object.__repr__(self)}>'
+        else:
+            return \
+                f'<visualization: ' \
+                f'{self._signature_[self._selector_].qiime_type}>'
+
+
+class ProxyCollection(Proxy):
```
ebolyen, line 90: We need things like `__iter__` and `__getitem__` etc

qiime2/sdk/result.py
```diff
@@ -556,3 +469,132 @@ def get_index_paths(self, relative=True):
...
+    def _get_result_fp(cls, directory, result_name):
+        result_fp = os.path.join(directory, result_name)
+
+        # Check if thing in .order file exists and if not try it with .qza at
+        # the end and if not try it with .qzv at the end
+        if not os.path.isfile(result_fp):
+            result_fp += '.qza'
+
+            if not os.path.isfile(result_fp):
+                result_fp = result_fp.rstrip('.qza')
```
ebolyen, line 531: rstrip doesn't mean what everyone wants it to mean, and removesuffix doesn't exist yet

qiime2/sdk/tests/test_config.py
```diff
@@ -0,0 +1,164 @@
...
+
+        list_execution_contexts = self._load_alias_execution_contexts(
+            list_return)
+        dict_execution_contexts = self._load_alias_execution_contexts(
+            dict_return)
+
+        list_expected = [{
+            'type': 'parsl',
+            'parsl_type':
+            "<class 'parsl.executors.high_throughput.executor."
```
ebolyen, line 82: If we use `v.__class__.__name__` then this can look like `HighThroughputExecutor` which will match the config better

qiime2/sdk/tests/test_result.py
```diff
@@ -93,15 +94,16 @@ def test_roundtrip_ordered_collection(self):
         with open(os.path.join(output_fp, '.order')) as fh:
             self.assertEqual(fh.read(), 'foo\nbar\n')
 
-        read_collection = Result.load_collection(output_fp)
+        read_collection = ResultCollection.load(output_fp)
```
ebolyen, line 97: I really like how this diff looks. It's neat.

ebolyen:
Also, let's add a test of provenance for the collection output-name jazz

qiime2/sdk/result.py
```diff
@@ -556,3 +469,132 @@ def get_index_paths(self, relative=True):
     def _repr_html_(self):
         from qiime2.jupyter import make_html
         return make_html(str(self._archiver.path))
+
+
+class ResultCollection:
```
ebolyen, line 474: If you add `.type` as a property, then some interface logic can end up being simpler

Oddant1:
> Also, let's add a test of provenance for the collection output-name jazz

For pipeline specifically

qiime2/sdk/proxy.py
Oddant1: I need a signature object to do that, but I don't have access to one. Should I just use the entire Signature object for _signature_ instead of just the outputs or something? That would make more sense with the name anyway lol.

qiime2/sdk/proxy.py
Oddant1: You can tell I just copy-pasted this from Result because otherwise I would have at least used the approach from the repr lol

qiime2/sdk/proxy.py
ebolyen: You would pass this in before it's evaluated by Parsl in theory. So you have the inputs, you can use solve and annotate the proxies with this new info

Oddant1: So `_signature_` would be the solved outputs?

ebolyen: yeah I think so

qiime2/sdk/result.py
Oddant1: I would want the return from `.type` to be a `TypeExp` of `Collection[Whatever]` right? I'm not positive how to construct one of those. I can get the internal bit by just grabbing it off the first element in the collection, but I'm not sure about wrapping it in the `Collection` part.

qiime2/sdk/result.py
```diff
@@ -556,3 +474,146 @@ def get_index_paths(self, relative=True):
...
+    def __setitem__(self, key, item):
+        self.collection[key] = item
+
+    def __getitem__(self, key):
+        return self.collection[key]
+
+    @property
+    def type(self):
+        # Get the type of the first element in the collection, all elements in
+        # the collection should have the same type.
```
ebolyen, line 576: This is only true for _outputs_, inputs or user-constructed versions could have any type they want.

So long as there's some input which would accept `Collection[Foo | Bar]` that will work out fine. We just can't have that type as an output, as we would have to treat the union weirdly to compose correctly.

We can use `UnionExp(v.type for v in self.collection.values())` to create a really dumb looking Union, then we can call `.normalize()` on the result to reduce it to a normal form

Oddant1: That is disgusting, but you're right

qiime2/sdk/proxy.py
```diff
@@ -39,21 +44,37 @@ def __init__(self, future, selector, signature=None):
...
+
     @property
     def _archiver(self):
         return self.result()._archiver
 
     @property
     def type(self):
-        return self._archiver.type
+        if self._signature_ is not None:
+            return self._signature_[self._selector_].qiime_type
```
ebolyen, line 64: Do we still need to index via the selector here? It seems like we can just have the type as a private attr

qiime2/sdk/proxy.py
```diff
@@ -67,42 +88,46 @@ def _get_element_(self, results):
...
-        else:
-            return \
-                f'<artifact: {self._signature_[self._selector_].qiime_type}>'
-
     def view(self, type):
         """If we want to view the result we need the future to be resolved
         """
         return self._get_element_(self._future_.result()).view(type)
 
+    def has_metadata(self):
```
ebolyen, line 111: You could replace this with `has_metadata = Artifact.has_metadata` which will get bound to this class in the usual way.

qiime2/sdk/proxy.py
```diff
@@ -22,7 +26,8 @@ def _create_proxy(self, selector):
         qiime_type = self._signature_[selector].qiime_type
```
ebolyen, line 26: You could probably pass `qiime_type` into the proxy constructors instead of `self._signature_` since each proxy is always just a single entry.

ebolyen (commented):
Only minor comments :D

qiime2/sdk/proxy.py
Oddant1: Shenanigans

Oddant1: Unfortunately, getting some circular imports here trying to implement these shenanigans

PR #636, opened 2022-08-24
Nonexistent user

Oddant1:
Resolve potential issues primarily related to running QIIME 2 in containers accessing the cache

### DIFF
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -88,6 +88,26 @@ def _copy_to_data(cache, ref):
         set_permissions(destination, READ_ONLY_FILE, READ_ONLY_DIR)
 
 
+def _get_user():
+    """getpass.getuser could fail for reasons explained below, so we use this
+    """
+    try:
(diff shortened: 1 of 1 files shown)


qiime2/core/cache.py
```diff
@@ -520,6 +533,18 @@ def __exit__(self, *args):
         _CACHE.cache = self.previously_entered_cache
         self.cache.named_pool = None
 
+    def _get_process_pool_name(self):
+        """Gets the necessary info to create/identify a process pool for this
+        process
+        """
+        pid = os.getpid()
+        user = getpass.getuser()
```
ebolyen, line 541: @Oddant1 I think this will run into the same issue. We should create some helper to create the "user" which everything in this class can use.

Oddant1: Yep good catch. Didn't even think that maybe we called getpass.getuser more than once.

ebolyen (approved):
Tested locally, and it appears to work!

Issue #639, opened 2022-09-01
qiime2 import error due to multiple users needing to access /tmp/qiime2 on Ubuntu server


jcmcnch:
**Bug Description**

I ran into a cryptic permissions error when trying out the same import command in two different user accounts on an Ubuntu remote server. Both are using the same `qiime2-2022.8` deployment, installed in the same way (according to qiime2 docs for conda, only difference being I used mamba to speed things up). 

Import command:

```
qiime tools import \
  --type 'SampleData[PairedEndSequencesWithQuality]' \
  --input-path manifest.tsv \
  --output-path 16s.qza \
  --input-format PairedEndFastqManifestPhred33V2
```

This is what I was seeing in terms of an error message:

```
Traceback (most recent call last):
  File "/home/inigo/miniconda3/envs/qiime2-2022.8/lib/python3.8/site-packages/q2cli/builtin/tools.py", line 157, in import_data
    artifact = qiime2.sdk.Artifact.import_data(type, input_path,
  File "/home/inigo/miniconda3/envs/qiime2-2022.8/lib/python3.8/site-packages/qiime2/sdk/result.py", line 299, in import_data
    return cls._from_view(type_, view, view_type, provenance_capture,
  File "/home/inigo/miniconda3/envs/qiime2-2022.8/lib/python3.8/site-packages/qiime2/sdk/result.py", line 334, in _from_view
    artifact._archiver = archive.Archiver.from_data(
  File "/home/inigo/miniconda3/envs/qiime2-2022.8/lib/python3.8/site-packages/qiime2/core/archive/archiver.py", line 379, in from_data
    path, process_alias, cache = cls._make_temp_path(uuid)
  File "/home/inigo/miniconda3/envs/qiime2-2022.8/lib/python3.8/site-packages/qiime2/core/archive/archiver.py", line 274, in _make_temp_path
    cache = get_cache()
  File "/home/inigo/miniconda3/envs/qiime2-2022.8/lib/python3.8/site-packages/qiime2/core/cache.py", line 67, in get_cache
    _CACHE.temp_cache = Cache()
  File "/home/inigo/miniconda3/envs/qiime2-2022.8/lib/python3.8/site-packages/qiime2/core/cache.py", line 169, in __init__
    self.path = pathlib.Path(self._get_temp_path())
  File "/home/inigo/miniconda3/envs/qiime2-2022.8/lib/python3.8/site-packages/qiime2/core/cache.py", line 264, in _get_temp_path
[...]

PR #640, opened 2022-09-01
2022.8 patches

Oddant1:
Starting to fix the multi-user issues with the cache first referenced here #639.


Oddant1:
Note that this branch is based on the 2022.8 release commit, not the couple of others that got merged into master since then. I have never created a patch PR before, so I may have done something wrong and what I did should be checked against the patching procedure more closely before we apply the patch.

Oddant1:
Actually was not a typo I am blind sometimes

PR #644, opened 2022-09-09
Mac culling bad

Oddant1:


qiime2/core/cache.py
```diff
@@ -191,6 +215,18 @@ def __init__(self, path=None, process_pool_lifespan=45):
         # We were used by this process
         USED_CACHES.add(self)
 
+        # Start thread that pokes things in the cache to ensre they aren't
+        # culled for being too old
+        self._thread_is_done = threading.Event()
```
ebolyen, line 220: @Oddant1, I realized on the walk home that we really only want to do this if `path` is `None`, since we should expect user-provided caches to be stable to this sort of nonsense.

qiime2/core/cache.py
```diff
@@ -207,6 +243,18 @@ def __exit__(self, *args):
         """
         _CACHE.cache = None
 
+    def __getstate__(self):
+        """We don't want to pickle any of the thread stuff because it won't
+        rehydrate, and we don't care about it
+        """
+        threadless_dict = self.__dict__.copy()
+
+        del threadless_dict['_thread_is_done']
```
ebolyen, line 252: We should check if our cache has a `._thread` at all before deleting, as we would get a keyerror otherwise.

i.e. a un-pickled cache is pickled again, or if the above change is made and not all caches have the silly thread watcher

ebolyen (changes requested):
@Oddant1, sorry, we need to make one more adjustment.

qiime2/core/cache.py
ebolyen: `hasattr()` is the easy way to do this

ebolyen (approved):
Going to double check this on our mac worker (which won't be running as root) to be 100% sure. But this looks good, thank you @Oddant1!

ebolyen:
Our mac worker passed, and so has Github! Everything seems to be working as expected on OS X as well.

PR #646, opened 2022-09-16
IMP: add `optional` to File and FileCollection.

Keegan-Evans:
No tests yet!
Also adds better error message for unimplemented pathmaker as well.

### DIFF
--- a/qiime2/plugin/model/tests/test_directory_format.py
+++ b/qiime2/plugin/model/tests/test_directory_format.py
@@ -1,0 +1,50 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2022, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+
+import shutil
+import unittest
+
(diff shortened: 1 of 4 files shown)


ebolyen:
Looking good! I would add one more test to demonstrate that leaving `optional` out entirely has the default behavior of `False`.

qiime2/plugin/model/tests/test_directory_format.py
```diff
@@ -0,0 +1,104 @@
...
+
+            format_object.validate()
+
+    def test_fails_missing_with_optional_default(self):
+        files_dir_fp = self.get_data_path('test_text_files/')
+
+        with self.assertRaisesRegex(ValidationError,
+                                    "Missing one or more files for "
+                                    "AllRequiredDefaultDirFmt: "
+                                    "\'test_text3.txt\'"):
```
ebolyen, line 89: This assertion is brittle to the sort order of the internal logic. I would make it similar to your above test and omit the specific filename missing (or make the directory format simpler with only a single property instead of 3).

Keegan-Evans:
Ok, I am going to add one more test that checks that it raises if you give it a file that it does not know how to store.

PR #650, opened 2022-10-04
Cache docs

Oddant1:
Add necessary documentation for the cache in a form that we can use for the dev docs.

### DIFF
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -206,25 +206,27 @@ def monitor_thread(cache_dir, is_done):
 class Cache:
     """General structure of the cache:
 
-    artifact_cache/
-    ├── data/
-    │   ├── uuid1
-    │   ├── uuid2
-    │   ├── uuid3
-    │   └── uuid4
-    ├── keys/
-    │   ├── bar.yaml
-    │   ├── baz.yaml
-    │   └── foo.yaml
-    ├── pools/
-    │   └── puuid1/
-    │       ├── uuid1 -> ../../data/uuid1/
-    │       └── uuid2 -> ../../data/uuid2/
-    ├── processes/
-    │   └── <pid>-<create-time>@<uname>/
-    │       ├── uuid3 -> ../../data/uuid3/
-    │       └── uuid4 -> ../../data/uuid4/
-    └── VERSION
+    ::
+
+        artifact_cache/
+        ├── data/
+        │   ├── uuid1
+        │   ├── uuid2
+        │   ├── uuid3
+        │   └── uuid4
+        ├── keys/
+        │   ├── bar.yaml
+        │   ├── baz.yaml
+        │   └── foo.yaml
+        ├── pools/
+        │   └── puuid1/
+        │       ├── uuid1 -> ../../data/uuid1/
+        │       └── uuid2 -> ../../data/uuid2/
+        ├── processes/
+        │   └── <pid>-<create-time>@<uname>/
+        │       ├── uuid3 -> ../../data/uuid3/
+        │       └── uuid4 -> ../../data/uuid4/
+        └── VERSION
 
     Data: The data directory contains all of the artifacts in the cache in
     unzipped form.
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -535,10 +535,8 @@ class Cache:
         >>> cache = Cache(cache_path)
         >>> pool = cache.create_pool(keys=['some', 'kinda', 'keys'])
         >>> keys = os.listdir(cache.keys)
-        >>> # Sort to make sure the order is actually what we want below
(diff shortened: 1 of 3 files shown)


Oddant1:
Blocking qiime2/dev-docs#57

qiime2/core/cache.py
```diff
@@ -484,23 +762,78 @@ def load(self, key):
...
+        True
+        >>> test_dir.cleanup()
         """
         os.remove(self.keys / key)
         self.garbage_collection()
 
     def clear_lock(self):
-        """Clears the lock on the cache
-        NOTE: Forcibly removes the lock outside of the locking library's API
+        """Clears the flufl lock on the cache.
```
ebolyen, line 810: This one probably needs more description, it may be worth referencing it in the GC method as well, just so a user has a more complete picture.

qiime2/core/cache.py
```diff
@@ -531,34 +883,120 @@ def _deallocate(self, symlink):
...
         return self.path / 'VERSION'
 
 
-# Assume we will make this its own class for now
 class Pool:
+    """Pools are folders in the cache that contain many symlinks to many
+    different piece of data. There are two types of pool:
+
+    **Process Pools:** These pools have names of the form
+    <pid>-<create-time>@<uname> based on the process that created them. They
```
ebolyen, line 968:
```suggestion
    <process-id>-<process-create-time>@<user> based on the process that created them. They
```

just to make it consistent with the env-var style above.

qiime2/core/cache.py
```diff
@@ -604,15 +1077,29 @@ def __enter__(self):
...
+        different cache set when we set this named pool, we would have errored
+        in __enter__.
         """
         _CACHE.cache = self.previously_entered_cache
         self.cache.named_pool = None
 
     def _get_process_pool_name(self):
-        """Gets the necessary info to create/identify a process pool for this
-        process
+        """Creates a process pool name of the format <pid>-<start_time>@<user>
```
ebolyen, line 1096:
```suggestion
        """Creates a process pool name of the format <process-id>-<process-create-time>@<user>
```

ebolyen (changes requested):
Really good job @Oddant1, I think there's a few places for improvement, but overall the prose is great.

qiime2/core/cache.py
Oddant1: I'm not sure I have the complete picture. I don't completely remember what motivated us to make this function. I remember we decided it might be useful to be able to remove this lock file (especially when it was the only lock we had), but I don't remember any particular justification or use case for this.

qiime2/core/cache.py
```diff
@@ -465,15 +708,50 @@ def _register_key(self, key, value, pool=False):
...
+        -------
+        Result
+            The loaded data pointed to by the key.
+
+        Raises
+        ------
+        ValueError
+            If the key does not reference any data meaning you probably tried
+            to load a pool.
+        ValueError
```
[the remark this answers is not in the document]
ebolyen: @Oddant1, wrong error-mode, not-found should translate to `KeyError` whereas pool vs data makes more sense as a `ValueError` (since the key exists...)

qiime2/core/cache.py
```diff
@@ -965,9 +976,9 @@ class Pool:
     different piece of data. There are two types of pool:
 
     **Process Pools:** These pools have names of the form
-    <pid>-<create-time>@<uname> based on the process that created them. They
-    only exist for the length of the process that created them and ensure data
-    that process is using stays in the cache.
+    <process-id>-<process-create-time>@$USER based on the process that created
```
ebolyen, line 985: I would stick with the OG formatting, just use `user` instead of `uname` to be consistent

qiime2/core/cache.py
```diff
@@ -1093,8 +1104,9 @@ def __exit__(self, *args):
         self.cache.named_pool = None
 
     def _get_process_pool_name(self):
-        """Creates a process pool name of the format <pid>-<start_time>@<user>
-        for the process that invoked this function.
+        """Creates a process pool name of the format
+        <process-id>-<process-create-time>@USER for the process that invoked
```
ebolyen, line 1114: same here

qiime2/core/cache.py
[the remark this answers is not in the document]
Oddant1: I was kinda thinking that, but I interpreted your initial comment as asking for the opposite and just didn't think about it too hard.

ebolyen:
Sorry for the merge conflicts

qiime2/core/cache.py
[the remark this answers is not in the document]
ebolyen: No worries!

ebolyen (approved):

Issue #649, opened 2022-10-03
enable registered descriptions of semantic types and formats


gregcaporaso:
This will facilitate documentation across different interface types and will enable automated generation of pages [like this](https://docs.qiime2.org/2022.8/semantic-types/#common-semantic-types), which are hard to keep up-to-date.

ebolyen:
Fixed in #655

PR #655, opened 2022-10-18
adds support for registering descriptions and example imports for artifact classes

gregcaporaso:
Adds support for registering descriptions and import examples for _artifact classes_. _Artifact class_ is a new term that we're using to refer to semantic types with a registered directory format. Previously these would be registered with `Plugin.register_semantic_type_to_format`, and that is still possible for backward compatibility. These can also now be registered with `Plugin.register_artifact_class`, which optionally takes a description and a dict of usage examples.  

Addresses #649.

### DIFF
--- a/qiime2/core/testing/plugin.py
+++ b/qiime2/core/testing/plugin.py
@@ -100,13 +100,17 @@ dummy_plugin.register_views(
     SingleIntFormat, RedundantSingleIntDirectoryFormat,
     citations=[citations['mayer2012walking']])
 
-dummy_plugin.register_semantic_type_to_format(
+dummy_plugin.register_artifact_class(
     IntSequence1,
-    artifact_format=IntSequenceDirectoryFormat
+    artifact_format=IntSequenceDirectoryFormat,
+    description="The first IntSequence",
+    examples=None
 )
-dummy_plugin.register_semantic_type_to_format(
+dummy_plugin.register_artifact_class(
     IntSequence2,
-    artifact_format=IntSequenceV2DirectoryFormat
+    artifact_format=IntSequenceV2DirectoryFormat,
+    description="The second IntSequence",
+    examples=None
 )
 dummy_plugin.register_semantic_type_to_format(
     IntSequence3,
--- a/qiime2/core/testing/plugin.py
+++ b/qiime2/core/testing/plugin.py
@@ -100,17 +100,56 @@ dummy_plugin.register_views(
     SingleIntFormat, RedundantSingleIntDirectoryFormat,
     citations=[citations['mayer2012walking']])
 
+
+# Create IntSequence1 import example usage example
+def is1_use(use):
+    def factory():
+        from qiime2.core.testing.format import IntSequenceFormat
+        from qiime2.plugin.util import transform
+        ff = transform([1, 2, 3], to_type=IntSequenceFormat)
+
+        ff.validate()
+        return ff
+
+    to_import = use.init_format('to_import', factory, ext='.hello')
+
+    ints = use.import_from_format('ints',
+                                  semantic_type='IntSequence1',
+                                  variable=to_import,
+                                  view_type='IntSequenceFormat')
+
+
 dummy_plugin.register_artifact_class(
     IntSequence1,
     directory_format=IntSequenceDirectoryFormat,
     description="The first IntSequence",
-    examples=None
+    examples={'IntSequence1 import example': is1_use}
 )
+
+
+# Create IntSequence2 import example usage example
+def is2_use(use):
+    def factory():
+        from qiime2.core.testing.format import IntSequenceFormatV2
+        from qiime2.plugin.util import transform
+        ff = transform([1, 2, 3], to_type=IntSequenceFormatV2)
+
+        ff.validate()
+        return ff
+
+    to_import = use.init_format('to_import', factory, ext='.hello')
+
+    ints = use.import_from_format('ints',
+                                  semantic_type='IntSequence2',
+                                  variable=to_import,
+                                  view_type='IntSequenceFormatV2')
+
+
 dummy_plugin.register_artifact_class(
     IntSequence2,
     directory_format=IntSequenceV2DirectoryFormat,
     description="The second IntSequence",
-    examples=None
+    examples={'IntSequence2 import example': is2_use}
 )
 dummy_plugin.register_semantic_type_to_format(
     IntSequence3,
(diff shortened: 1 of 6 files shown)


qiime2/core/testing/plugin.py
```diff
@@ -100,33 +100,76 @@
...
+        from qiime2.core.testing.format import IntSequenceFormat
+        from qiime2.plugin.util import transform
+        ff = transform([1, 2, 3], to_type=IntSequenceFormat)
+
+        ff.validate()
+        return ff
+
+    to_import = use.init_format('to_import', factory, ext='.hello')
+
+    ints = use.import_from_format('ints',
```
ebolyen, line 116: The linter is upset about these, you can delete the variable assignment and the usage example won't care

qiime2/core/testing/plugin.py
```diff
@@ -100,33 +100,76 @@
...
     citations=[citations['mayer2012walking']])
 
-dummy_plugin.register_semantic_type_to_format(
+
+# Create IntSequence1 import example usage example
+def is1_use(use):
+    def factory():
+        from qiime2.core.testing.format import IntSequenceFormat
+        from qiime2.plugin.util import transform
+        ff = transform([1, 2, 3], to_type=IntSequenceFormat)
```
ebolyen, line 109: This is probably the best use of `transform` within a `plugin` context I've seen. I always forget this function exists.

qiime2/plugin/plugin.py
```diff
@@ -99,14 +103,13 @@ def actions(self):
...
-            for type_ in record.type_expression:
-                types[str(type_)] = \
-                    SemanticTypeRecord(semantic_type=type_, plugin=self)
-
-        return types
+    @property
+    def type_formats(self):
+        # self.type_formats was replaced with self.artifact_classes - this
+        # property provides backward compatibility
+        return self.artifact_classes
```
ebolyen, line 112: :+1:

qiime2/plugin/plugin.py
```diff
@@ -248,22 +251,77 @@ def register_semantic_types(self, *type_fragments):
...
-        self.type_formats.append(TypeFormatRecord(
-            type_expression=semantic_type, format=artifact_format,
-            plugin=self))
+        if description is None:
+            description = ""
+        if examples is None:
+            examples = {}
+
+        registered_artifact_classes = \
+            set(str(e.type_expression) for e in self.artifact_classes)
```
ebolyen, line 272: This may be worth storing on the plugin instance as a private member, as otherwise we'll be constructing this set for each registration, resulting in n^2 iterations through self.artifact_classes.

A good option might be to turn `types` into a mappingproxy of an internal dict, then you can simply check against `self.types` as things are being constructed.

qiime2/plugin/plugin.py
```diff
@@ -79,7 +83,7 @@ def __init__(self, name, version, website, package=None, project_name=None,
         self.views = {}
         self.type_fragments = {}
         self.transformers = {}
-        self.type_formats = []
+        self.artifact_classes = []
```
ebolyen, line 86: This might be better as a dictionary for some of the registration behavior

qiime2/sdk/plugin_manager.py
```diff
@@ -201,7 +201,7 @@ def _integrate_plugin(self, plugin):
                 )
 
             self.formats[name] = record
-        self.type_formats.extend(plugin.type_formats)
+        self.artifact_classes.extend(plugin.artifact_classes)
```
ebolyen, line 204: This was missing originally, but we should do a similar check as the other properties to make sure duplicate registrations (across plugins) are forbidden.

ebolyen (changes requested):
Thanks @gregcaporaso, this all looks good to me, but one thing that I think will simplify a lot of the code would be to turn `artifact_classes` into a dictionary keyed by semantic type.

Then anywhere you cared about a list, `artifact_classes.values()` would be fine, and the rest of the time, you actually are interested in the semantic type for duplicate checking which would be a key.

qiime2/plugin/plugin.py
ebolyen: actually instead of `self.types` `self.artifact_classes` would be fine I think.

qiime2/sdk/plugin_manager.py
gregcaporaso: I added this test and validated that it works manually and reports a useful error message (`ValueError: Duplicate semantic type ('PCoAResults') defined in plugins: 'sapienns' and 'types'`). I didn't see where/if this type of functionality was being unit tested, so I didn't add unit tests for it now.

qiime2/plugin/plugin.py
gregcaporaso: Good call - thanks for catching that! Made this change and most of my commit responding to your comments is updates to address this.

qiime2/sdk/plugin_manager.py
ebolyen: There's no pleasant way to test for this, so eyeballing it is fine with me.

qiime2/sdk/plugin_manager.py
```diff
@@ -350,15 +359,14 @@ def get_directory_format(self, semantic_type):
                 "Must provide a semantic type via `semantic_type`, not %r" %
                 semantic_type)
 
-        dir_fmt = None
-        for artifact_class_record in self.artifact_classes:
+        # Evan, ideally we could just lookup semantic_type in
+        # self.artifact_classes but properties get in the way. Is there a way
+        # to strip properties so this could be simplified to return
+        # self.artifact_classes[semantic_type] while catching a KeyError?
```
ebolyen, line 365: There's not a global helper for this, but there is a `.duplicate()` method which allows you to override attributes of a type, so `(SomeType % Foo).duplicate(predicate=IntersectionExp())` would accomplish your goal. It's not pretty and it's also not recursive, so any variant fields would need to be recursively duplicated as well.

The behavior below is correct and the above sounds like too much work. So let's ignore this for now.

qiime2/sdk/plugin_manager.py
```diff
@@ -344,15 +359,14 @@ def get_directory_format(self, semantic_type):
...
-                dir_fmt = type_format_record.format
-                break
-
-        if dir_fmt is None:
-            raise TypeError(
-                "Semantic type %r does not have a compatible directory format."
-                % semantic_type)
-
-        return dir_fmt
+        # Evan, ideally we could just lookup semantic_type in
```
ebolyen, line 362:
```suggestion
        # TODO: ideally we could just lookup semantic_type in
```

ebolyen (commented):
Just promoting the comment to a TODO, since it's useful context and asks the kind of question which would necessitate some nicer mechanisms than what I described.

ebolyen:
Thanks @gregcaporaso!

gregcaporaso:
Thanks for the reviews @ebolyen!

Issue #657, opened 2022-10-24
support obtaining artifacts and metadata from URLs in usage examples


gregcaporaso:
This was done a lot in the _Cancer Microbiome Intervention Tutorial_, and it was useful there. For example, see the following pages after selecting "View Source". 

<img width="1257" alt="Screen Shot 2022-10-24 at 11 44 43 AM" src="https://user-images.githubusercontent.com/192372/197601765-a6b7bc4e-8a4e-4e8c-9aa5-cdd201ef2102.png">

[Artifacts from URLs](https://docs.qiime2.org/jupyterbooks/cancer-microbiome-intervention-tutorial/030-tutorial-downstream/010-filtering.html)
[Metadata from URLs](https://docs.qiime2.org/jupyterbooks/cancer-microbiome-intervention-tutorial/020-tutorial-upstream/020-metadata.html)

Implementing a helper for this in `qiime2.sdk.usage` will enable us to access this broadly when creating usage examples, and avoid putting `.qza` files that may need to be updated in repositories.

PR #658, opened 2022-10-24
imp: adds helpers to get metadata/artifacts from URLs in usage examples

gregcaporaso:
Addresses #657 

Still needs:
 - [x] doc strings
 - [x] unit tests
 - [x] error detection and reporting on failed downloads (e.g., report 404 with the URL rather than failing when trying to load nonsense data as Metadata)
 - [x] ~`init_metadata_column_from_url`~ ~this is hooked up by passing `column` to `init_metadata_from_url`~ going to skip this in this PR altogether
 - [x] update handling of integration of `{epoch}` in URLs to replace with the most recent epoch
 - [x] is this encoding/replacement of `{epoch}` a reasonable way to handle this? _we decided to use f-strings instead, based on feedback from @ebolyen_
 - [x] address new TODOs throughout

Submitting a draft PR so we can use it in our usage sprint example.

### DIFF
--- a/qiime2/sdk/usage.py
+++ b/qiime2/sdk/usage.py
@@ -909,7 +909,7 @@ class Usage:
         """
         return self._usage_variable(name, factory, 'metadata')
 
-    def init_metadata_from_url(self, name: str, url: str,
+    def init_metadata_from_url(self, name: str, url: str, column: str = None,
                                replace_url_epoch: bool = True
                                ) -> UsageVariable:
         """Communicate that metadata should be obtained from a url."""
@@ -924,7 +924,11 @@ class Usage:
             with tempfile.NamedTemporaryFile() as f:
                 f.write(data.content)
                 f.flush()
-                return qiime2.Metadata.load(f.name)
+                md = qiime2.Metadata.load(f.name)
+                if column is None:
+                    return md
+                else:
+                    return md.get_column(column)
 
         return self.init_metadata(name, factory)
 
--- a/qiime2/sdk/usage.py
+++ b/qiime2/sdk/usage.py
@@ -851,9 +851,9 @@ class Usage:
         return url.replace('{epoch}', '2022.8')
 
     def init_artifact_from_url(self, name: str, url: str,
-                               replace_url_epoch: bool=True) -> UsageVariable:
-        """Communicate that an artifact should be obtained from a url.
-        """
+                               replace_url_epoch: bool = True
+                               ) -> UsageVariable:
+        """Communicate that an artifact should be obtained from a url."""
         if replace_url_epoch:
             url = self._replace_url_epoch(url)
 
@@ -910,9 +910,9 @@ class Usage:
         return self._usage_variable(name, factory, 'metadata')
 
     def init_metadata_from_url(self, name: str, url: str,
-                               replace_url_epoch: bool=True) -> UsageVariable:
-        """Communicate that metadata should be obtained from a url.
-        """
+                               replace_url_epoch: bool = True
+                               ) -> UsageVariable:
+        """Communicate that metadata should be obtained from a url."""
         if replace_url_epoch:
             url = self._replace_url_epoch(url)
 
--- a/qiime2/sdk/usage.py
+++ b/qiime2/sdk/usage.py
@@ -930,14 +930,14 @@ class Usage:
         return url.replace('{epoch}', epoch)
 
     def _request_url(self, url):
-        import requests
+        import urllib.request
+        import urllib.error
 
         try:
-            data = requests.get(url)
-            data.raise_for_status()
-        except requests.exceptions.RequestException as ex:
+            data = urllib.request.urlopen(url)
+        except urllib.error.URLError as ex:
             raise ValueError(
-                'Could not obtain URL: %s\n Requests exception: %s' %
+                'Could not obtain URL: %s\n Exception: %s' %
                 (url, str(ex)))
 
         return data
@@ -988,7 +988,7 @@ class Usage:
             data = self._request_url(url)
 
             with tempfile.NamedTemporaryFile() as f:
-                f.write(data.content)
+                f.write(data.read())
                 f.flush()
                 try:
                     result = qiime2.Artifact.load(f.name)
@@ -1045,7 +1045,7 @@ class Usage:
             data = self._request_url(url)
 
             with tempfile.NamedTemporaryFile() as f:
-                f.write(data.content)
+                f.write(data.read())
                 f.flush()
                 try:
                     md = qiime2.Metadata.load(f.name)
--- a/qiime2/sdk/usage.py
+++ b/qiime2/sdk/usage.py
@@ -963,7 +963,8 @@ class Usage:
         # Examples
         # --------
         # >>> import qiime2
-        # >>> url = 'https://data.qiime2.org/{qiime2.__release__}/data/tutorials/moving-pictures/table.qza' # noqa: E501
+        # >>> url = (f'https://data.qiime2.org/{qiime2.__release__}/data/'
+        # ...        'tutorials/moving-pictures/table.qza')
         # >>> mvp_table = use.init_artifact_from_url('mvp_table', url)
         # >>> mvp_table
         # <ExecutionUsageVariable name='mvp_table', var_type='artifact'>
@@ -988,7 +989,7 @@ class Usage:
 
         return self.init_artifact(name, factory)
 
-    def init_metadata_from_url(self, name: str, url: str, column: str = None,
+    def init_metadata_from_url(self, name: str, url: str,
                                ) -> UsageVariable:
         """Obtain metadata from a url.
 
@@ -1015,7 +1016,8 @@ class Usage:
         Examples
         --------
         >>> import qiime2
-        >>> url = f'https://data.qiime2.org/{qiime2.__release__}/tutorials/moving-pictures/sample_metadata.tsv' # noqa: E501
+        >>> url = (f'https://data.qiime2.org/{qiime2.__release__}/tutorials/'
+        ...        'moving-pictures/sample_metadata.tsv')
         >>> print(url)
         https://data.qiime2.org/20...
         >>> md = use.init_metadata_from_url('md', url)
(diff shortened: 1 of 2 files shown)


gregcaporaso:
This can be used as follows:

```python
def feature_table_filter_samples(use):
    feature_table = use.init_artifact_from_url(
        'feature_table',
        'https://docs.qiime2.org/2022.8/data/tutorials/moving-pictures/table.qza'
        )
    sample_metadata = use.init_metadata_from_url(
        'sample_metadata',
        'https://data.qiime2.org/2022.8/tutorials/moving-pictures/sample_metadata.tsv'
        )

    filtered_table, = use.action(
        use.UsageAction(plugin_id='feature_table', action_id='filter_samples'),
        use.UsageInputs(table=feature_table, metadata=sample_metadata,
                        where="[body-site] IN ('left palm', 'right palm')"),
        use.UsageOutputNames(filtered_table='filtered_table')
    )
```

gregcaporaso:
We should also add ability to get a metadata column from a url.

qiime2/sdk/usage.py
```diff
@@ -922,6 +922,146 @@ def init_format(self, name: str,
         """
         return self._usage_variable(name, factory, 'format')
 
+    def _replace_url_epoch(self, url):
+        # Obtaining epoch modeled on qiime2.metadata.io.MetadataFileError
+        import qiime2
+
+        epoch = qiime2.__release__
```
lizgehret, line 929: I like this approach!

qiime2/sdk/usage.py
```diff
@@ -922,6 +922,146 @@ def init_format(self, name: str,
...
+            data.raise_for_status()
+        except requests.exceptions.RequestException as ex:
+            raise ValueError(
+                'Could not obtain URL: %s\n Requests exception: %s' %
+                (url, str(ex)))
+
+        return data
+
+    def init_artifact_from_url(self, name: str, url: str,
+                               replace_url_epoch: bool = True
```
lizgehret, line 946: I like this too - on the off-chance that we ever don't need the epoch in a given artifact URL, this will be helpful.

qiime2/sdk/usage.py
```diff
@@ -922,6 +922,146 @@ def init_format(self, name: str,
...
+                    raise ValueError(
+                        'Could not load Artifact from URL data: %s\n'
+                        ' Original exception: %s'
+                        % (url, str(ex)))
+
+            return result
+
+        return self.init_artifact(name, factory)
+
+    def init_metadata_from_url(self, name: str, url: str, column: str = None,
```
lizgehret, line 1005: Love the addition of the column parameter, this is so helpful!

lizgehret (approved):
This great, thanks for adding this @gregcaporaso! Everything looks reasonable to me, and it seems like you have good test coverage for the common use-cases.

Let's let @ebolyen take a final look over this one before it's merged (in case there's any missing context on something that could cause problems).

qiime2/sdk/usage.py
```diff
@@ -922,6 +922,146 @@ def init_format(self, name: str,
         """
         return self._usage_variable(name, factory, 'format')
 
+    def _replace_url_epoch(self, url):
+        # Obtaining epoch modeled on qiime2.metadata.io.MetadataFileError
+        import qiime2
```
lizgehret, line 927:
```suggestion
```
`qiime2` is being imported at the top of this file, so it shouldn't be needed as an additional import here.

qiime2/sdk/usage.py
```diff
@@ -922,6 +922,146 @@ def init_format(self, name: str,
...
+
+        epoch = qiime2.__release__
+        return url.replace('{epoch}', epoch)
+
+    def _request_url(self, url):
+        import requests
```
lizgehret, line 933: It's interesting - the `requests` module isn't being found when CI is running your tests. I'm wondering if it actually does need to be added at the top of the file, or maybe in the recipe file as well.

lizgehret (commented):
Added a couple of additional comments inline re: CI failures.

qiime2/sdk/usage.py
gregcaporaso: Oh, I wonder if it's not actually a framework dependency?

ebolyen: I might recommend using stdlib for this (even if it's slightly unpleasant), since there's not too much going on. If we were doing lots of URL requests, then a 3rd party lib might be a good choice.

Since almost everyone uses requests, it's easy for us to end up in a dependency bind someday, so not using it would avoid that.

ebolyen:
Big picture questions:

* These seems like the first example of "utility" methods. It seems these are not generally expected to be overridden by each interface. It occurs to me that it would be perfectly valid for them to do so however. Are we cool with that? It seems reasonable enough.

* I think with python's f-strings we might be able to not worry about the epoch replacement, is it very much work to do:
  ```python
  def example(use):
      import qiime2
      use.init_artifact_from_url(f'http://whatever.org/{qiime2.__epoch__}/foo.qza')
  ```
  This avoids any unanticipated behavior, and we should be able to consider `__epoch__` to be stable. Otherwise, I think the default should probably be `False`. The epoch replacement parameter also makes implementing interface overriding more complicated, as they would all need some mechanism to repeat the epoch logic.

* The column parameter on `init_metadata_from_url` is a pretty good idea, however will the lack of friction cause any confusion when relating it to other methods? Said differently, is the current "syntactic salt" of having to split the two operations beneficial as it requires acknowledging the difference, or should we go the other way and also add `use.init_metadata_column` which is a similar utility method? If that happened, should we have `init_metadata_column_from_url` to keep things clear?

  * An alternative would be to add `column` to `init_metadata` as well, but this would break the API for each interface which implements those, so I don't think we want to do that.

gregcaporaso:
Thanks for the feedback @lizgehret and @ebolyen!

@ebolyen:
| `use.init_artifact_from_url(f'http://whatever.org/{qiime2.__epoch__}/foo.qza')`

Did you mean `qiime2.__release__` here? `qiime2.__epoch__` doesn't appear to be defined.

ebolyen:
Oh yeah, sorry. That's what I get for doing quick drive-by comments.

gregcaporaso:
@lizgehret, @ebolyen - I partially addressed the comments here. 

> These seems like the first example of "utility" methods. It seems these are not generally expected to be overridden by each interface. It occurs to me that it would be perfectly valid for them to do so however. Are we cool with that? It seems reasonable enough.

I agree - generally don't expect them to be overridden, but I don't think there is harm if they are. 

> I think with python's f-strings...

That makes sense to me, and I implemented this change. For folks using the previous version from this PR, you would change something like this:

```python
metadata_url = \
            'https://data.qiime2.org/{epoch}/tutorials/moving-pictures/sample_metadata.tsv'
use.init_metadata_from_url('md', metadata_url)
```
to something like this:

```python
import qiime2
metadata_url = \
            f'https://data.qiime2.org/{qiime2.__release__}/tutorials/moving-pictures/sample_metadata.tsv'
use.init_metadata_from_url('md', metadata_url)
```

> The column parameter on init_metadata_from_url is a pretty good idea, however...

I think you're right that we shouldn't take this approach. Something I didn't notice until today is that the `var_type` on the result was incorrect in the current implementation in this PR (it's `metadata` when it should be `column`). 

Going with `use.init_metadata_column` and `use.init_metadata_column_from_url` seems to make the most sense to me. I spent some time on this this afternoon but am stuck without working code at the moment - not sure what I'm doing wrong. I probably won't have time to get back to this till late this week - if someone else wants to take a crack and commit to this PR that's fine with me. [Here's a diff](https://github.com/qiime2/qiime2/files/9864731/diff.txt) of the code that I have locally right now where this is not working.

qiime2/sdk/usage.py
```diff
@@ -922,6 +922,131 @@ def init_format(self, name: str,
...
+        Returns
+        -------
+        UsageVariable
+            This particular return class can be changed by a driver which
+            overrides :meth:`usage_variable`.
+
+        Examples
+        --------
+        >>> import qiime2
+        >>> url = f'https://data.qiime2.org/{qiime2.__release__}/tutorials/moving-pictures/sample_metadata.tsv' # noqa: E501
```
gregcaporaso, line 1018: I can't figure out how to get a multi-line string working as input in a doc string. As it stands, I'm disabling the max line length check for flake8. Open to other ideas though.

ebolyen: I think this will do what you want. Assuming I remembered how to show continued expressions in doctest.

```suggestion
        >>> url = (f'https://data.qiime2.org/{qiime2.__release__}/tutorials/'
        ...        'moving-pictures/sample_metadata.tsv')
```

lizgehret: I tested @ebolyen's suggestion locally, and that worked for me!

gregcaporaso: I could have sworn I tried that but I guess not! Or maybe something unrelated was failing when I tried and mistook it for this failing. In any case, thank you both!

gregcaporaso:
I decided to drop the column selection support from this PR for now as I was continuing to run into trouble with it as I developed unit tests and real-world examples and I don't want to hold this PR any more due to this. So, if you were previously doing something like:

```
md = use.init_metadata_from_url('md', metadata_url, column='body-site')
```

You should now do:

```python
column = 'body-site'
md = use.init_metadata_from_url('md', metadata_url)
mdc = use.get_metadata_column('mdc', column, md)
```

lizgehret:
These changes look great @gregcaporaso! Excited to have this available for usage examples moving forward. Unless @ebolyen has any change requests, this looks g2g 👍

Oddant1:
Maybe not directly related, but when I make dev this branch my QIIME 2 version is 2019.4.0+151.g1339294 which prevents me from running some of the usage examples in other repositories that use qiime2.__release__

lizgehret:
So after a `make dev` on my machine, this is the QIIME 2 version that I'm seeing:
```
QIIME 2 version: 2021.8.0.dev0+68.g1339294
```
@Oddant1 what QIIME 2 release is your conda env created from? I'm wondering if that is (at least part of) the inconsistency here.

Oddant1:
@lizgehret I figured it out and deleted my comment (probably shoulda just edited it). I just needed to do `git fetch --all --tags`

ebolyen (approved):
Thanks for the reviews everyone!

PR #660, opened 2022-10-24
Cache docs

Oddant1:
Merge didn't quite do what I wanted it to but eh, these just didn't make it into the last PR. I think I just thought I had pushed them but didn't.


ebolyen (approved):
Thanks @Oddant1! When I squash these, the merge will stop looking weird.

PR #663, opened 2022-11-08
NEW: Add a class method that removes the cache at a given path if …

Oddant1:
Adds functionality to the cache to support qiime tools cache_*


Oddant1:
Blocking qiime2/q2cli#278

ebolyen (approved):
LGTM, thanks @Oddant1!

PR #665, opened 2022-12-10
HACK: add short sleep for cold cache contention

ebolyen:
This is kind of dumb, but it seems to work.

Makes ridiculous commands like:
```
 for x in $(seq 1 6); do chmod -R a+x /tmp/qiime2; rm -rf /tmp/qiime2; for i in $(seq 1 11); do qiime diversity alpha --i-table ~/Downloads/feature-table.qza --p-metric shannon --o-alpha-diversity foobar-${i} & done; wait; done
```

work successfully on my machine

---
When multiple processes are creating a new cache for the first time, there is a race condition and no good option for creating a lock. This adds a small timeout so that the winner has a chance to create a consistent structure.

Worst case, it just blows up and everything should work fine the very next time it is used.

### DIFF
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -145,10 +145,23 @@ def _get_temp_path():
     # this if we are the owner of the /tmp/qiime2  directory or in other words
     # the first person to run QIIME 2 with this /tmp since the /tmp was wiped
     if not os.path.exists(cache_dir):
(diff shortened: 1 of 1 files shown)


Oddant1:
That's as good of a solution as any at least for now given that we can't properly lock it when the directory doesn't exist at all yet.

PR #666, opened 2022-12-14
SQUASH: Temporarily no op set_permissions

Oddant1:
Currently setting permissions has been causing us some pain on panfs, and it doesn't necessarily seem to be gaining us anything, so we are disabling it at least for now.


ebolyen (approved):
Thanks @Oddant1!

PR #668, opened 2022-12-15
SQUASH: Put validate on the cache called by cli

Oddant1:
Put validate method used by qiime tools cache-validate on the cache


Oddant1:
blocking qiime2/q2cli#283

PR #673, opened 2023-01-10
IMP: registered usage examples now enforce 10 MiB size limits

ebolyen:


ebolyen:
I think we can actually punt this to the next release instead of adding it to the patch, it doesn't do anything strictly necessary.

lizgehret:
Hey @ebolyen do you want this to be included in 2023.2 or should we punt to 2023.4?

ebolyen:
Oh I remember I wanted to add some tests for the measure keyword

ebolyen:
Let's bump this.

PR #674, opened 2023-01-24
SQUASH: Usage example fixed

Oddant1:
Fixes a usage example still using the moving-pictures metadata URL with the epoch interpolated.

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -29,7 +29,6 @@ requirements:
     - dill
     - psutil
     - flufl.lock
-    - tornado
 
 test:
   requires:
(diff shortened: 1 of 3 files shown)


ci/recipe/meta.yaml
```diff
@@ -33,6 +33,8 @@ requirements:
 test:
   requires:
     - pytest
+    - tornado
+    - notebook
```
ebolyen, line 37: Well, wasn't expecting that!

I think this is still a good thing, as it at least now proves the jupyter extension module is importable..., although they don't have any doctests lol.

ebolyen (approved):

PR #675, opened 2023-01-31
NEW: Start adding Collection

Oddant1:
Adds new Collection type to support collections of outputs

### DIFF
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -770,7 +770,7 @@ class Cache:
         """
         with self.lock:
             pool = self._create_collection_pool(ref_collection, key)
-            # self._register_key(key, ref_collection, pool=True, ordered=True)
+
             for ref in list(ref_collection.values()):
                 pool.save(ref)
 
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -62,17 +62,6 @@ cache: %s
 framework: %s
 """
 
-_KEY_TEMPLATE = """\
-origin:
- %s
-data:
- %s
-pool:
- %s
-order:
- %s
-"""
-
 # Thread local indicating the cache to use
 _CACHE = threading.local()
 _CACHE.cache = None
@@ -689,13 +678,17 @@ class Cache:
             for key in self.get_keys():
                 with open(self.keys / key) as fh:
                     loaded_key = yaml.safe_load(fh)
-                referenced_pools.add(loaded_key['pool'])
-                referenced_data.add(loaded_key['data'])
 
-            # Since each key has at most a pool or data, we will end up with a
-            # None in at least one of these sets. We don't want it
-            referenced_pools.discard(None)
-            referenced_data.discard(None)
+                if 'data' in loaded_key:
+                    referenced_data.add(loaded_key['data'])
+                elif 'pool' in loaded_key:
+                    referenced_pools.add(loaded_key['pool'])
+                # This really should never be happening unless someone messes
+                # with things manually
+                else:
+                    raise ValueError(f"The key '{key}' in the cache"
+                                     f"'{self.path}' does not point to"
+                                     " anything")
 
             # Walk over pools and remove any that were not referred to by keys
             # while tracking all data within those that were referenced
@@ -810,16 +803,24 @@ class Cache:
 
         key_fp = self.keys / key
 
-        if pool and collection is not None:
-            key_fp.write_text(_KEY_TEMPLATE % (
-                key, '', value, [str(v) for v in collection.values()]))
-        elif collection is not None:
-            raise ValueError('An ordered Collection key can only be made for a'
-                             ' pool.')
-        elif pool:
-            key_fp.write_text(_KEY_TEMPLATE % (key, '', value, ''))
+        key = {}
+        key['origin'] = key
+
+        if pool:
+            key['pool'] = value
+
+            if collection is not None:
+                key['order'] = \
+                    [{k: str(v.uuid)} for k, v in collection.items()]
         else:
-            key_fp.write_text(_KEY_TEMPLATE % (key, value, '', ''))
+            key['data'] = value
+
+            if collection is not None:
+                raise ValueError('An ordered Collection key can only be made'
+                                 ' for a pool.')
+
+        with open(key_fp, 'w') as fh:
+            yaml.dump(key, fh)
 
     def read_key(self, key):
         """Reads the contents of a given key.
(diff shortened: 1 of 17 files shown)


qiime2/core/testing/method.py
```diff
@@ -138,3 +138,11 @@ def type_match_list_and_set(ints: list, strs1: list, strs2: set) -> list:
 
 def union_inputs(ints1: Union[dict, list], ints2: list) -> list:
     return [0]
+
+
+def list_of_ints(ints: list) -> list:
```
ebolyen, line 143: Forgot about how IntSequence works. I think the type here for a collection of them would actually be:
`typing.List[list]` (you can access the inner fields by using `.__args__` -> `(list,)`)


Alternatively, we may want to disallow list and dict in transformers, although it will be a huge pain to rewrite these test actions, so if it can be done without that, I'm all for it.

Oddant1:
Blocking qiime2/q2cli#288

qiime2/core/cache.py
```diff
@@ -761,7 +762,19 @@ def save(self, ref, key):
...
-    def _register_key(self, key, value, pool=False):
+    def save_collection(self, ref_collection, key):
+        """Saves a Collection to a pool in the cache with the given key. This
+        pool's key file will keep track of the order of the Collection.
+        """
+        with self.lock:
+            pool = self._create_collection_pool(ref_collection, key)
+
+            for ref in list(ref_collection.values()):
+                pool.save(ref)
```
ebolyen, line 773: This could take a while, do we need to do something similar to `_rename_to_data` so that we aren't locking a cache for the duration it takes a collection to be (re-)written to a cache?

qiime2/core/cache.py
```diff
@@ -788,10 +801,24 @@ def _register_key(self, key, value, pool=False):
...
         else:
-            key_fp.write_text(_KEY_TEMPLATE % (key, value, ''))
+            key_dict['data'] = value
+
+            if collection is not None:
+                raise ValueError('An ordered Collection key can only be made'
+                                 ' for a pool.')
+
+        with open(key_fp, 'w') as fh:
+            yaml.safe_dump(key_dict, fh)
```
ebolyen, line 821: I was thinking we may want to set `default_flow_style=False`, but after testing it with a collection, it seems to work fine. :+1:

qiime2/core/cache.py
```diff
@@ -859,19 +887,45 @@ def load(self, key):
...
         """
         with self.lock:
-            try:
-                key_values = self.read_key(key)
-                path = self.data / key_values['data']
-            except TypeError as e:
+            key_values = self.read_key(key)
+
+            if 'order' in key_values:
+                return self.load_collection(key)
```
ebolyen, line 893: This must be for the CLI's benefit? I would generally expect an error, since the return type is dramatically different, but I see why it would be significantly simpler to just get _something_ out of the cache if it's real.

qiime2/core/type/collection.py
```diff
@@ -95,6 +109,28 @@ def validate_field(self, name, field):
...
+            return all(v in contained_expr for v in value.values())
+        # The Collection's default view type is dict, but it is also allowed to
+        # be passed around as a list
+        elif len(value) > 0:
+            return all(v in contained_expr for v in value)
+
+        return False
+
+    def __iter__(self):
+        return self._view.values()
```
ebolyen, line 130:
```suggestion
```

`__iter__` is very special to the type system, so I would avoid implementing it here

qiime2/core/type/signature.py
```diff
@@ -318,6 +318,155 @@ def _assert_valid_views(self, inputs, parameters, outputs):
                     " Pipelines do not support function annotations (found one"
                     " for parameter: %r)." % name)
 
+    def coerce_given_parameters(self, provenance, **user_input):
```
ebolyen, line 321: (Note to self to pick up from here tomorrow.)

ebolyen (changes requested):
Main thing I see so far is just the time that the cache may be locked when storing a collection (only really comes up in a cross-cache situation).

qiime2/core/type/signature.py
ebolyen, line 321: It would be nice if these layers did not have to consider provenance, that feels like something that should stay (be trapped) in `action.py` since provenance is quite mutable whereas things in `core.type` are quite functional.

In the case of parameters, you can pretty much do that without any change here, but the inputs and outputs seem trickier. So probably not worth addressing right now.

I really wish something like this was straight-forward:

```python
map(invocation.inputs, lambda name, spec, value: value._view(...)
```
where this "layer" basically just resolves the difference between lists/dicts/singular elements and transforms between them as needed. Or perhaps there's some dictionary of generic "Argument" objects which you can apply transformations to and it internally magics away the distinction.

ebolyen (changes requested):
It's mostly just some really bizarre edge-cases at this point. Nothing strictly blocks this PR, but we will need to circle back to fix at some point before release.

qiime2/core/cache.py
```diff
@@ -761,7 +762,19 @@ def save(self, ref, key):
...
 
-    def _register_key(self, key, value, pool=False):
+    def save_collection(self, ref_collection, key):
+        """Saves a Collection to a pool in the cache with the given key. This
+        pool's key file will keep track of the order of the Collection.
+        """
+        with self.lock:
+            pool = self._create_collection_pool(ref_collection, key)
+
+            for ref in list(ref_collection.values()):
```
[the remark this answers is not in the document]
Oddant1: Got so used to casting it as a list that I started doing it everywhere without thinking

qiime2/core/cache.py
Oddant1: Yes, this is for the CLI's benefit. I didn't really want it to work this way, but it makes the CLI so much simpler.

lizgehret:
This all _generally_ makes sense at a high level. I'll do another pass after you've addressed all of Evan's review comments!

qiime2/core/cache.py
Oddant1: pool.save is already pretty efficient, it creates a symlink in the pool then calls Cache._rename_to_data

qiime2/core/cache.py
Oddant1: Follow up on this, should save dispatch to save_collection or error? It would be nice to keep this behavior consistent. I can also always make it so load doesn't dispatch to load_collection, but IIRC that was looking like it was going to add a bit on annoyance on the CLI, but should be completely doable.

Oddant1: I'm honestly inclined to just kinda have save/load defer to save/load collection. It is nice to make people be explicit but idk it just feels like quality of life in this instance

PR #676, opened 2023-03-07
Pipeline resumption

Oddant1:
Allow for pipelines that failed mid-run to reuse all applicable results they created before failing when the user reruns them.

### DIFF
--- a/qiime2/sdk/context.py
+++ b/qiime2/sdk/context.py
@@ -16,6 +16,10 @@ class Context:
             self.cache = parent.cache
         else:
             self.cache = get_cache()
+            # Only ever do this on the root context. We only want to index the
+            # pool once before we start adding our own stuff to it.
+            if self.cache.named_pool is not None:
+                self.cache.named_pool.index_pool()
 
         self._parent = parent
         self._scope = None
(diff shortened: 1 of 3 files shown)


qiime2/sdk/context.py
```diff
@@ -36,13 +47,62 @@ def get_action(self, plugin: str, action: str):
...
+                hashable_arguments = make_hashable(hashable_arguments)
+                invocation = IndexedInvocation(
+                    plugin_action, hashable_arguments)
+
+                if invocation in self.cache.named_pool.index:
+                    outputs = {}
+                    for key in action_obj.signature.outputs.keys():
+                        output = self.cache.named_pool.index[invocation][key]
+                        outputs[key] = \
+                            qiime2.sdk.Result.load(self.cache.data / output)
```
ebolyen, line 91: This should be `self.cache.named_pool.load()` right? That should handle collections correctly

Oddant1:
blocking qiime2/q2cli#293

ebolyen (changes requested):
This all looks good, and the tests are nice. I only see a few issues that need to be addressed.

ebolyen (approved):

PR #679, opened 2023-04-04
rename test file to be consistent with artifact version being tested

colinvwood:


ebolyen (approved):
lol, good catch, thanks @colinvwood!

PR #681, opened 2023-05-04
SQUASH: Rework config loading

Oddant1:
Loading configs in the way we initially envisioned is a bit more complex than we initially hoped. This got grosser than I intended. Refactors/unit tests will come tomorrow. Fortunately, we should not need this to start messing with Parsl.


Oddant1:
This now also contains the changes from parsl to parallel in the public API

ebolyen (approved):
LGTM, thanks @Oddant1!

PR #688, opened 2023-05-31
Support de facto collections of proxy outputs

### LINKED ISSUE #687: ResultCollections of ProxyArtifacts don't work when returned from a pipeline
**Bug Description**
It is not possible to return a collection (list, dict, or otherwise) of ProxyArtifact's as this situation is missed by the final stages of the pipeline executor. The result is an error indicating that `ProxyArtifact` is not in instance of `Result`.

**Workaround**
```python
def some_pipeline(ctx, ...):
    # stuff
    ...
    # before returning the collection:
    for key, value in results.items():
        results[key] = value.result()
    return results
```

Oddant1:
Closes #687 by adding support for de facto collections of proxy results


qiime2/core/testing/plugin.py
```diff
@@ -882,6 +882,15 @@ def factory():
     description='Takes a collection and returns a collection'
 )
 
+dummy_plugin.pipelines.register_function(
+    function=de_facto_collection_pipeline,
+    inputs={},
+    parameters={},
+    outputs=[('output', Collection[Mapping])],
+    name='Returns de facto ResultCollection',
+    description='Takes noting and returns de facto ResultCollection'
```
lizgehret, line 891:
```suggestion
    description='Takes nothing and returns de facto ResultCollection'
```

lizgehret (commented):
all looks reasonable - do the tests pass locally @Oddant1?

Oddant1:
@lizgehret yes

lizgehret (approved):

PR #690, opened 2023-06-28
Ensure input collections not proxies

Oddant1:
Handle the case where we have a de-facto collection of proxy inputs. Basically the case where in a pipeline you create a list, append proxies to the list, then pass that list into an action.


Oddant1:
Unit tests should be good now

qiime2/core/testing/method.py
```diff
@@ -150,6 +150,10 @@ def dict_of_ints(ints: int) -> int:
     return ints
 
 
+def returns_int(int: int) -> int:
```
lizgehret, line 153: ⭐

qiime2/core/testing/method.py
```diff
@@ -150,6 +150,10 @@ def dict_of_ints(ints: int) -> int:
...
+    return int
+
```
lizgehret, line 155: ⭐

qiime2/core/testing/pipeline.py
```diff
@@ -195,6 +195,48 @@ def internal_fail_pipeline(ctx, ints1, ints2, int1, string, fail=False):
...
+    returns_int = ctx.get_action('dummy_plugin', 'returns_int')
+    list_of_ints = ctx.get_action('dummy_plugin', 'list_of_ints')
+    num_ints = 3
+
+    ints = []
+    for i in range(num_ints):
+        ints_ret, = returns_int(i)
+        ints.append(ints_ret)
+
+    if non_proxies:
```
lizgehret, line 208: Just for my own understanding, why do you add 1 to `num_ints`?

qiime2/core/testing/pipeline.py
```diff
@@ -195,6 +195,48 @@ def internal_fail_pipeline(ctx, ints1, ints2, int1, string, fail=False):
...
+    dict_of_ints = ctx.get_action('dummy_plugin', 'dict_of_ints')
+    num_ints = 3
+
+    ints = {}
+    for i in range(num_ints):
+        ints_ret, = returns_int(i)
+        ints[i + 1] = ints_ret
+
+    if non_proxies:
+        ints[num_ints + 2] = ctx.make_artifact(SingleInt, num_ints + 1)
```
lizgehret, line 230: Same question as above, but also why you're adding 2 to this index

qiime2/sdk/action.py
```diff
@@ -84,6 +76,83 @@ def _run_parsl_action(action, ctx, execution_ctx, args, kwargs, inputs=[]):
         return results
 
 
+def _map_arg(arg, futures):
```
lizgehret, line 79: It looks like this was also just refactored from code that was removed on line 425 - so is it then the `_is_all_proxies()` helper that's ensuring that proxies are mapped to input collections?

lizgehret (commented):
just a few questions for my own understanding, but it all seems pretty reasonable and tests seem comprehensive 🤷‍♀️

qiime2/core/testing/pipeline.py
Oddant1: The +1 here is from a previous implementation, and I'm actually going to get rid of it. The way I was doing things before, the ints list would have been [1,2,3] so I needed the +1 to append a 4 instead of another 3. Not like it would break anything to append another 3, I just wanted a 4. Now though this will make [0,1,2,4] which is lame.

qiime2/core/testing/pipeline.py
Oddant1: Similar deal here to the above, but I am going to leave a +1 (changed from +2) in the key so the dict is {'1':0, '2':1, '3':2}. Early on in writing this stuff I tried to avoid having the keys in one of these dicts be monotonically increasing from 0. I avoided that because if things went wrong somehow and the dict turned into a list then was re-keyed it would be given keys that were monotonically increasing from 0. Any other set of keys in the output lets me know that it retained its keys the whole way through.

qiime2/sdk/action.py
```diff
@@ -48,31 +49,22 @@ def _run_parsl_action(action, ctx, execution_ctx, args, kwargs, inputs=[]):
...
-    for arg in args:
-        if isinstance(arg, Proxy):
-            # Same as above with the hackiness
-            resolved_result = inputs[arg._future_]
-            remapped_args.append(arg._get_element_(resolved_result))
-        else:
-            remapped_args.append(arg)
+    args = []
+    for arg in mapped_args:
+        unmapped = _unmap_arg(arg, inputs)
```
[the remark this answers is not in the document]
Oddant1: The function `map_arg` was factored out of `_bind_parsl` and the function `_unmap_arg` was factored out of `_run_parsl_action`. I'll talk about unmap here, and I will add a comment talking about map. Read the map explanation first.

lizgehret:
lgtm, merging. thanks @Oddant1!

PR #691, opened 2023-07-05
NEW: adds ResultCollection to Usage API

lizgehret:
Adds ResultCollections to Usage API with the following steps (notes for myself while developing):

1. Change base Usage class to accept Collections
2. Implement in ExecutionUsage class
3. Use dummy plugin to test that this works

Co-Authored by: @Oddant1

### DIFF
--- a/qiime2/sdk/usage.py
+++ b/qiime2/sdk/usage.py
@@ -888,9 +888,9 @@ class Usage:
         ...          'Bar': qiime2.Artifact.import_data('IntSequence1', [4, 5, 6])})
         ...     return a
         ...
-        >>> my_collection = use.init_result_collection('my_collection', factory)
-        >>> my_collection
-        <ExecutionUsageVariable name='my_collection', var_type='result_collection'>
+        >>> int_collection = use.init_result_collection('int_collection', factory)
+        >>> int_collection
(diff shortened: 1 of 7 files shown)


lizgehret:
Accompanying CLI PR: [#302](https://github.com/qiime2/q2cli/pull/302)

qiime2/sdk/usage.py
```diff
@@ -1608,6 +1664,12 @@ def init_artifact(self, name, factory):
 
         return variable
 
+    def init_result(self, name, factory):
```
ebolyen, line 1667: This seems to be the wrong name

ebolyen (commented):
This all looks pretty on the level, but we will need a way to assert something about a given `key` in a collection, which will be important for q2-mystery-stew

ebolyen:
For the assertion, I believe we discussed just adding a `key=...` parameter to the assertion methods, that way we don't have to invent a de-referencing operation in the API itself. But if you guys want to make a de-reference operation (turning a `result_collection` variable into an `artifact` variable), I'm not opposed.

ebolyen (approved):
Need to pull this down, but it looks very reasonable. Nice work!

lizgehret (commented):
recent additions look good!

PR #692, opened 2023-07-06
Socket leaks

Oddant1:
Fixed an issue with not closing opened sockets.

Made it so if QIIME 2 is being run outside of a conda env (usually a Docker thing) we don't try to get a conda_prefix and use that in a filepath.

NOTE: Breaking API changes

Removed `get_mapping` function. `get_config_from_fp` now returns a tuple of a parsl.Config object and a mapping dict.

Now NEED to with in a ParallelConfig before executing in parallel

### DIFF
--- a/qiime2/sdk/parallel_config.py
+++ b/qiime2/sdk/parallel_config.py
@@ -58,7 +58,7 @@ def setup_parallel(config_fp=None):
     path or looks through several default paths if no path is provided and
     loads a vendored config as a last resort
     """
-    config = PARALLEL_CONFIG.parallel_config
+    parallel_config = PARALLEL_CONFIG.parallel_config
     mapping = PARALLEL_CONFIG.action_executor_mapping
 
     # If we don't have a filepath or a currently existing config then get the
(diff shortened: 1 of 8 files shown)


qiime2/sdk/parallel_config.py
```diff
@@ -140,11 +156,11 @@ def _get_vendored_config():
             config_fp = fp_
         # 4. Check in conda env
         # ~/miniconda3/env/{env_name}/conf
-        elif os.path.exists(fp_ := VENDORED_FP):
+        elif VENDORED_FP is not None and os.path.exists(fp_ := VENDORED_FP):
             config_fp = fp_
         # 5. Write the vendored config to the vendored location and use
         # that
-        else:
+        elif VENDORED_FP is not None:
```
Oddant1, line 163: What do we actually want to do in this case?

qiime2/sdk/parallel_config.py
```diff
@@ -69,38 +52,49 @@ def setup_parallel(config_fp=None):
...
     # they are doing things wrong (probably forgot to resolve their future
     # inside of their context manager).
-    if PARALLEL_CONFIG.parallel_config != config:
+    if PARALLEL_CONFIG.parallel_config != parallel_config:
+        # If a config was already loaded, clean up the DFK and clear it for a
+        # new one. If there wasn't already a config loaded, we get an error
+        # which we except and ignore
+        try:
+            dfk = parsl.dfk()
+            dfk.cleanup()
```
Oddant1, line 76: This has mitigated the socket leak issue, but I am still getting a "too many open files" error on OSX on the Mac Mini

Oddant1:
I did some testing on the mac mini. On the master branch running the framework tests uses 62 threads and 105 ports. On this branch, they use 46 threads and 85 threads.

HOWEVER

When I run `ulimit -n` in a free standing terminal on the mac mini, I get a file-descriptor limit of 256. When I do the same in the VSCode terminal, I get a file-descriptor limit of 1048575. . .  @colinvwood saw similar results. The tests pass on both branches when I run `make test` in the VSCode terminal, and the tests fail on both branches when I run them in a free standing terminal and again same for Colin.

So the issue here appears to be largely related to the way OSX sets max open files limits for a given process. That being said, I would like for the tests to still open fewer file descriptors and function properly.

Oddant1:
At this point we are definitely telling the dfk to cleanup every time we are done with it, but loading a config with a HighThroughputExecutor is still just leaking open sockets all over the place. dfk.cleanup does not circumvent that.

Oddant1:
Blocking qiime2/dev-docs#61 and qiime2/q2cli#303

PR #693, opened 2023-07-10
Add __contains__ to ResultCollection

Oddant1:
Add __contains__ method to ResultCollection


Oddant1:
I found a few more things to clean up here ,so I am going to open a new PR with a branch name that better indicates what's going on

PR #696, opened 2023-07-13
No conda

Oddant1:
Handle the case where we try to parallelize things when running QIIME 2 outside of a conda environment (usually involves Docker). Previously we were reliant on the conda_prefix to get a path to the vendored config. Now if there is no conda_prefix we just load the vendored config directly from memory.

### DIFF
--- a/qiime2/sdk/parallel_config.py
+++ b/qiime2/sdk/parallel_config.py
@@ -82,7 +82,6 @@ def setup_parallel(config_fp=None):
         # already have a loaded config
         if config_dict == {} and PARALLEL_CONFIG.parallel_config is None:
             config_fp = _get_vendored_config()
-<<<<<<< HEAD
             config_dict = get_config(config_fp)
 
         # Now if we actually have a config dict, we want to load the config. We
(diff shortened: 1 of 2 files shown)


qiime2/sdk/parallel_config.py
```diff
@@ -138,13 +147,14 @@ def _get_vendored_config():
...
         # 4. Check in conda env
         # ~/miniconda3/env/{env_name}/conf
-        elif os.path.exists(fp_ := VENDORED_FP):
+        elif CONDA_PREFIX != '' and os.path.exists(fp_ := VENDORED_FP):
             config_fp = fp_
         # 5. Write the vendored config to the vendored location and use
         # that
-        else:
+        elif CONDA_PREFIX != '':
             with open(VENDORED_FP, 'w') as fh:
```
colinvwood, line 158: VENDORED_FP will end up being a relative path here `etc/qiime2_config.toml`. Just making sure that's intended.

Oddant1: If they are in a conda environment, the conda prefix will be the absolute path to the base of the env  (ex: `/home/anthony/miniconda3/envs/q2-dev`). If we append `etc/qiime2_config.toml` to this we have an absolute path to a specific file. If they are not in a conda environment, there will not be a conda prefix, and we will not try to access this path at all.

colinvwood: Misread not equal as equal late on Friday

colinvwood (approved):

PR #697, opened 2023-07-13
Socket leaks2

Oddant1:


qiime2/sdk/parallel_config.py
```diff
@@ -224,3 +248,21 @@ def __exit__(self, *args):
...
+
+
+def _check_env(cls):
+    if 'QIIMETEST' not in os.environ:
+        raise ValueError(
+            f"Do not instantiate the class '{cls}' when not testing")
+
+
+class _TEST_EXECUTOR_(parsl.executors.threads.ThreadPoolExecutor):
+    """We needed multiple kinds of executor to ensure we were mapping things
```
lizgehret, line 260: 👍

qiime2/sdk/parallel_config.py
```diff
@@ -13,14 +13,35 @@
...
+import pkg_resources
 
 from parsl.config import Config
 
-
 PARALLEL_CONFIG = threading.local()
 PARALLEL_CONFIG.parallel_config = None
 PARALLEL_CONFIG.action_executor_mapping = {}
 
+# We write a default config to a location in the conda env if there is an
```
lizgehret, line 24: Looks like these are the same changes as #696 - should this be separated?

Oddant1:
Those mentions are actually irrelevant, meant to mention 695 not this one which is 697

qiime2/sdk/parallel_config.py
Oddant1: Good catch. I'll remove the comment and let it come from the other PR because that functionality isn't addressed here. The way the module scope variables are created here is fine regardless though

colinvwood (approved):

qiime2/core/type/signature.py
```diff
@@ -367,6 +367,7 @@ def _coerce_given_parameter(self, param, spec):
         """ Coerce parameter to be appropriate for callable
         """
         view_type = spec.view_type
+        # raise ValueError(f'{view_type}\n{param}')
```
lizgehret, line 370: Is this a stray debug statement?

qiime2/sdk/parallel_config.py
```diff
@@ -242,6 +261,18 @@ def _check_env(cls):
             f"Do not instantiate the class '{cls}' when not testing")
 
 
+class _TEST_EXECUTOR_(parsl.executors.threads.ThreadPoolExecutor):
+    """We needed multiple kinds of executor to ensure we were mapping things
```
lizgehret, line 265: This is a really helpful comment 👍

lizgehret (commented):
generally lgtm - same failure as #688, do these tests also pass locally @Oddant1?

qiime2/core/type/signature.py
Oddant1: yes

lizgehret:
There isn't a direct way to test this locally using `netstat` on Mac (like there is on Linux) - but when the framework tests are run without the updates from this branch, the test runner crashes, and CPU usage is at 81% (screenshots shown below).
Screen Shot 2023-08-07 at 11 04 07 AM
Screen Shot 2023-08-07 at 11 03 57 AM

Here are the stats and CPU usage when this feature branch is checked out and tests are run locally (which all pass, and the test runner doesn't crash).
Screen Shot 2023-08-07 at 11 06 03 AM

PR #698, opened 2023-07-19
Change error message

### LINKED ISSUE #514: Importing non-existent semantic type complains about missing directory format
**Bug Description**
When importing, if a user specifies a non-existent semantic type, QIIME 2 doesn't tell them that the type doesn't exist, instead it complains about the type not having a compatible directory format.

```python
import qiime2
# Peanut[TheDog] is a made-up semantic type
qiime2.Artifact.import_data('Peanut[TheDog]', None)
...
TypeError: Semantic type Peanut[TheDog] does not have a compatible directory format.
```

**Expected behavior**
I think seeing a message along the lines of "No such semantic type as Peanut[TheDog] registered in this deployment of QIIME 2" would be a good start.

**Questions**
1. `qiime2.sdk.parse_type` is implicated in this --- should this util perform this validation/verification, or should it stay in `import_data` (or be elsewhere altogether)?

**Comments**
1. This problem is probably limited to lazy evaluation of the type (specifying the string value`'Foo'` vs the type object `Foo`).

**References**
1. [Original forum post](https://forum.qiime2.org/t/help-importing-a-concatenated-fastq-file-as-multiplexed-paired-end-barcode-in-sequence/13035)

Oddant1:
Closes #514 by making the error message say the semantic type is not registered instead of saying it does not have a registered format.

Here's the thing about this. This error is being raised when a semantic type does not exist AND when a semantic type does not have a directory format. Currently, there does not seem to be a good way to tell if a semantic type has been registered without a corresponding format.

In other words, the previous error message was lying when the type itself wasn't registered. This one is lying when the type is registered but has no format. There doesn't seem to currently be a good way to distinguish between the two situations. I believe this error message will generally be less confusing in the scenarios where it is likely to be encountered.

### DIFF
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -381,5 +381,5 @@ class PluginManager:
                 return artifact_class_record.format
 
         raise TypeError(
-            "Semantic type %r does not have a compatible directory format."
+            "Semantic type %r is not a registered semantic type."
             % semantic_type)
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -380,6 +380,11 @@ class PluginManager:
             if semantic_type <= artifact_class_record.semantic_type:
                 return artifact_class_record.format
 
+        # TODO: We need a good way to tell if a semantic type is registered but
+        # does not have a directory format. The previous error was causing a
+        # lot of confusion, but this one is also not entirely accurate.
+        #
+        # Reference: https://github.com/qiime2/qiime2/issues/514
         raise TypeError(
             "Semantic type %r is not a registered semantic type."
             % semantic_type)
--- a/qiime2/sdk/plugin_manager.py
+++ b/qiime2/sdk/plugin_manager.py
@@ -382,7 +382,7 @@ class PluginManager:
 
         # TODO: We need a good way to tell if a semantic type is registered but
         # does not have a directory format. The previous error was causing a
-        # lot of confusion, but this one is also not entirely accurate.
+        # lot of confusion.
         #
         # Reference: https://github.com/qiime2/qiime2/issues/514
         raise TypeError(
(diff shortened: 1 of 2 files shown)


qiime2/sdk/plugin_manager.py
```diff
@@ -380,6 +380,11 @@ def get_directory_format(self, semantic_type):
...
                 return artifact_class_record.format
 
+        # TODO: We need a good way to tell if a semantic type is registered but
+        # does not have a directory format. The previous error was causing a
+        # lot of confusion, but this one is also not entirely accurate.
+        #
+        # Reference: https://github.com/qiime2/qiime2/issues/514
         raise TypeError(
-            "Semantic type %r does not have a compatible directory format."
+            "Semantic type %r is not a registered semantic type."
```
gregcaporaso, line 389:
```suggestion
            ("Semantic type %r is invalid, either because it doesn't have "
             "a compatible directory format, or because it's not "
             "registered.")
```

gregcaporaso (commented):
@Oddant1, how about this error message? This at least provides two paths for the developer to follow (I believe this will only ever be encountered by developers, unless a really poorly tested plugin is released into the wild).

Oddant1:
@gregcaporaso I pushed a commit changing the error message, but I'm not seeing it here for some reason.

qiime2/sdk/plugin_manager.py
```diff
@@ -380,6 +380,12 @@ def get_directory_format(self, semantic_type):
             if semantic_type <= artifact_class_record.semantic_type:
                 return artifact_class_record.format
 
+        # TODO: We need a good way to tell if a semantic type is registered but
+        # does not have a directory format. The previous error was causing a
+        # lot of confusion, but this one is also not entirely accurate.
```
gregcaporaso, line 385:
```suggestion
        # lot of confusion.
```

I think it is accurate now, right?

gregcaporaso (changes requested):
This looks good, just one suggestion. Since the suggestion is only on a comment I don't need to see this again before it's merged.

PR #700, opened 2023-07-27
Validate result collection keys

### LINKED ISSUE #699: restrict allowed characters in `Collection` keys
Because `Collection` keys are often used in filenames, we should restrict the allowed characters to avoid characters that are can be problematic in file paths. Our initial set of allowed characters (which may be expanded in the future if needed) will be we will allow `A-Z`, `a-z`, `0-9`, `+`, `-`, `_`, and `.`. 

If a value that will be used as a key contains any of these characters, the actions should fail with an error as early as possible. Given that we will ideally fail before work is done, it may make sense to have a public validate function that takes a list of keys and fails with an informative error message if any disallowed characters are present (e.g., a static `Collection.validate_keys` method). This could then be called when the keys are first defined. 

I initially implemented this as part of https://github.com/qiime2/q2-feature-table/pull/277 for an action that returns a `Collection` - see [this commit](https://github.com/gregcaporaso/q2-feature-table/commit/c23848820ea16faf270a941c6cb87e30b4ed9899).

Oddant1:
Ensure that result collection keys only contain the following characters: `A-Z`, `a-z`, `0-9`, `+`, `-`, `_`, and `.`.

Close #699

### DIFF
--- a/qiime2/sdk/tests/test_result.py
+++ b/qiime2/sdk/tests/test_result.py
@@ -591,7 +591,7 @@ class TestResultCollection(unittest.TestCase):
         with self.assertRaisesRegex(
                 KeyError,
                 'ResultCollection keys may only contain the following'
-                ' characters:.*'):
+                ' characters:.*valid key'):
             ResultCollection({'not a valid key': 0})
 
     def test_invalid_key_added(self):
@@ -600,7 +600,7 @@ class TestResultCollection(unittest.TestCase):
         with self.assertRaisesRegex(
                 KeyError,
                 'ResultCollection keys may only contain the following'
-                ' characters:.*'):
+                ' characters:.*valid key'):
             collection['not a valid key'] = 0
 
 
--- a/qiime2/sdk/tests/test_result.py
+++ b/qiime2/sdk/tests/test_result.py
@@ -584,14 +584,15 @@ class TestResultCollection(unittest.TestCase):
 
     def test_collection_non_str_keys(self):
         with self.assertRaisesRegex(
-                TypeError, 'All ResultCollection keys must be strings'):
+                KeyError, 'ResultCollection keys must be strings and may only '
+                'contain the following characters:.*1'):
             ResultCollection({1: 0})
 
     def test_invalid_key_init(self):
         with self.assertRaisesRegex(
                 KeyError,
-                'ResultCollection keys may only contain the following'
-                ' characters:.*valid key'):
+                'ResultCollection keys must be strings and may only contain '
+                'the following characters:.*valid key'):
             ResultCollection({'not a valid key': 0})
 
     def test_invalid_key_added(self):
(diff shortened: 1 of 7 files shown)


qiime2/sdk/util.py
```diff
@@ -118,3 +120,9 @@ def actions_by_input_type(string):
                 commands.append((pg, actions))
 
     return commands
+
+
+def validate_result_collection_key(key):
+    if not bool(re.match('^[a-zA-Z0-9+-_.]+$', key)):
+        raise KeyError('ResultCollection keys may only contain the following'
+                       ' characters: A-Z, a-z, 0-9, +, -, _, and .')
```
gregcaporaso, line 128:
```suggestion
                       ' characters: A-Z, a-z, 0-9, +, -, _, and .\n'
                       'The following key contains disallowed '
                       f'characters: {key}')
```

qiime2/sdk/tests/test_result.py
```diff
@@ -587,6 +587,22 @@ def test_collection_non_str_keys(self):
                 TypeError, 'All ResultCollection keys must be strings'):
             ResultCollection({1: 0})
 
+    def test_invalid_key_init(self):
+        with self.assertRaisesRegex(
+                KeyError,
+                'ResultCollection keys may only contain the following'
+                ' characters:.*'):
```
gregcaporaso, line 594:
```suggestion
                ' characters:.*valid key'):
```

qiime2/sdk/tests/test_result.py
```diff
@@ -587,6 +587,22 @@ def test_collection_non_str_keys(self):
...
+            ResultCollection({'not a valid key': 0})
+
+    def test_invalid_key_added(self):
+        collection = ResultCollection()
+
+        with self.assertRaisesRegex(
+                KeyError,
+                'ResultCollection keys may only contain the following'
+                ' characters:.*'):
```
gregcaporaso, line 603:
```suggestion
                ' characters:.*valid key'):
```

qiime2/sdk/util.py
```diff
@@ -118,3 +120,9 @@ def actions_by_input_type(string):
                 commands.append((pg, actions))
 
     return commands
+
+
+def validate_result_collection_key(key):
```
gregcaporaso, line 125: I recommend adding a couple of unit tests for this function itself, especially since it's public facing the in sdk.

gregcaporaso (changes requested):
This looks great @Oddant1 - just a couple of quick suggestions.

gregcaporaso (approved):
@Oddant1, I ended up pushing a couple of commits while working on https://github.com/qiime2/q2-feature-table/pull/277. This all looks good to me, but you should review my changes.

lizgehret (commented):
this lgtm @Oddant1 - once merge conflicts are resolved, i can get this one merged!

gregcaporaso:
I'm not sure why that flake8 issue came up here - should I move that commit over to a separate pull request?

lizgehret (approved):

PR #707, opened 2023-08-24
IMP: rename init_result_collection -> init_artifact_collection

lizgehret:
This PR renames init_result_collection -> init_artifact_collection in anticipation of visualization collections being utilized in the future and needing a separate init method when that time comes.

### DIFF
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -243,7 +243,7 @@ def optional_inputs(use):
 
 
 def collection_list_of_ints(use):
-    ints = use.init_result_collection('ints', artifact_collection_factory)
+    ints = use.init_artifact_collection('ints', artifact_collection_factory)
 
     out, = use.action(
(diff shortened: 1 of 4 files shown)


Oddant1:
I wonder if it would be a better approach to parameterize `init_result_collection` with `artifact` and `visualization` or to just implement `init_visualization_collection` now. Just thinking about this now if/when someone wants visualization collections as inputs in the future it might be a good idea to just already have them here, but that might actually be some of @colinvwood's bloatware lol because I'm not sure if that's a scenario that will ever actually happen. @ebolyen what do you think about this? I'm inclined to say it's a good idea to just cover our bases now, but I'm not sure if that's even a real base.

Oddant1:
I think I'm good with doing it this way for now. Keeping `init_artifact_collection` and `init_visualization_collection` seperate might be excessive or even BLOATWARE as @colinvwood would put it, but it would make things clean.

PR #708, opened 2023-08-24
IMP: adds support for kebab-case cache keys

### LINKED ISSUE #667: support `kabob-case` for cache entry identifiers
At present cache entry identifiers (e.g., the `test_table` in `/projects/microbiome/jgc-cache/:test_table`) must be valid Python identifiers. It would be nice if this restriction were loosened to support kabob case (e.g.,`test-table`) just for convenience (I tend to prefer kabob case over snake case when possible).

Oddant1:
Closes qiime2/qiime2#667 by adding support for kebab cased cache keys

### DIFF
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -805,11 +805,15 @@ class Cache:
             this to ensure no one creates keys that cause issues when we try to
             load them.
         """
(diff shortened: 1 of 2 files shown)


lizgehret (approved):
lgtm!

PR #709, opened 2023-08-24
Ensure filepath is str in Result.save

### LINKED ISSUE #704: Support saving via pathlib.Path in Artifact API
**Improvement Description**

I think it would be useful if users could save Artifacts to disk locations specified via `pathlib.Path` objects instead of only strings.

**Current Behavior**

Currently, attempting to save an Artifact to a `Path` object results in the following error:

```
AttributeError: 'PosixPath' object has no attribute 'rstrip'
```

arising from the following line: https://github.com/qiime2/qiime2/blob/34db42329029de3a0030c937beecef3382dc927f/qiime2/sdk/result.py#L225

because there is no `rstrip` method associated with `PosixPath`

**Proposed Behavior**

I propose that some transformation be done to allow interfacing with `pathlib` objects such that Artifacts are saved correctly.

**Reproducible Example**

```python
from pathlib import Path
from qiime2 import Artifact

art = Artifact.load("./table.qza")  # Can be any artifact

outfile_str = "table.str.qza"
outfile_plib = Path("table.plib.qza")

art.save(outfile_str)
print(f"===== Saved table to {outfile_str} =====")

art.save(outfile_plib)
print(f"===== Saved table to {outfile_plib} =====")
```

Output:

```
===== Saved table to table.str.qza =====
Traceback (most recent call last):
  File "script.py", line 12, in <module>
    art.save(outfile_plib)
  File "/home/ec2-user/mambaforge/envs/qiime2-2023.5/lib/python3.8/site-packages/qiime2/
sdk/result.py", line 225, in save
    filepath = filepath.rstrip('.')
AttributeError: 'PosixPath' object has no attribute 'rstrip'
```

---

See https://github.com/biocore/biom-format/issues/894 for a similar issue on the BIOM data structure.

Using Q2-2023.5 environment

Oddant1:
Closes qiime2/qiime2#704 by casting the filepath to a string before trying to do string manipulations to it.

### DIFF
--- a/qiime2/sdk/result.py
+++ b/qiime2/sdk/result.py
@@ -222,6 +222,8 @@ class Result:
 
         # This accounts for edge cases in the filename extension
         # and ensures that there is only a single period in the ext.
+        # Caste to str incase we received a pathlib.Path or similar
+        filepath = str(filepath)
         filepath = filepath.rstrip('.')
         ext = '.' + ext.lstrip('.')
 
--- a/qiime2/sdk/tests/test_artifact.py
+++ b/qiime2/sdk/tests/test_artifact.py
@@ -256,6 +256,26 @@ class TestArtifact(unittest.TestCase, ArchiveTestingMixin):
         self.assertEqual(artifact1.view(list),
                          artifact2.view(list))
 
+    def test_roundtrip_pathlib(self):
+        fp1 = pathlib.Path(os.path.join(self.test_dir.name, 'artifact1.qza'))
+        fp2 = pathlib.Path(os.path.join(self.test_dir.name, 'artifact2.qza'))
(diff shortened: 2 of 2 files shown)


lizgehret (approved):
lgtm!

lizgehret:
Oh actually, maybe it would be good to add a test for this @Oddant1

lizgehret (approved):
this looks good, thanks for adding that test!

ebolyen:
Hate to mention this after the fact, but there's a [protocol for paths](https://peps.python.org/pep-0519/) as an alternative to calling `str()`. i.e. you can call [`os.fpath()`](https://docs.python.org/3/library/os.html#os.fspath)

PR #710, opened 2023-08-28
Add a method that starts with _

Oddant1:
Adds a method to the `dummy_plugin` to be used in testing by qiime2/q2cli#306


Oddant1:
blocking qiime2/q2cli#306

colinvwood (approved):

PR #723, opened 2023-10-10
Make TypeMap aware of ProxyArtifact

Oddant1:
Makes TypeMap aware of ProxyArtifacts so pipelines with TypeMaps work in parallel. Blocking bokulich-lab/q2-moshpit#62

### DIFF
--- a/qiime2/core/type/signature.py
+++ b/qiime2/core/type/signature.py
@@ -619,7 +619,8 @@ class PipelineSignature:
             inner = UnionExp(
                 (self._infer_type(key, v) for v in value.values()))
             return Collection[inner.normalize()]
-        if isinstance(value, qiime2.sdk.Artifact):
+        if isinstance(value, qiime2.sdk.Artifact) or \
+                isinstance(value, qiime2.sdk.proxy.ProxyArtifact):
             return value.type
         else:
             return infer_primitive_type(value)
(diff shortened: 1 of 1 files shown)


qiime2/core/type/signature.py
```diff
@@ -619,7 +619,8 @@ def _infer_type(self, key, value):
             inner = UnionExp(
                 (self._infer_type(key, v) for v in value.values()))
             return Collection[inner.normalize()]
-        if isinstance(value, qiime2.sdk.Artifact):
+        if isinstance(value, qiime2.sdk.Artifact) or \
```
colinvwood, line 622: I think you can do 
`isinstance(value, (qiime2.sdk.Artifact, qiime2.sdk.proxy.ProxyArtifact))`

qiime2/core/type/signature.py
```diff
@@ -619,7 +619,8 @@ def _infer_type(self, key, value):
...
+                isinstance(value, qiime2.sdk.proxy.ProxyArtifact):
```
colinvwood:
```suggestion
        if isinstance(value, (qiime2.sdk.Artifact, qiime2.sdk.proxy.ProxyArtifact)):
```

colinvwood (approved):

PR #726, opened 2023-10-20
Add support for result collections to usage api

colinvwood:
Add support to the base usage and artifact api usage classes for constructing and accessing members from result collections.

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -30,7 +30,6 @@ requirements:
     # importlib_resources 6.1.0 which is what we were using when this was
     # pinned. Pinning this to the version that was working for us was simpler
     # than digging around in importlib_resources to find a compatible version
-    # there
     - networkx =3.1
     - dill
     - psutil
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -25,6 +25,12 @@ requirements:
     - tzlocal <3
     - python-dateutil
     - bibtexparser
+    # This is pinned because networkx 3.2 was trying to use
+    # importlib.resources.files which apparently doesn't exist on
+    # importlib_resources 6.1.0 which is what we were using when this was
+    # pinned. Pinning this to the version that was working for us was simpler
+    # than digging around in importlib_resources to find a compatible version
+    # there
     - networkx =3.1
     - dill
     - psutil
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -25,7 +25,7 @@ requirements:
     - tzlocal <3
     - python-dateutil
     - bibtexparser
-    - networkx
+    - networkx =3.1
     - dill
     - psutil
     - flufl.lock
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -264,3 +264,26 @@ def collection_dict_of_ints(use):
     )
 
     out.assert_output_type(semantic_type='SingleInt', key='Foo')
+
+
+def construct_and_access_collection(use):
+    ints_a = use.init_artifact('ints_a', single_int1_factory)
+    ints_b = use.init_artifact('ints_b', single_int2_factory)
+
+    rc_in = use.construct_collection(
+        'rc_in', 'artifact', {'a': ints_a, 'b': ints_b}
+    )
+
+    rc_out, = use.action(
+        use.UsageAction(plugin_id='dummy_plugin', action_id='dict_of_ints'),
+        use.UsageInputs(ints=rc_in),
+        use.UsageOutputNames(output='rc_out')
+    )
+
+    ints_b = use.access_collection_member('ints_b', rc_out, 'b')
+
+    same_int = use.action(  # noqa: F841
+        use.UsageAction(plugin_id='dummy_plugin', action_id='returns_int'),
+        use.UsageInputs(int=ints_b),
+        use.UsageOutputNames(output='same_int')
+    )
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -270,8 +270,8 @@ def construct_and_access_collection(use):
     ints_a = use.init_artifact('ints_a', single_int1_factory)
     ints_b = use.init_artifact('ints_b', single_int2_factory)
 
-    rc_in = use.construct_collection(
-        'rc_in', 'artifact', {'a': ints_a, 'b': ints_b}
+    rc_in = use.construct_artifact_collection(
+        'rc_in', {'a': ints_a, 'b': ints_b}
     )
 
     rc_out, = use.action(
@@ -280,6 +280,6 @@ def construct_and_access_collection(use):
         use.UsageOutputNames(output='rc_out')
     )
 
-    ints_b_from_collection = use.access_collection_member(  # noqa: F841
+    ints_b_from_collection = use.get_artifact_collection_member(  # noqa: F841
         'ints_b_from_collection', rc_out, 'b'
     )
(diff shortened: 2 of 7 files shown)


ci/recipe/meta.yaml
```diff
@@ -25,7 +25,12 @@ requirements:
     - tzlocal <3
     - python-dateutil
     - bibtexparser
-    - networkx
+    # This is pinned because networkx 3.2 was trying to use
+    # importlib.resources.files which apparently doesn't exist on
+    # importlib_resources 6.1.0 which is what we were using when this was
+    # pinned. Pinning this to the version that was working for us was simpler
+    # than digging around in importlib_resources to find a compatible version
+    - networkx =3.1
```
lizgehret, line 33: Bloatware

lizgehret (commented):
initial review looks good, i'm going to pull this down to test and make sure everything works as expected

ci/recipe/meta.yaml
colinvwood: the comment especially

Oddant1: "Why did we pin this? Is it safe to unpin it? Let me dig through the commit history to see if I can figure out why this was pinned." - Everyone every time we're looking at a pin

ebolyen: So... like do we still need this pin?

qiime2/sdk/usage.py
```diff
@@ -1210,6 +1210,123 @@ def factory():
...
+                'collection.'
+            )
+            raise ValueError(msg)
+
+        def factory():
+            from qiime2 import ResultCollection
+            # NOTE: these usage variables are assumed to have been
+            # materialized at this point
+            members_dict = {
+                key: member.value for key, member in members.items()
```
ebolyen: I think these should be `member.execute()` here, since we're a factory. The expectation is we will materialize a real thing at the end, so if we need to resolve a chain of execution, so be it.

ebolyen: (Execute memoizes the result, so it won't double-evaluate the expression).

qiime2/sdk/usage.py
```diff
@@ -1210,6 +1210,123 @@ def factory():
...
+        >>> collection_2
+        <ExecutionUsageVariable name='collection_2', var_type='artifact_collection'>
+        >>> first_member = use.access_collection_member(
+        ...     'first_member', collection_2, 'a'
+        ... )
+        >>> first_member
+        <ExecutionUsageVariable name='first_member', var_type='artifact'>
+        '''  # noqa: E501
+        def factory():
+            return collection.value[key]
```
ebolyen: Use `execute` here as well.

qiime2/plugins.py
```diff
@@ -189,6 +189,47 @@ def init_artifact_collection(self, name, factory):
 
         return variable
 
+    def construct_collection(self, name, member_type, members):
+        # make sure members is dict to avoid repeated type checking
+        if type(members) is list:
+            members = {str(i): member for i, member in enumerate(members)}
```
ebolyen: I think this logic should be in the parent class.

qiime2/core/testing/examples.py
```diff
@@ -264,3 +264,22 @@ def collection_dict_of_ints(use):
...
 
     out.assert_output_type(semantic_type='SingleInt', key='Foo')
+
+
+def construct_and_access_collection(use):
+    ints_a = use.init_artifact('ints_a', single_int1_factory)
+    ints_b = use.init_artifact('ints_b', single_int2_factory)
+
+    rc_in = use.construct_collection(
+        'rc_in', 'artifact', {'a': ints_a, 'b': ints_b}
```
ebolyen: Could we actually infer the type based on the var_types of the incoming dict/list?

ebolyen (changes requested):
I like it! Few comments on factories and where specific logic lives in base vs child classes.

qiime2/sdk/usage.py
```diff
@@ -1210,6 +1210,123 @@ def factory():
             return artifact
         return self._usage_variable(name, factory, 'artifact')
 
+    def construct_collection(
```
ebolyen: I was going to comment that this should perhaps be `init_collection` to match the others, but I realized this doesn't accept a factory, which is why I suspect you didn't name it such. I think that's probably a good reason to leave it as its own thing.

qiime2/sdk/usage.py
```diff
@@ -1210,6 +1210,123 @@ def factory():
...
+            }
+
+            return ResultCollection(members_dict)
+
+        if member_type == 'artifact':
+            return self._usage_variable(name, factory, 'artifact_collection')
+
+        return self._usage_variable(name, factory, 'viz_collection')
+
+    def access_collection_member(
```
ebolyen: I think this should probably be `get_collection_member` to match `get_metadata_column` which I think also implies that the `collection` parameter should be `variable` to also match.

ebolyen (changes requested):
Sorry to do this, but I think we need to align the API name slightly for consistency.

qiime2/plugins.py
```diff
@@ -189,6 +189,47 @@ def init_artifact_collection(self, name, factory):
...
+
+        variable = super().construct_collection(name, member_type, members)
+
+        var_name = variable.to_interface_name()
+
+        str_namespace = {str(name) for name in self.namespace}
```
ebolyen: Actually, self.namespace is already a set of strings.

qiime2/plugins.py
```diff
@@ -189,6 +189,47 @@ def init_artifact_collection(self, name, factory):
...
+        diff = set(member.name for member in members.values()) - str_namespace
+        if diff:
+            msg = (
+                f'{diff} not found in driver\'s namespace. Make sure '
+                'that all ResultCollection members have been properly '
+                'created.'
+            )
+            raise ValueError(msg)
```
ebolyen: Should this logic get hoisted into the parent class?

ebolyen (changes requested):
More comments

qiime2/core/testing/examples.py
colinvwood: Is this resolved now by our discussion yesterday (dropping the `member_type` parameter)?

qiime2/plugins.py
colinvwood: It's actually a set of "repr raw variable name" or something weird, so this is necessary

ci/recipe/meta.yaml
colinvwood: still need it because importlib_resources is still 6.1.0 in our environments.

qiime2/sdk/usage.py
```diff
@@ -1241,25 +1238,31 @@ def construct_collection(
...
         if not all(
-            member.var_type == member_type for member in members.values()
+            member.var_type == 'artifact' for member in members.values()
         ):
+            raise ValueError('Expected only artifacts in the collection.')
+
+        str_ns = {str(name) for name in self.namespace}
+        diff = set(
+            str(member.to_interface_name()) for member in members.values()
+        ) - str_ns
```
ebolyen, line 1260:
```suggestion
        diff = set(
            str(member.to_interface_name()) for member in members.values()
        ) - self.namespace
```

ebolyen (approved):
Looks good! `self.namespace` is already a set of names as strings

qiime2/sdk/usage.py
Oddant1: It's actually lying about `self.namespace` being a set of strings. It's actually a set of `<class 'qiime2.plugins.ArtifactAPIUsageVariable.repr_raw_variable_name'>
`, so if we don't cast it to str the sets are always disjoint unless both sets are empty.

ebolyen: GROSS. Ok, thank you good deal then.

Issue #731, opened 2023-11-03
replay-supplement output zip files should unzip to a top-level directory rather than the current working directory


gregcaporaso:
Currently when unzipping a reproducibility supplement, all files are extracted to the current working directory.  This should be updated so that the files are extracted to a directory with the same base name as the zip file. 

Current behavior:

```bash
(qiime2-amplicon-2023.9) ➜  q2results qiime tools replay-supplement \
  --in-fp . \
  --recurse \
  --out-fp ./reproducibility-supplement.zip
parsing unweighted_unifrac_emperor.qzv
parsing alpha-rarefaction.qzv
metadata written to recorded_metadata/
The <class 'q2cli.core.usage.ReplayCLIUsage'> replay script was written to cli_replay.sh.
metadata written to recorded_metadata/
The <class 'qiime2.core.archive.provenance_lib.usage_drivers.ReplayPythonUsage'> replay script was written to python3_replay.py.
The citations bibtex file was written to citations.bib.
The reproducibility package was written to /Users/gregcaporaso/temp/q2results/reproducibility-supplement.zip.

(qiime2-amplicon-2023.9) ➜  q2results ls
alpha-rarefaction.qzv          reproducibility-supplement.zip unweighted_unifrac_emperor.qzv

(qiime2-amplicon-2023.9) ➜  q2results unzip reproducibility-supplement.zip
Archive:  reproducibility-supplement.zip
   creating: recorded_metadata/
  inflating: cli_replay.sh
  inflating: citations.bib
  inflating: python3_replay.py
   creating: recorded_metadata/demux_emp_single_0/
   creating: recorded_metadata/diversity_alpha_rarefaction_0/
   creating: recorded_metadata/diversity_core_metrics_phylogenetic_0/
  inflating: recorded_metadata/diversity_core_metrics_phylogenetic_0/metadata_0.tsv
  inflating: recorded_metadata/diversity_alpha_rarefaction_0/metadata_1.tsv
  inflating: recorded_metadata/demux_emp_single_0/barcodes_0.tsv

(qiime2-amplicon-2023.9) ➜  q2results ls
alpha-rarefaction.qzv          cli_replay.sh                  recorded_metadata              unweighted_unifrac_emperor.qzv
citations.bib                  python3_replay.py              reproducibility-supplement.zip
```

It would help users stay organized and understand what files were created if that unzip command resulted in a top-level `reproducibility-supplement` directory with all of the files under that.

colinvwood:
[...]

PR #732, opened 2023-11-07
nest everything in folder in supplement zip

colinvwood:
addresses #731


gregcaporaso (changes requested):
@colinvwood, this looks good but I notice there are no replay_supplement tests to update here. Would you add a test that calling replay_supplement creates the expected zip file, and then that unzipping that results in the expected directory structure? I think that will be our first replay_supplement test.

colinvwood:
@gregcaporaso the replay supplement tests are in q2cli because qiime2 can't depend on q2cli

colinvwood:
@gregcaporaso tests updated in qiime2/q2cli#325

gregcaporaso (approved):
Looks good! I tested locally and everything is working as expected now.

Issue #721, opened 2023-10-05
enable access to validation_level=min on data imports


gregcaporaso:
To speed up import of very large data sets (e.g., collections of `.fastq.gz` files), we should enable users to change the format validation level on import. This functionality already exists on the formats, but isn't accessible through the existing import functionality.

colinvwood:
I believe this was closed by #734

gregcaporaso:
Confirmed, thanks @colinvwood. 

```shell
qiime tools import --help
Usage: qiime tools import [OPTIONS]

  Import data to create a new QIIME 2 Artifact. See https://docs.qiime2.org/
  for usage examples and details on the file types and associated semantic
  types that can be imported.

Options:
  --type TEXT             The semantic type of the artifact that will be
                          created upon importing. Use `qiime tools list-types`
                          to see what importable semantic types are available
                          in the current deployment.                [required]
  --input-path PATH       Path to file or directory that should be imported.
                                                                    [required]
  --output-path ARTIFACT  Path where output artifact should be written.
                                                                    [required]
  --input-format TEXT     The format of the data to be imported. If not
                          provided, data must be in the format expected by the
                          semantic type provided via --type. Use `qiime tools
                          list-formats --importable` to see which formats of
                          input data are importable.
  --validate-level [min|max]
                          How much to validate the imported data before
                          creating the artifact. A value of "max" will
                          generally read the entire file or directory, whereas
                          "min" will not usually do so.       [default: "max"]
  --help                  Show this message and exit.
```

PR #734, opened 2023-11-20
IMP: Artifact.import_data now has validate_level

ebolyen:
TODO: 
 - [x] write tests with a sacrificial format with hidden errors
 - [x] ? swap out md5sum implementation when host has a faster binary available
 - [x] md5sum tests

### DIFF
--- a/qiime2/core/testing/transformer.py
+++ b/qiime2/core/testing/transformer.py
@@ -47,7 +47,7 @@ def _7(data: list) -> IntSequenceFormat:
     ff = IntSequenceFormat()
     with ff.open() as fh:
         for int_ in data:
-            fh.write('%d\n' % int_)
+            fh.write(f'{int_}\n')
     return ff
 
 
--- a/qiime2/core/tests/test_util.py
+++ b/qiime2/core/tests/test_util.py
@@ -6,12 +6,15 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
+import shutil
 import unittest
 import tempfile
 import pathlib
 import collections
 import datetime
 import dateutil.relativedelta as relativedelta
+
+import pytest
 
 import qiime2.core.util as util
 from qiime2.core.testing.type import Foo, Bar, Baz
@@ -209,6 +212,10 @@ class TestMD5SumDirectory(unittest.TestCase):
 
             self.assertEqual(exp, obs)
 
+    @pytest.mark.skipif(
+        shutil.which('md5sum') is None,
+        reason='md5sum executable is required'
+    )
     def test_single_file_md5sum_native(self):
         with tempfile.TemporaryDirectory() as tempdir:
             fp = pathlib.Path(tempdir) / 'test.txt'
@@ -216,11 +223,8 @@ class TestMD5SumDirectory(unittest.TestCase):
                 fh.write('contents\n')
 
             exp = 'e66545a2155380046fce3fdbd32a6b4f'
-            try:
-                obs = util.md5sum_native(fp)
-                self.assertEqual(exp, obs)
-            except FileNotFoundError:
-                pass  # md5sum executable not available
+            obs = util.md5sum_native(fp)
+            self.assertEqual(exp, obs)
 
     def test_single_file_nested(self):
         nested_dir = self.test_path / 'bar'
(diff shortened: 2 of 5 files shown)


ebolyen:
fixes #721

ebolyen:
Ok, it turns out `%timeit` was just a bad idea, it only took 13 mins in python vs 9 mins in `md5sum`.

I could take or leave that last commit.

qiime2/core/testing/transformer.py
```diff
@@ -47,7 +47,7 @@ def _7(data: list) -> IntSequenceFormat:
     ff = IntSequenceFormat()
     with ff.open() as fh:
         for int_ in data:
-            fh.write('%d\n' % int_)
+            fh.write(f'{int_}\n')
```
Oddant1, line 50: :star:

qiime2/core/util.py
```diff
@@ -116,14 +118,32 @@ def duration_time(relative_delta):
...
     return md5.hexdigest()
 
 
+def md5sum_native(filepath):
+    result = subprocess.run(['md5sum', str(filepath)],
+                            check=True, capture_output=True, text=True)
+    _, digest = from_checksum_format(result.stdout)
+    return digest
+
+
```
Oddant1, line 146: Why did this whole deal need to happen?

qiime2/core/testing/transformer.py
colinvwood: 😂 this was actually necessary for the tests, not just me being petty

qiime2/core/util.py
Oddant1: Ah nvm answered my own question

colinvwood: Using the binary at all? -> for speed
Detecting whether it's present? -> not a universally present program (I don't have it e.g.)

ebolyen: Speed mostly, it's not super important as we're in the same order of magnitude for running md5sum, but it could save up to an hour on larger imports.

qiime2/core/tests/test_util.py
```diff
@@ -198,6 +198,30 @@ def test_single_file(self):
...
+            fp = pathlib.Path(tempdir) / 'test.txt'
+            with open(fp, 'w') as fh:
+                fh.write('contents\n')
+
+            exp = 'e66545a2155380046fce3fdbd32a6b4f'
+            obs = util.md5sum_python(fp)
+
+            self.assertEqual(exp, obs)
+
+    def test_single_file_md5sum_native(self):
```
ebolyen, line 219: A better approach is to skip this whole test:
https://docs.pytest.org/en/7.1.x/how-to/skipping.html#id1

ebolyen: (Sorry I should have commented on the except FileNotFoundError)

qiime2/core/tests/test_util.py
```diff
@@ -198,6 +201,31 @@ def test_single_file(self):
                 ('foobarbaz.txt', '93b048d0202e4b06b658f3aef1e764d3')
             ]))
 
+    def test_single_file_md5sum_python(self):
+        with tempfile.TemporaryDirectory() as tempdir:
```
fabianegli, line 205: It seems that you rely on pytest anyway, so why not use the pytest [tmp_path](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html) fixture? The same applies for other places where tempfile is currently in use.

gregcaporaso: Thanks for the suggestion @fabianegli! Since we're getting close to release time I think we shouldn't hold this up for this change, though it may be useful to consider this as a recommendation generally. Would you want to create an issue for this including info on the benefits it would offer?

gregcaporaso:
@Oddant1, did you review this one already, and if so is it ready for merge? If so, merge whenever!

qiime2/core/tests/test_util.py
fabianegli: No need to hold off anything. It was just meant as a comment for anyone who wants to go ahead and use pytest fixtures. The benefit is mostly less lines of code (to write and read) and less indentation. Also, if necessary, the root of the temporary directories and files for all tests can be configured in `pytest.ini`.

PR #735, opened 2023-11-30
IMP: Adds assert_no_nans_in_tables

Oddant1:
Add testing util that asserts no tables in the given file contain NaNs


Oddant1:
blocking qiime2/q2-demux#156

lizgehret:
Hey @Oddant1 - @ebolyen and I chatted about this, and the main thing that should be adjusted is the location of the new assertion. Since this assertion is being used only in the context of plugin actions, it should be moved from util.py -> plugins.py

lizgehret (changes requested):
Hey @Oddant1 - @ebolyen and I chatted about this, and the main thing that should be adjusted is the location of the new assertion. Since this assertion is being used only in the context of plugin actions, it should be moved from util.py -> plugins.py

qiime2/plugins.py
```diff
@@ -490,3 +490,16 @@ def exec_module(self, module):
...
+    '''
+    Checks for NaNs present in any of the tables in the indicated file then
+    resets to the head of the file.
+    '''
+    from pandas import read_html
+
+    tables = read_html(fh)
+    for df in tables:
+        assert not df.isnull().values.any()
+    fh.seek(0)
```
ebolyen: Wrong plugin, sorry. We want the one which a plugin would import `qiime2.plugin.testing` or something like that.

ebolyen (changes requested):

lizgehret:
Tests all pass locally and I've tested `demux summarize` on the MP demux artifact, confirming that NaNs aren't present in the table. The test failure we see here is most likely due to a parsl version change (cc @Oddant1 to investigate), but this and the [chained q2-demux PR](https://github.com/qiime2/q2-demux/pull/156) are good to merge.

PR #739, opened 2024-01-05
refactor replay namespaces

colinvwood:


qiime2/core/archive/provenance_lib/replay.py
```diff
@@ -90,152 +89,346 @@ class ActionCollections():
...
         '''
-        super().__setitem__(key, f'<{self.data[key]}>')
+        counter = 0
+        unique_name = f'{name}_{counter}'
+        names = [record.name for record in self._usg_var_ns.values()]
 
+        # no-provenance nodes are stored with angle brackets around them
+        while unique_name in names or f'<{unique_name}>' in names:
+            counter += 1
+            unique_name = f'{name}_{counter}'
```
Oddant1, line 265: Ew. Probably not worth our while to do it a better way (could namespace the counter in a global dict or something but that is probably over engineering). But still ew.

qiime2/core/archive/provenance_lib/replay.py
```diff
@@ -90,152 +89,346 @@ class ActionCollections():
...
+                if output_name not in rc_ns[action_id]:
+                    artifacts = {rc_key: provnode._uuid}
+                    rc_ns[action_id][output_name] = ResultCollectionRecord(
+                        collection_uuid=str(uuid4()), members=artifacts
+                    )
+                else:
+                    rc_ns[action_id][output_name].members[rc_key] = \
+                        provnode._uuid
+
+        self.result_collection_ns = rc_ns
```
Oddant1: Minor point, but it's a bit awkward to me that we actually set this attribute here. It seems like we only call this in the constructor and should probably change it so this returns rc_ns and the constructor actually sets the attribute.

qiime2/core/archive/provenance_lib/replay.py
```diff
@@ -90,152 +89,346 @@ class ActionCollections():
...
+
+                hashed_contents = self.hash_result_collection(record.members)
+                hashed_contents_with_keys = \
+                    self.hash_result_collection_with_keys(record.members)
+
+                c_to_c[hashed_contents] = record.collection_uuid
+                c_to_c[hashed_contents_with_keys] = record.collection_uuid
+
+        self.artifact_uuid_to_rc_uuid = a_to_c
+        self.rc_contents_to_rc_uuid = c_to_c
```
Oddant1: Same as `make_result_namespace_collection` I would probably return these two dicts and actually set the attributes in the constructor.

Oddant1 (commented):
A few initial comments. I think it does look better overall

ebolyen (approved):
I'm relying on the tests a bit here, but these changes look good. 

It looks like creating the methods have made things a little bit easier to understand from the outside in, since I don't have to track the various dictionaries anymore.

Nice job!

PR #740, opened 2024-01-10
Try to figure out where/how to insert the cache here

### LINKED ISSUE #730: Enable OwnedPaths (i.e. FileFormat and DirectoryFormat with write mode) to initialize in a Cache
**Addition Description**
This feature is largely for accelerating importing by avoiding unnecessary disk IO. 

**Current Behavior**
Currently creating an OwnedPath in QIIME 2 will use $TMPDIR which can be problematic if the destination is ultimately a Cache on a different block device. This will result in a copy of all of the data.

**Proposed Behavior**
We should have a mechanism to cause any instantiated OwnedPaths to appear in a target location. Plugin code/transformers should not be aware of this, so the immediate call-site will not control this. However the originating call-site (such as the Artifact/Python 3 API) should have a way to control it.

Oddant1:
Closes #730


Oddant1:
@ebolyen is this along the lines of what we want? It seems to be doing what we wanted. This isn't the part of the framework I'm most familiar with.

ebolyen:
Almost, I would lift this logic up to `qiime2.core.path`.

Is the process pool the best place? It seems kind of reasonable, although it's looking less like a pool every day. The automated cleanup is a nice feature of it.

We should also test this with your parsl situation with the provenance path not showing up. My hunch is it will still work since there's a flufl lock, but worth confirming before we break everything.

Oddant1:
Yeah while we were messing with that I was wondering if this functionality would resolve that

Oddant1:
Ok so we do want to put this in `__new__` for OwnedPath or something?

Oddant1:
Something like how it is now and stop passing in prefix? I had figured this would bork things pretty bad, but maybe not.

ebolyen:
Yep, in theory that's all we should need to do.

qiime2/core/path.py
```diff
@@ -24,6 +24,12 @@ def _party_parrot(self, *args):
 
 class OwnedPath(_ConcretePath):
     def __new__(cls, *args, **kwargs):
+        from qiime2.core.cache import get_cache
+
+        cache = get_cache()
+        kwargs['prefix'] = os.path.join(cache.process_pool.path,
+                                        'q2-%s-' % cls.__name__)
```
ebolyen: Let's put this under a `tmp/` directory in the process pool (creating if needed). Maybe there should be a helper on the cache itself to return the temp directory?

Oddant1:
This does not work how I thought it did at all. The actual path is one of the args, and the prefix doesn't seem to do much if anything when passed up to pathlib.Path

Oddant1:
@ebolyen I think putting the cache usage in `OutPath.__new__` does something closer to what you wanted.

qiime2/core/path.py
```diff
@@ -92,14 +92,20 @@ def _destruct(cls, path):
         else:
             os.unlink(path)
 
-    def __new__(cls, dir=False, **kwargs):
```
ebolyen, line 95: Where there any callers of this? I swear we set `prefix` in a lot of places for stuff like `q2-provenance`... although come to think of it, cache has probably eroded out most of those callsites, but there may still be a few left.

Oddant1: Only existing use is https://github.com/qiime2/qiime2/blob/c95b12bde54170b18a80c2c95f13decfb893ee14/qiime2/core/format.py#L23 and tests

Oddant1:
The child processes will not clean these tmp dirs out when they exit because their termination does not call exit funcs.

Oddant1:
This may actually be useful for #790

ebolyen:
Superseded by #790

PR #741, opened 2024-01-11
IMP: Ensure default args are captured when index checking

Oddant1:
When doing pipeline resumption default values are not being captured correctly for indexing causing artifacts that could be recycled to be regenerated.

### DIFF
--- a/qiime2/core/tests/test_pipeline_resumption.py
+++ b/qiime2/core/tests/test_pipeline_resumption.py
@@ -123,7 +123,7 @@ class TestPipelineResumption(unittest.TestCase):
             self.assertNotEqual(identity_uuid, complete_identity_uuid)
             self.assertNotEqual(viz_uuid, complete_viz_uuid)
 
-    def test_resumable_pipeline_OLD(self):
+    def test_resumable_pipeline(self):
         with self.pool:
             with self.assertRaises(PipelineError) as e:
                 self.pipeline(
--- a/qiime2/core/tests/test_pipeline_resumption.py
+++ b/qiime2/core/tests/test_pipeline_resumption.py
@@ -96,7 +96,7 @@ class TestPipelineResumption(unittest.TestCase):
         with self.cache:
             with self.assertRaises(PipelineError) as e:
                 self.pipeline(
-                    self.ints1, self.ints2, self.int1, 'Hi', self.md1,
+                    self.ints1, self.ints2, self.md1, self.int1, 'Hi',
                     fail=True)
 
             ints1_uuids, ints2_uuids, int1_uuid, list_uuids, dict_uuids, \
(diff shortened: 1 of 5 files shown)


Oddant1:
The real functional change here is in signature.py the rest is testing a lot of the diff was changing the order of the arguments because I added default values to a couple things

ebolyen (approved):
I haven't done a comprehensive review, but the fundamental change here looks good to me.

The reasoning as I understand it is:
Within a `bind` the defaults are already mapped to kwargs, so from the inside there's no such thing as a default.

But within `context`, we're using collate_inputs to resolve some of the signature, but as we're outside of bind, defaults aren't a part of kwargs yet as we're not doing anything funky to the callable via `decorator`. So by having collate_inputs include the defaults, nothing changes from within the `bind` as they were already there, but now `context` has an easy way to see what the inside of `bind` sees. 

I also don't think anyone uses this method and it did not previously have a convenient definition. But now it's actually pretty useful.

Oddant1:
@ebolyen it is also actually getting used in `bound_callable`. Its job was originally just to put all `args` and `kwargs` in a dict together. That step was already happening in `bound_callable`, but I needed it to happen in `context `as well, so I factored it out into a method on `signature`. 

That method just now also puts default values there if need be. In `bound_callable` it won't need to do that but in `context` it will when they don't provide an argument. I don't think this is a problem, but I do have a question. How do we feel about this:
___
image
___

I created a modded version of `core_metrics` that does not pass `sampling_depth` to `rarefy`. The result is that in `context` the `rarefy` call has `NOVALUE` for `sampling_depth`. This error is being raised intentionally by me. If I don't raise it then we get a cache miss and try to rerun `rarefy `which then fails because no value was provided for `sampling_depth`, so it doesn't pass regular signature checking.

 This shouldn't be an issue because it should ensure an cache miss because an artifact can't have been created with `NOVALUE` for a required argument, but it makes me feel like maybe things need to be refactored so input checking is happening for `context` in the same place and same way as `bind`. . . But `bind` does it by rewriting the signature of the `bound_callable`, so that may not be viable.

Oddant1:
@ebolyen maybe something more like fd9e137? With either approach, I wonder if it's worth it to dedup the redundant work happening here because `context` and `bind` do a lot of the same stuff to set up the arguments now.

Oddant1:
What about an approach like fd9e137 where we rewrite the signature in the same way as `bind`? I wonder if it's worth it to dedup the redundant work happening here

qiime2/core/tests/test_pipeline_resumption.py
```diff
@@ -123,19 +123,19 @@ def test_resumable_pipeline_no_pool(self):
             self.assertNotEqual(identity_uuid, complete_identity_uuid)
             self.assertNotEqual(viz_uuid, complete_viz_uuid)
 
-    def test_resumable_pipeline(self):
+    def test_resumable_pipeline_OLD(self):
```
lizgehret:
```suggestion
    def test_resumable_pipeline(self):
```

lizgehret (approved):
lgtm, thanks @Oddant1

colinvwood (approved):

PR #742, opened 2024-01-11
BUG: [Issue 738] dangling recycle references

### LINKED ISSUE #738: Recycle uses dangling references in cache induced by OS X tmpdir cleanup
**Bug Description**
Dangling references cause an error when loading an artifact via `recycle`.
This is usually going to be a result of the `data/` directory being cleaned without the reference also getting cleaned.
Because garbage collection opens all of the references, any timeout/last-modified counter is refreshed for references, but will not be refreshed for the directories within data (as only the dir-entries of `data/` are read, rather than each file).

1. Use OS X
2. Run a failing command with default recycle argument.
3. Wait less than 2 days and run any other successful QIIME command (causes references from 2 to be read again).
4. Wait another day for OS X to delete the data directories from step 2.
5. Re-run the command from 2 with the same default recycle.

**Expected behavior**
We should either detect dangling references and delete them during GC, or tolerate dangling references and regenerate data from the source QZAs as needed.

**Computation Environment**
- OS: OS X
- QIIME 2 Release 2023.9

Oddant1:
Closes #738. At this point the changes to GC are a bit of a cludge to solve the specific problem the issue was created for. The change to indexing makes that specific code resistant to failure when backing data has been spontaneously deleted which makes some amount of sense since that's probably the only place where we will reliably be able to continue if data is missing.

### DIFF
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -722,6 +722,9 @@ class Cache:
                                      f" '{self.path}' does not point to"
                                      " anything")
 
+                if (data and not os.path.exists(self.data / data)) and (pool and not os.path.exists(self.pools / pool)):
+                    os.remove(self.keys / key)
+
             # Walk over pools and remove any that were not referred to by keys
             # while tracking all data within those that were referenced
             for pool in self.get_pools():
@@ -729,7 +732,10 @@ class Cache:
                     shutil.rmtree(self.pools / pool)
                 else:
                     for data in os.listdir(self.pools / pool):
-                        referenced_data.add(data)
+                        if os.path.exists(self.data / data):
+                            referenced_data.add(data)
+                        else:
+                            os.remove(self.pools / pool / data)
 
             # Add references to data in process pools
             for process_pool in self.get_processes():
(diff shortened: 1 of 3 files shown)


qiime2/core/cache.py
```diff
@@ -754,6 +760,18 @@ def garbage_collection(self):
                     set_permissions(target, None, USER_GROUP_RWX)
                     shutil.rmtree(target)
 
+    def _check_dangling_reference(self, data_path, key_path):
+        """ If the data specified does not exist then remove the key
+        """
+        if not os.path.exists(data_path):
+            warnings.warn(f"Dangling reference {key_path}. Data at {data_path}"
+                          " does not exist. Reference will be removed")
+            os.path.remove(data_path)
```
```suggestion
            os.path.remove(key_path)
```

(I assume)

ebolyen (commented):
This seems fine (other than a typo), good call on the pools potentially suffering a similar fate.

qiime2/core/cache.py
Oddant1: Whoops yeah good catch. Probably made that while moving things into that helper

Oddant1:
This is currently taking the approach of "If we find dangling references during garbage collection warn the user about the dangling reference and delete the reference." It is still possible to see the error @lizgehret saw, but the data causing the error should be cleaned up by garbage collection as the failing action exits, and if garbage collection were run in any other way after the dangling reference is created but prior to reading it then the error would never be seen at all.

lizgehret (commented):
let's just add a test for this that will delete the data associated with the ref, and will then return that warning before removing the dangling ref. then this will be g2g 👍

PR #743, opened 2024-01-18
Unwrap the results from tomlkit.load so we have normal Python types

Oddant1:
Fix a bug where toml was giving us wrapped results from the parser that didn't work the same as Python built-ins.


gregcaporaso (approved):
Looks good @Oddant1!

PR #745, opened 2024-01-22
IMP: Add Jobs and Threads primitive types

colinvwood:


ebolyen (approved):
this all looks right to me!

ebolyen:
Just reviewed the linked PRs and this was not nearly as bad as I was expecting. It all looks fine to me :)

I had one note about adding a helper in this PR in the q2galaxy PR.

qiime2/sdk/util.py
```diff
@@ -183,3 +185,26 @@ def get_available_usage_drivers() -> Dict[str, 'UsageDriver']:
         entry_point.name: entry_point.resolve() for entry_point in
         iter_entry_points(group='qiime2.usage_drivers')
     }
+
+
+def get_available_cores(one_less: bool = False):
```
ebolyen: Put this in `qiime2.plugin` if it's for plugins, and I would actually recommend `psutil` it's a bit more robust than the stdlib for this things (and we already depend on it).

lizgehret (approved):
once `get_available_cores` is moved, this should be g2g @colinvwood

PR #746, opened 2024-01-24
refer to qiime2 organization's default 'community health docs'...

gregcaporaso:
… rather than local ones


gregcaporaso:
Replaced by #841.

PR #755, opened 2024-02-26
BUG: composite semantic-type in `TypeMap` where child-type has properties; breaks CLI

### LINKED ISSUE #748: BUG: composite semantic type in `TypeMap` where child type has properties; breaks CLI
**Bug Description**
Having a composite type where the parent type's attribute `filed_names!=None` and the child type has properties, in the keys of a `TypeMap`, will cause an error when calling the action that uses said `TypeMap` from the CLI.

I will illustrate an example using `q2-moshpit`, but I believe this would happen in any plugin where there is a `TypeMap` where the parent type's attribute `filed_names!=None` and the child type has properties.

1. Set up environment
```bash
mamba env create -n q2-shotgun-2023.11 --file https://raw.githubusercontent.com/qiime2/distributions/dev/2024.2/shotgun/passed/qiime2-shotgun-macos-latest-conda.yml
conda activate q2-shotgun-2023.11
```
2. Download some data
```bash
cd <download here>
wget wget https://polybox.ethz.ch/index.php/s/Zhk2igDBGRJTi1n/download -O reads.qza
qiime moshpit build-kraken-db --p-collection viral --output-dir ./viral-db
```
3. Generate reports.
```bash
qiime moshpit classify-kraken2 --i-seqs reads.qza --i-kraken2-db ./viral-db/kraken2_database.qza --p-quick --p-memory-mapping False --p-threads 16 --output-dir ./kraken-results
```
4. Collate results. _This should fail._
```
qiime moshpit collate-kraken2-reports --i-kraken2-reports kraken-results/reports.qza --o-collated-kraken2-reports kraken-results/collated_reports.qza --verbose
```
> This is the action that uses a `TypeMap` where the parent type's attribute `filed_names!=None` and the child type has properties.

**Expected behavior**
Action works.

**Computation Environment**
- MacOS: Sonoma 14.3
- QIIME 2 Release 2023.11

**Comments**
1. `qiime2.core.type.meta.TypeMap` allows developers to design a mapping between inputs and outputs in actions, effectively constraining the inputs that a user can provide and restricting the respective outputs. 
3. A `TypeMap` object is initialized with a dictionary where keys represent the semantic types of the inputs to the action and values represent the semantic types of the outputs. Both keys and values can be tuples, effectively allowing the mapping of multiple inputs or parameters to multiple outputs. 
6. Now, let's consider a `TypeMap` where in the keys there is a composite type where the parent type has a non `None` value for the attribute `filed_names` and the child type has a property. Following an example from the documentation: `Kitchen[Knife % Properties(['fillet', 'sharp'])]`.  It's important to note that in the definition of the parent type `Kitchen` the `filed_names` attribute is not `None`.
```python
Knife = SemanticType('Knife', variant_of=[Kitchen.field['utensil']])
Kitchen = SemanticType('Kitchen', field_names=['utensil'],field_members={'utensil': [Knife]})
```
- The `TypeMap` under consideration might look like this:
```python
input, output = TypeMap({
   Kitchen[Knife % Properties(['fillet', 'sharp'])]: Int, 
   Kitchen[Knife % Properties('sharp')]: Int, 
   Kitchen[Knife % Properties('fillet')]: Int, 
})
```
7. Now for the sake of completion let's define an action that uses this `TypeMap`
```python
plugin.methods.register_function(
    function=q2_foo.bar.toy_action,
    inputs={"knife": input},
    parameters={},
    outputs={"number": output},
    name="Toy action",
    description="Explains bug"
)
```
8. Now then, running `qiime foo toy-action --i-knife knife.qza --o-number number.qza` will throw the following error:
```
File "<path to your env>/lib/python3.8/site-packages/qiime2/core/type/template.py", line 119, in validate_field_count
    raise TypeError("Expected only %r fields, got %r" % (exp, count))
TypeError: Expected only 0 fields, got 1
```
9. Why does this happen?
   1. It seems to me that the command triggers the initialization of an object that represents the action. This initialization is based on the signature of the action. Said signature is computed and saved when calling `qiime dev refresh-cache`. 
   3. At the same time, this action-object contains other objects that represent the options in the command (inputs, parameters, and outputs). 
   4. When one of these options contains a variable from a `TypeMap`  (e.g. `--i-knife`) then a `TypeMap` object is instantiated based on the signature of the option which comes in the form of a dictionary. This dictionary contains strings rather than objects, for example, the name of the semantic type (e.g. `'Kitchen'`) rather than an instance of said semantic type. So using this signature from the action the new semantic types are created (e.g. calling `SemanticType('Kitchen')`, `SemanticType('Knife')`), collected, and passed to the `TypeMap` class as a dictionary. 
   5. The creation of the semantic types happens in line 222 in `qiime2/core/type/parse.py` and it's what causes the error downstream because it lacks some attributes whose presence is verified downstream.
   7. Coming back to our example. The `TypeMap` invocation will then compute intersections of the different keys/branches passed (e.g. `Kitchen[Knife % Properties('sharp')] & Kitchen[Knife % Properties('fillet')]`. See the `__init__` method of `TypeMap` from more context.).
   8. For composite types this intersection operation recurses into the child type (e.g. `Knife`) and the properties, computes the intersection between the two keys/branches (e.g. `Properties('sharp') & Properties('fillet') = Properties(['sharp', 'fillet'])`), and then recurses out, producing a new semantic type (e.g. `Knife % Properties(['sharp', 'fillet']`).
   10. On the way out during the recursion, the construction of the composite type with the intersection is attempted (e.g. `Kitchen[Knife % Properties(['sharp', 'fillet']]`), by calling `qiime2.core.type.grammar.TypeExp.duplicate(fields=new_fields, predicate=new_predicate)`. Where `new_fields=(Knife % Properties(['sharp', 'fillet'],)` and `new_predicate=()`. 
   11. This function calls another function that validates that the composite semantic type (`Kitchen`) has indeed the same number of fields as the ones being passed, where `field=(Knife % Properties(['sharp', 'fillet'],)` so in this case `len(field)=1`. So if `len(field)!=len(field_names)` the validation errors out. 
   12. `field_names` is an attribute from the semantic type `Kitchen` which was left to the default (`None`) during the creation of the semantic type (described in step iv) therefore `1=len(field)!=len(None)=0`). Thus the validation fails, since it expects 0 fields but gets 1 instead. 

**Solution?**
When instantiating the new semantic types based on the signature of the action/option (`qiime2/core/type/parse.py:222`) one needs to also specify if the semantic type at hand has field names, or is a variant of another semantic type etc. E.g.

```python
221        if not json_ast['builtin']:
-              base_template = semantic.SemanticType(name).template
+              base_template = semantic.SemanticType(
+                  name, 
+                  field_names=json_ast['field_names'], 
+                  field_members=json_ast['field_members'], 
+                  variant_of=json_ast['variant_of']
+              ).template
```

`json_ast` is the dictionary that contains all the relevant information for a given semantic type. For example `name = json_ast["name"]` contains the name of the semantic type.

The catch is that this information is not readily available in `json_ast` or the call to `ast_to_type()` for that matter, which is the function where the `semantic.SemanticType()` call happens. What's worse, the child type in the composite type is instantiated before the parent type, so at the time when the child class is instantiated (`Knife`), there is no instance of the parent type `Kitchen` that can be given to `variant_of` to instantiate it correctly. Failing to specify this will also throw an error. 

So it seems to me that a major change is needed to fix this, since already the order of the calls has to be rearranged. Also, the information that is gathered from the action signature at chache time must now include this information about the semantic types (`field_names`, `field_members`, `variant_of`) such that it is available in `json_ast` and can be used to instantiate them.

Sann5:
Closes #748. Look there for a detailed description of the bug and the proposed solution.

### DIFF
--- a/qiime2/core/type/tests/test_parse.py
+++ b/qiime2/core/type/tests/test_parse.py
@@ -100,7 +100,7 @@ class TestParsing(unittest.TestCase):
         self.assertIs(W1.mapping, X1.mapping)
 
     def test_TypeMap_with_properties(self):
-        inp, _ = TypeMap({
+        I, OU = TypeMap({
             C1[Foo % Properties(['A', 'B', 'C'])]: Str,
             C1[Foo % Properties(['A', 'B'])]: Str,
             C1[Foo % Properties(['A', 'C'])]: Str,
@@ -110,10 +110,18 @@ class TestParsing(unittest.TestCase):
             C1[Foo % Properties(['C'])]: Str,
         })
 
-        try:
-            inp = ast_to_type(inp.to_ast(), scope={})
-        except Exception as e:
-            self.fail(f"Raised exception: {e}")
+        scope = {}
+        i = ast_to_type(I.to_ast(), scope=scope)
+        o = ast_to_type(OU.to_ast(), scope=scope)
+
+        self.assertEqual(scope[id(I.mapping)], [i, o])
+        self.assertEqual(len(scope), 1)
+
+        # Assert mapping is the same after ast_to_type call
+        self.assertEqual(I.mapping.lifted, i.mapping.lifted)
+
+        # Assert that the mapping object is the same in both i and o
+        self.assertIs(i.mapping, o.mapping)
 
     def test_syntax_error(self):
         with self.assertRaisesRegex(ValueError, "could not be parsed"):
--- a/qiime2/core/type/tests/test_parse.py
+++ b/qiime2/core/type/tests/test_parse.py
@@ -99,6 +99,22 @@ class TestParsing(unittest.TestCase):
         self.assertIs(V1.mapping, W1.mapping)
         self.assertIs(W1.mapping, X1.mapping)
 
+    def test_TypeMap_with_properties(self):
+        inp, _ = TypeMap({
+            C1[Foo % Properties(['A', 'B', 'C'])]: Str,
+            C1[Foo % Properties(['A', 'B'])]: Str,
+            C1[Foo % Properties(['A', 'C'])]: Str,
+            C1[Foo % Properties(['B', 'C'])]: Str,
+            C1[Foo % Properties(['A'])]: Str,
+            C1[Foo % Properties(['B'])]: Str,
+            C1[Foo % Properties(['C'])]: Str,
+        })
+
+        try:
+            inp = ast_to_type(inp.to_ast(), scope={})
+        except Exception as e:
+            self.fail(f"Raised exception: {e}")
+
     def test_syntax_error(self):
         with self.assertRaisesRegex(ValueError, "could not be parsed"):
             string_to_ast('$')
--- a/qiime2/core/type/parse.py
+++ b/qiime2/core/type/parse.py
@@ -219,7 +219,14 @@ def ast_to_type(json_ast, scope=None):
 
         name = json_ast['name']
         if not json_ast['builtin']:
-            base_template = semantic.SemanticType(name).template
+            base_template = semantic.SemanticType(
+                name,
+                field_names=['field' + str(i) for i in range(len(fields))],
+                field_members={
+                    ('field' + str(i)):
+                    [child] for i, child in enumerate(fields)
+                },
+            ).template
         elif name == 'Visualization':
             return visualization.Visualization
         elif name in {'List', 'Set', 'Tuple', 'Collection'}:


Sann5:
@ebolyen here it is. So far it looks like it doesn't break anything. I'm in the process of devising a unit test for the update.

qiime2/core/type/tests/test_parse.py
```diff
@@ -99,6 +99,22 @@ def test_typevars(self):
...
+            C1[Foo % Properties(['A', 'B'])]: Str,
+            C1[Foo % Properties(['A', 'C'])]: Str,
+            C1[Foo % Properties(['B', 'C'])]: Str,
+            C1[Foo % Properties(['A'])]: Str,
+            C1[Foo % Properties(['B'])]: Str,
+            C1[Foo % Properties(['C'])]: Str,
+        })
+
+        try:
+            inp = ast_to_type(inp.to_ast(), scope={})
```
ebolyen: I think a perfect test would be to assert something about the "re-hydrated" type, but this isn't bad either. (You don't really need the try-except as encountering an exception in a test is fairly _unexceptional_.)

Sann5: @ebolyen thanks for the feedback. I modified the test to probe the "re-hydrated" type, essentially mimicking the assertions from the test immediately before this one. Let me know if you think something else should be tested :)

ebolyen (approved):
This looks great, thank you @Sann5!

PR #760, opened 2024-03-26
Dunder fix to galaxy error

Clockwork-Rat:
This is a test approach to avoid breaking existing logic.

---

Include any questions for reviewers, screenshots, sample outputs, etc.


ebolyen:
I think we need to check if the `view_type` was set and is an instance of FormatBase. The current changes will just declare everything to be a format, since the semantic type `type` is always required.

PR #761, opened 2024-04-02
We can accept masked 0o41755 permissions on the cache as well…Probably, we will need to test this

Oddant1:
We can probably accept permissions a bit below what we try to set


Oddant1:
Do we even NEED x  for self? Can probably be `0o41644`

Oddant1:
We probably can't realistically do this, we may end up closing this PR

ebolyen:
Closing since we don't have a good understanding of why the permissions are getting knocked off the QIIME 2 tmp directory, and we do need those for multi-tenant systems.

PR #762, opened 2024-04-03
Allow Collection[Visualization]

Oddant1:
Allow `Collection[Visualization]` as a valid output type, this is necessary for [q2-boots](https://github.com/qiime2/q2-boots). 

Note: This is still pending usage changes both here in the framework and most likely in the cli and perhaps also mystery-stew


ebolyen (approved):
Turns out this works as is. Thanks @Oddant1!

PR #768, opened 2024-05-09
Flatten pipelines

Oddant1:
Nested pipelines should now flatten

### DIFF
--- a/qiime2/sdk/action.py
+++ b/qiime2/sdk/action.py
@@ -438,13 +438,17 @@ class Action(metaclass=abc.ABCMeta):
         # pipeline that calls two other pipelines within it and execute both of
         # those internal pipelines simultaneously.
         if isinstance(self, qiime2.sdk.action.Pipeline):
-            execution_ctx['parsl_type'] = 'DFK'
-            # NOTE: Do not make this a python_app(join=True). We need it to run
-            # in the parsl main thread
-            future = join_app()(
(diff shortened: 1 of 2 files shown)


qiime2/sdk/action.py
```diff
@@ -438,13 +438,23 @@ def _bind_parsl(self, ctx, *args, **kwargs):
...
-            # NOTE: Do not make this a python_app(join=True). We need it to run
-            # in the parsl main thread
-            future = join_app()(
-                    _run_parsl_action)(self, ctx, execution_ctx,
-                                       mapped_args, mapped_kwargs,
-                                       inputs=futures)
+            # If ctx._parent is None then this is the root pipeline and we want
+            # to dispatch it to a join_app
+            if ctx._parent is None:
+                execution_ctx['parsl_type'] = 'DFK'
```
ebolyen: I think this should get lifted above the `if ctx._parent is None`. That way the direct/flattened binds will also show that they are running as DFK, which is true, instead of  having `parsl_type` unset.

ebolyen (changes requested):
Just a tiny fix

qiime2/sdk/action.py
Oddant1: Good point

ebolyen (approved):
This is an exciting change, thanks @Oddant1!

PR #769, opened 2024-05-23
Swap self.get_transformer for qiime2.plugin.util.transform

### LINKED ISSUE #757: update `TestPluginBase.transform_format` to identify transformers available in all currently installed plugins
Currently `TestPluginBase.transform_format` will only discover transformers that are defined in the plugin being tested. This should be adapted to support discovery of transformers in any installed plugin to simplify plugin testing. 

This should be possible by using `qiime2.plugin.util.transform` internally, rather than [`self.get_transformer`](https://github.com/qiime2/qiime2/blob/630696237166f60765f7bbe6e3c668ea0e1495d9/qiime2/plugin/testing.py#L244).

Oddant1:
Closes #757


Oddant1:
I confirmed this test fails with the old code

gregcaporaso (approved):
Looks good, thanks @Oddant1!

PR #770, opened 2024-05-25
Hooks

Oddant1:


qiime2/sdk/action.py
```diff
@@ -286,8 +287,20 @@ def bound_callable(*args, **kwargs):
                     self.signature.transform_and_add_callable_args_to_prov(
                         provenance, **callable_args)
 
-                outputs = self._callable_executor_(
-                    scope, callable_args, output_types, provenance)
+                action = UsageAction(self.plugin_id, self.name)
```
```suggestion
                action = UsageAction(self.plugin_id, self.id)
```

ebolyen:
This has drifted too far from #820, but it's still a great idea. I think we should pick it up again soon (and I'm hoping it will be a pretty small PR based on the newer refactor).

PR #771, opened 2024-05-28
If something isn't a directory then it also isn't a cache

Oddant1:
If something is a file or doesn't exist on the file system then it isn't a cache. 

`Cache.is_cache` previously assumed it was receiving a directory and errored if it didn't. Now it just says "that's not a cache" because it isn't.

### DIFF
--- a/qiime2/core/tests/test_cache.py
+++ b/qiime2/core/tests/test_cache.py
@@ -158,6 +158,48 @@ class TestCache(unittest.TestCase):
         """
         self.assertFalse(Cache.is_cache(self.not_cache_path))
 
+    def test_file_is_not_cache(self):
+        """Verfies that a single file is not a cache
+        """
+        # Create a thing and make sure it exists and is a file
+        x = tempfile.NamedTemporaryFile()
+        os.path.isfile(x.name)
+
+        # Make sure that file is not a cache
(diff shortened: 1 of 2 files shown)


Oddant1:
~~Blocking qiime2/q2cli#332~~
- [x] comment edited

qiime2/core/tests/test_cache.py
```diff
@@ -158,6 +158,48 @@ def test_is_not_cache(self):
...
+                string.ascii_uppercase +
+                string.digits,
+                k=10))
+
+        while os.path.exists(random_string):
+            if tries == 10:
+                raise ValueError(
+                    "We randomly generated 10 alphanumeric strings 10 "
+                    "characters long and somehow all 10 existed as paths on "
+                    "your filesystem.")
```
colinvwood, line 189: This is guaranteed to be raised if you run this in Chloe's or Andrew's downloads folder

colinvwood (approved):

qiime2/core/tests/test_cache.py
colinvwood: @cherman2

cherman2: Not nice!

cherman2:
LGTM!

cherman2 (approved):

PR #772, opened 2024-06-03
Use ConfigParser instead of SafeConfigParser

athos-ribeiro:
The configparser's SafeConfigParser has been renamed to ConfigParser in Python 3.2 [1]. It was finally removed in Python 3.12 [2].

[1] https://docs.python.org/dev/whatsnew/3.2.html#configparser
[2] https://docs.python.org/3/whatsnew/3.12.html#configparser


athos-ribeiro:
Closing as per https://github.com/qiime2/q2-sample-classifier/pull/229

PR #773, opened 2024-06-27
Plugin Docs

ebolyen:
A lot of text here to review... Sorry about that.


qiime2/core/type/meta.py
```diff
@@ -90,6 +90,75 @@ def to_ast(self):
...
+    | Parameter A | Parameter B | Result      |
+    +=============+=============+=============+
+    | True        | InputTypeA  | ResultTypeA |
+    +-------------+-------------+-------------+
+    | False       | InputTypeA  | ResultTypeB |
+    +-------------+-------------+-------------+
+    | False       | InputTypeB  | ResultTypeB |
+    +-------------+-------------+-------------+
+
+    Where if the user provide ``True`` to Parameter A, they MUST provide
```
cherman2:
```suggestion
    Where if the user provides ``True`` to Parameter A, they MUST provide
```

qiime2/core/type/meta.py
```diff
@@ -90,6 +90,75 @@ def to_ast(self):
 
 
 class TypeMap(ImmutableBase):
+    """A table of input types which match to output types.
+
+    The TypeMap is best thought of as a table in which QIIME 2 is trying to
+    find a row which matches the user's input. Once found, the row-wise search
```
cherman2:
```suggestion
    find a row that matches the user's input. Once found, the row-wise search
```

cherman2: So there isnt which twice in the sentence.

qiime2/core/type/meta.py
```diff
@@ -90,6 +90,75 @@ def to_ast(self):
...
+    +-------------+-------------+-------------+
+    | False       | InputTypeB  | ResultTypeB |
+    +-------------+-------------+-------------+
+
+    Where if the user provide ``True`` to Parameter A, they MUST provide
+    ``InputTypeA`` to Parameter B, and will receive ``ResultTypeA``.
+    Otherwise, they may pass ``False`` to Parameter A, and provide either
+    ``InputTypeA`` or ``InputTypeB``, but will now receive ``ResultTypeB``.
+
+    Note that there is not a solution for ``True`` and ``InputTypeB``, so the
```
cherman2:
```suggestion
    Note that there is not a valid result for ``True`` to be given as the value for ``Parameter A`` and ``InputTypeB`` to be given as a value for  ``Parameter B`` , so the
```

cherman2: This sentence was confusing to me so I tried to clarify

qiime2/core/type/meta.py
```diff
@@ -90,6 +90,75 @@ def to_ast(self):
...
+    TypeMap does not permit that.
+
+    This can be used to constrain dependent input parameters to a more limited
+    domain than they would otherwise possess if they were treated
+    independently.
+
+    If a TypeMap is used exclusively to constrain inputs but does not impact
+    the output in any way, then the convention is to use
+    :py:data:`Visualization` to indicate a nonsense output and that final
```
cherman2:
```suggestion
    :py:data:`Visualization` to indicate a "nonsense" output and that final
```

qiime2/core/type/meta.py
```diff
@@ -90,6 +90,75 @@ def to_ast(self):
...
+    type variable is ignored (an unbound output variable has no effect so
+    ``Visualization`` distinguishes the intention from an accidental omission).
+
+    It is also possible to define multiple outputs which are dependent on
+    inputs, so long as the value of the dictionary is a tuple. This will result
+    in additional type variables to be used in the output registration.
```
cherman2, line 144: I think an example would be helpful here.

qiime2/core/type/primitive.py
```diff
@@ -29,6 +29,70 @@ def get_name(self):
...
 class Range(_PrimitivePredicateBase):
+    """A predicate which defines a contiguous range of allowable values.
+
+    Can be used with :py:data:`.Int` and :py:data:`.Float`.
+
+    Parameters
+    ----------
+    [start] : number
+      When provided as the first argument, the start of the range.
+      Will be ``None`` (meaning negative infinity) when not provided.
```
cherman2:
```suggestion
      When provided as the first argument, the value will be the start of the range.
      Will be ``None`` (meaning negative infinity) when not provided.
```

qiime2/sdk/usage.py
```diff
@@ -491,7 +491,7 @@ def __init__(self, name: str, factory: Callable[[], Any],
             Most are self-explanatory, but "format" indicates that the factory
             produces a QIIME 2 file format or directory format, which is used
             for importing data.
-        use : Usage
+        usage : Usage
```
cherman2, line 494: I don't understand why this was changed. (I am sure its not wrong, but yeah just unsure as to why)

qiime2/core/type/semantic.py
```diff
@@ -240,6 +240,64 @@ def update_ast(self, ast):
...
+          property if they match. This means that a subtype (with a set of
+          properties ``X``) of an expression (with properties ``Y``)
+          must have properties such that ``X >= Y``. As a consequence, each
+          additional property narrows the type further.
+        exclude : tuple[str]
+          Treated as an inverse of ``include``. The absence of a property
+          is not sufficient for an excluded property to match. Excluded
+          properties must be explicitly defined on the type. As a consequence,
+          this field is rarely if ever used.
+          It may be deprecated in the future.
```
cherman2, line 265: Is this what you mean by the conceptual opposite?

qiime2/plugin/__init__.py
```diff
@@ -23,4 +23,308 @@
...
+A regular list:
+
+>>> ['a', 'b', 'c'] in List[Str]
+True
+
+A collection is also compatible:
+
+>>> {'arbitrary_key': 'a'} in List[Str]
+True
+"""
```
cherman2, line 60: this just ignores the key, is that correct?

qiime2/plugin/__init__.py
```diff
@@ -23,4 +23,308 @@
...
+--------
+>>> from qiime2.plugin import Visualization
+>>> Visualization
+Visualization
+"""
+
+Int = Int
+"""
+An integer without any particular bounds.
+It can use the predicates :py:class:`.Range`, :py:func:`.Start`, and
```
cherman2, line 105: Can it use these functions or can it be used with this functions?

qiime2/plugin/plugin.py
```diff
@@ -5,11 +5,64 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
-
+"""
+.. For the docs below, we set this artificial global context
```
cherman2:
```suggestion
.. For the docs below, we set the following artificial global context
```

qiime2/plugin/plugin.py
```diff
@@ -178,8 +347,50 @@ def decorator(validator):
...
+        This decorator may be used with or without arguments.
+
+        Parameters
+        ----------
+        _fn : Callable
+            Ignore this parameter as it is the mechanism to allow argumentless
+            decoration
+        citations : list of CitationRecord
+          Citations to associate with a result whenever this transformer is
+          used internally. Can also use an entire :py:class:`Citations` object.
```
cherman2: these don't have the same indentition but I think that doesn't matter

qiime2/plugin/plugin.py
```diff
@@ -178,8 +347,50 @@ def decorator(validator):
...
+        decorator
+          A decorator which can be applied to a function which takes a single
+          argument with a type annotation and a single return annotation.
+          This decorated function will then be executed whenever data matching
+          the type annotation exists and there is a need to view that data as
+          the return annotation.
+
+        Notes
+        -----
+        Since the function is entirely defined by the input and output type the
```
cherman2:
```suggestion
        Since the function is entirely defined by the input and output type, the
```

qiime2/plugin/plugin.py
```diff
@@ -233,6 +444,23 @@ def decorator(transformer):
             return decorator(_fn)
 
     def register_semantic_types(self, *type_fragments):
+        """Register semantic type fragments to the plugin
```
cherman2, line 447: Have you defined semantic type framgrments anywhere. I am not sure familar with this term.

qiime2/plugin/plugin.py
```diff
@@ -327,11 +613,80 @@ def __init__(self, plugin):
...
+        ----------
+        function : callable
+          A function which will be called when the user uses this method.
+        inputs : dict[str, type expression]
+          A dictionary of function-parameter names to semantic type
+          expressions. The keys of the dictionary must match the parameter
+          names which are on the ``function``. Collections and type variables
+          are also permitted so long as they contain semantic types.
+        parameters : dict[str, type expression]
+          A dictionary of function0parameter names to primitive type
```
cherman2:
```suggestion
          A dictionary of function parameter names to primitive type
```

qiime2/plugin/plugin.py
```diff
@@ -350,10 +705,62 @@ def register_function(self, function, inputs, parameters, outputs, name,
...
+          Whether this action is deprecated and should be migrated away from.
+        examples : dict[str, callable]
+          A dict of example name to usage example functions which take a
+          single argument (the usage driver, a.k.a. ``use``). Each function
+          will carry out some example situation for this action when executed.
+
+        Notes
+        -----
+        Unlike methods and pipelines, there are no registered outputs as every
+        visualizer returns a single output.
```
cherman2:
```suggestion
        visualizer returns a single ``Visualization``  output.
```

qiime2/plugin/plugin.py
```diff
@@ -350,10 +705,62 @@ def register_function(self, function, inputs, parameters, outputs, name,
...
+        function : callable
+          A function which will be called when the user uses this method. This
+          function receives an ``output_dir`` as the first argument.
+        inputs : dict[str, type expression]
+          A dictionary of function-parameter names to semantic type
+          expressions. The keys of the dictionary must match the parameter
+          names which are on the ``function``. Collections and type variables
+          are also permitted so long as they contain semantic types.
+        parameters : dict[str, type expression]
+          A dictionary of function0parameter names to primitive type
```
cherman2:
```suggestion
          A dictionary of function parameter names to primitive type
```

qiime2/plugin/plugin.py
```diff
@@ -373,11 +780,68 @@ def register_function(self, function, inputs, parameters, name,
...
+        function : callable
+          A function which will be called when the user uses this method. This
+          function receives a Context object ``ctx`` as its first argument.
+        inputs : dict[str, type expression]
+          A dictionary of function-parameter names to semantic type
+          expressions. The keys of the dictionary must match the parameter
+          names which are on the ``function``. Collections and type variables
+          are also permitted so long as they contain semantic types.
+        parameters : dict[str, type expression]
+          A dictionary of function0parameter names to primitive type
```
cherman2:
```suggestion
          A dictionary of function parameter names to primitive type
```

cherman2 (approved):
LGTM. I have just reviewed "code" and it all makes sense to me. I will pull down the render if you want me to look at that  @ebolyen

qiime2/core/type/meta.py
ebolyen: Probably, but it might be better as a "how to" as this docstring is already really long.

qiime2/core/type/semantic.py
ebolyen: Unfortunately no, exclude is a different kind of opposite than property itself which is the opposite of union

qiime2/plugin/__init__.py
ebolyen: yep!

qiime2/plugin/__init__.py
ebolyen: I think of it more as "using" a predicate, since the predicate is literally an attribute attached to the whole type expression. i.e. 
```python
>>> type = Int % Range(0, 5)
>>> type.predicate
Range(0, 5)
```

qiime2/sdk/usage.py
ebolyen: The parameter name is `usage` not `use` in the `__init__`. We generally will name the first argument of an example `use` so I think that's why it was named wrong here.

ebolyen:
Shoot. I was trying to batch these, but apparently some diffs were large and thus hidden. I don't get the point of that.

qiime2/plugin/plugin.py
```diff
@@ -178,8 +347,50 @@ def decorator(validator):
...
-        A transformer has the type Callable[[type], type]
+        """ **Decorator** which registers a transformer to convert data
+
+        This decorator may be used with or without arguments.
+
+        Parameters
+        ----------
+        _fn : Callable
+            Ignore this parameter as it is the mechanism to allow argumentless
+            decoration
```
ebolyen:
```suggestion
          Ignore this parameter as it is the mechanism to allow argumentless
          decoration
```

qiime2/plugin/plugin.py
ebolyen: Yeah, but fixed anyway. I went with an indent of `2` above to match most of the indents.

qiime2/plugin/plugin.py
ebolyen: Nope. Hopefully the parameter section clarifies what that means a bit...

qiime2/plugin/plugin.py
ebolyen: lol oops. You can tell I was "done." with all of the typing and then copy-pasted.

ebolyen:
Time to re-wrap the text to 79 chars

ebolyen:
Thanks for the review @cherman2! I reworked the language on the typemap row which doesn't exist a bit in https://github.com/qiime2/qiime2/pull/773/commits/a059f867ada3b173a6489dc766e8f5134458b756, let me know if that's clearer.

qiime2/core/type/meta.py
cherman2: Makes sense to me!

qiime2/plugin/plugin.py
cherman2: It does. I wasn't sure if you wanted to address it before you used it in a sentence but I definitely understood before i was done with the doc string

cherman2:
> Thanks for the review @cherman2! I reworked the language on the typemap row which doesn't exist a bit in [a059f86](https://github.com/qiime2/qiime2/commit/a059f867ada3b173a6489dc766e8f5134458b756), let me know if that's clearer.

yeah that clarifies things for me!

cherman2:
Also, I have no power of merge here, so you will have to merge

Issue #759, opened 2024-03-18
uninformative error message when wrong type is passed for citations list


gregcaporaso:
In hindsight this is obvious, but took me a few mins to figure out. 

If accidentally providing a citation, rather than a list of citations, when registering plugins or actions, the error message is uninformative. For example

```python
plugin = Plugin(
    name="dwq2",
    version=__version__,
    website="https://cap-lab.bio/developing-with-qiime2/",
    package="q2_dwq2",
    description=description,
    short_description="Developing with QIIME 2 demo plugin.",
    # the following should be citations=[citations['DWQ2']]
    citations=citations['DWQ2']
)
```

```bash
$ qiime dwq2 --citations
Traceback (most recent call last):
  File "/Users/jgc/miniconda3/envs/dwq2/bin/qiime", line 11, in <module>
    sys.exit(qiime())
  File "/Users/jgc/miniconda3/envs/dwq2/lib/python3.8/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/Users/jgc/miniconda3/envs/dwq2/lib/python3.8/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/Users/jgc/miniconda3/envs/dwq2/lib/python3.8/site-packages/click/core.py", line 1686, in invoke
    sub_ctx = cmd.make_context(cmd_name, args, parent=ctx)
  File "/Users/jgc/miniconda3/envs/dwq2/lib/python3.8/site-packages/click/core.py", line 943, in make_context
    self.parse_args(ctx, args)
  File "/Users/jgc/miniconda3/envs/dwq2/lib/python3.8/site-packages/q2cli/click/command.py", line 43, in parse_args
    return super().parse_args(ctx, args)
  File "/Users/jgc/miniconda3/envs/dwq2/lib/python3.8/site-packages/click/core.py", line 1644, in parse_args
    rest = super().parse_args(ctx, args)
  File "/Users/jgc/miniconda3/envs/dwq2/lib/python3.8/site-packages/click/core.py", line 1408, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "/Users/jgc/miniconda3/envs/dwq2/lib/python3.8/site-packages/click/core.py", line 2400, in handle_parse_result
    value = self.process_value(ctx, value)
  File "/Users/jgc/miniconda3/envs/dwq2/lib/python3.8/site-packages/click/core.py", line 2362, in process_value
[...]

PR #774, opened 2024-07-01
Citations single record

colinvwood:
addresses #759

### DIFF
--- a/qiime2/core/cite.py
+++ b/qiime2/core/cite.py
@@ -26,7 +26,7 @@ fields: dict
 """
 
 
-def make_citations_tuple(citations):
+def _make_citations_tuple(citations):
     if citations is None:
         return tuple()
     elif isinstance(citations, CitationRecord):
--- a/qiime2/core/cite.py
+++ b/qiime2/core/cite.py
@@ -15,6 +15,19 @@ import bibtexparser as bp
 CitationRecord = collections.namedtuple('CitationRecord', ['type', 'fields'])
 
 
+def make_citations_tuple(citations):
+    '''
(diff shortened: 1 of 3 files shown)


ebolyen:
@colinvwood, sorry I made some conflicts in #773.

While you are at it, the docstrings explaining how to use `citations` can be updated to indicate a single citation record is acceptable.

qiime2/core/cite.py
```diff
@@ -26,6 +26,16 @@
 """
 
 
+def make_citations_tuple(citations):
```
gregcaporaso: @colinvwood, any reason to not make this a private (`_make_citations_tuple`) function?

gregcaporaso (changes requested):
@colinvwood, could add a test by making an Action in one of the test plugins provide a single citation that's not in a `list`?

gregcaporaso:
I tested this locally and it works as expected, so this is ready to go pending addressing my review comments.

qiime2/core/cite.py
colinvwood: I guess because it's used outside of the module it's defined in. But we mainly use underscores just to differentiate between the public api and everything else huh? I can change it in that case.

colinvwood:
> @colinvwood, could add a test by making an Action in one of the test plugins provide a single citation that's not in a `list`?

I changed all registrations in `core.testing.plugin` that register a single citation to drop the list.

qiime2/core/cite.py
gregcaporaso: Yeah, that's the typical way we use it.

PR #775, opened 2024-07-10
Parallel fixes

Oddant1:


ebolyen (approved):
Looks good, and it seems to even have simplified things a bit.

As a future enhancement, it does appear that _bind_parsl isn't exactly a "bind" method, as it will call things outright, but that seems to have always been the case. A nice refactor in the future would be to have _bind_parsl return a callable which is then evaluated. I suspect that will further simplify things, and now that the run_parsl_action is in the closure already, we're halfway there.

PR #779, opened 2024-07-18
Make bind_parsl more bind like

Oddant1:
Make _bind_parsl look more like _bind


ebolyen:
Going to close this in favor of #820

PR #780, opened 2024-07-24
Add new test config

Oddant1:
Add test config to be used by plugins that do not set QIIMETEST


lizgehret:
lgtm, thanks @Oddant1!

PR #782, opened 2024-07-25
Add test config to TestPluginBase

Oddant1:
Move the test config to TestPluginBase


Oddant1:
@lizgehret, @ebolyen actually had a better idea. Use TestPluginBase which I had forgotten existed.

ebolyen (approved):

PR #783, opened 2024-08-05
type -> view_type

Oddant1:


Oddant1:
@ebolyen the real Artifact has a `_view` and a `view`. the` _view` has an extra kwarg. Should we replicate that pattern here? I'm assuming not. My guess would be we discussed that back in the day and decided not to, but I don't remember. If your answer is "no" again I'll leave a comment explaining why we only have 'view' here

ebolyen (approved):

PR #786, opened 2024-08-08
BUG: `Artifact.import_data` doesn't store checksums when the `view_type` is a `FormatBase`

### LINKED ISSUE #756: `Artifact.import_data` doesn't store checksums when the `view_type` is a FormatBase
**Bug Description**
Generally it's expected that file-formats will have a checksum performed during import (via `Artifact.import_data`), but if a caller provides `view_type` as a python class, the checksums are skipped even if the `view_type` is a file or directory format.

This currently impacts the Galaxy imports as they provide the view_type explicitly as their respective classes.

The CLI appears to be unimpacted because it passes `view_type` as a string which is then always assumed to be a format.

All of the logic here is a bit suspect and could be cleaned up.

**References**
1. https://github.com/qiime2/qiime2/blob/dev/qiime2/sdk/result.py#L301
2. https://github.com/qiime2/qiime2/blob/dev/qiime2/sdk/result.py#L320-L331

lizgehret:
fixes #756 

Still needs tests added. cc @ebolyen @Oddant1 does this seem reasonable?

### DIFF
--- a/qiime2/sdk/result.py
+++ b/qiime2/sdk/result.py
@@ -322,7 +322,9 @@ class Artifact(Result):
 
         format_ = None
         md5sums = None
-        if is_format:
+        if is_format and issubclass(view_type, qiime2.core.format.FormatBase):
+            path = pathlib.Path(str(view))
+        elif is_format:
             path = pathlib.Path(view)
             if path.is_file():
                 md5sums = {path.name: util.md5sum(path)}
--- a/qiime2/sdk/result.py
+++ b/qiime2/sdk/result.py
@@ -322,10 +322,12 @@ class Artifact(Result):
 
         format_ = None
         md5sums = None
-        if is_format and issubclass(view_type, qiime2.core.format.FormatBase):
-            path = pathlib.Path(str(view))
-        elif is_format:
-            path = pathlib.Path(view)
+        if is_format:
+            if issubclass(view_type, qiime2.core.format.FormatBase):
+                path = pathlib.Path(str(view))
+            else:
+                path = pathlib.Path(view)
+
             if path.is_file():
                 md5sums = {path.name: util.md5sum(path)}
             elif path.is_dir():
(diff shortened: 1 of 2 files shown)


Oddant1:
I think once you hit [here](https://github.com/qiime2/qiime2/blob/7dc2ddb5a91bdbf80102baa95d935cde2235fdf8/qiime2/sdk/result.py#L321) if you have a FormatBase you're going to need to cast it to str. This seems sane in general though.

NVM this

qiime2/sdk/result.py
```diff
@@ -317,7 +322,9 @@ def import_data(cls, type, view, view_type=None, validate_level='max'):
 
         format_ = None
         md5sums = None
-        if is_format:
+        if is_format and issubclass(view_type, qiime2.core.format.FormatBase):
+            path = pathlib.Path(str(view))
+        elif is_format:
             path = pathlib.Path(view)
             if path.is_file():
                 md5sums = {path.name: util.md5sum(path)}
```
```suggestion
        if is_format:
            if issubclass(view_type, qiime2.core.format.FormatBase):
                path = pathlib.Path(str(view))
            else:
                path = pathlib.Path(view)

            if path.is_file():
                md5sums = {path.name: util.md5sum(path)}
```

colinvwood:
awaiting tests?

qiime2/tests/test_artifact_api.py
```diff
@@ -118,6 +118,20 @@ def test_reload_fails(self):
...
+    def test_import_base_format_stores_checksums(self):
+        from qiime2 import Artifact
+        from qiime2.core.testing.type import IntSequence2
+        from qiime2.core.testing.format import IntSequenceFormat
+        from qiime2.core.util import load_action_yaml
+        from qiime2.plugin.util import transform
+
+        ff = transform([1, 2, 3,], to_type=IntSequenceFormat)
+        ff2 = Artifact.import_data(IntSequence2, ff,
+                                   view_type='IntSequenceFormat')
```
ebolyen: I think this line needs to be a class (instead of the name as a string) in order for the original bug to appear.

You can verify by "un-fixing" the PR and verifying the test fails.

ebolyen (approved):

PR #790, opened 2024-08-14
Alias as join app

Oddant1:
Supersedes #785

### DIFF
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -1219,6 +1219,9 @@ class Cache:
             os.remove(target)
 
     def get_tmp_path(self):
+        """Creates a tmp dir inside of the current process pool and returns a
+        path to it. If the pool already exists just returns the path
+        """
         path = os.path.join(self.process_pool.path, 'tmp')
         os.makedirs(path, exist_ok=True)
         return path
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -1218,6 +1218,11 @@ class Cache:
         if target.exists():
             os.remove(target)
 
+    def get_tmp_path(self):
+        path = os.path.join(self.process_pool.path, 'tmp')
+        os.makedirs(path, exist_ok=True)
+        return path
+
     @property
     def data(self):
         """The directory in the cache that stores the data.
@@ -1837,8 +1842,3 @@ class Pool:
 
         item = IndexedCollectionElement(item_name, int(idx), int(total))
         outputs[output_name][item] = value
-
-    def get_tmp_path(self):
-        path = os.path.join(self.path, 'tmp')
-        os.makedirs(path, exist_ok=True)
-        return path
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -1837,3 +1837,9 @@ class Pool:
 
         item = IndexedCollectionElement(item_name, int(idx), int(total))
         outputs[output_name][item] = value
+
+    def get_tmp_path(self):
+        path = os.path.join(self.path, 'tmp')
+        if not os.path.isdir(path):
+            os.makedirs(path)
+        return path
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -1840,6 +1840,5 @@ class Pool:
 
     def get_tmp_path(self):
         path = os.path.join(self.path, 'tmp')
-        if not os.path.isdir(path):
-            os.makedirs(path)
+        os.makedirs(path, exist_ok=True)
         return path
--- a/qiime2/core/path.py
+++ b/qiime2/core/path.py
@@ -90,7 +90,7 @@ class OutPath(OwnedPath):
         else:
             os.unlink(path)
 
-    def __new__(cls, dir=False, **kwargs):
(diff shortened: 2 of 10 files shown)


Oddant1:
Supposedly Python multiprocessing subprocesses exit with os.exit which isn't supposed to call either atexit funcs or weakref finalizers, but it seems like when the parsl subprocesses exit they call the weakref finalizers but not the registered atexit func

Oddant1:
Check failure modes on multi node systems for did not follow the rules and set a global cache.

Check what happens when the inode limit is low. Are we exhausting a bunch of inodes by changing the way scope works

qiime2/core/cache.py
```diff
@@ -1837,3 +1837,9 @@ def _add_collection_index_output(self, outputs, name, value):
 
         item = IndexedCollectionElement(item_name, int(idx), int(total))
         outputs[output_name][item] = value
+
+    def get_tmp_path(self):
+        path = os.path.join(self.path, 'tmp')
+        if not os.path.isdir(path):
+            os.makedirs(path)
```
```suggestion
        os.makedirs(path, exist_ok=True)
```

qiime2/core/path.py
```diff
@@ -96,10 +94,16 @@ def __new__(cls, dir=False, **kwargs):
...
         """
+        from qiime2.core.cache import get_cache
+
+        cache = get_cache()
+        tmp_path = cache.process_pool.get_tmp_path()
+        prefix = 'q2-%s-' % cls.__name__
+
         if dir:
-            name = tempfile.mkdtemp(**kwargs)
+            name = tempfile.mkdtemp(prefix=prefix, dir=tmp_path)
```
ebolyen, line 104: I would remove `**kwargs` from the signature of `__new__` if it isn't used anymore.

qiime2/sdk/action.py
```diff
@@ -108,6 +108,22 @@ def _unmap_arg(arg, inputs):
     return unmapped
 
 
+@join_app
+def _deferred_alias(provenance, name, output, ctx, inputs=[]):
+    output = output.result()
+    return qiime2.sdk.util.create_future(
+        _alias(provenance, name, output, ctx))
```
ebolyen: I would push the result of _alias up out of the expression. It blocks (which it should), but it looks at first like an expression that is minting a closure, (i.e. `map(make_apply(a, b,c), ...)`) rather than evaluating in place.

```python
output = output.result()
alias = _alias(provenance, name, output, ctx)
reutrn qiime2.sdk.util.create_future(alias)
```

qiime2/sdk/action.py
```diff
@@ -409,7 +419,8 @@ def _run_parsl_action(action, ctx, execution_ctx, mapped_args,
             # in the right cache
             with ctx.cache:
                 exe = action._bind(
-                    lambda: qiime2.sdk.Context(parent=ctx), execution_ctx)
+                    lambda: ctx if ctx._parent is None
+                    else qiime2.sdk.Context(parent=ctx), execution_ctx)
```
ebolyen: It isn't clear when the `_parent` isn't `None`.  From poking around it seems like perhaps everything ends up in the same parent context? I guess that could work, now that destructors don't exist.

qiime2/sdk/result.py
```diff
@@ -238,8 +238,13 @@ def clone_original(into):
...
             into.rmdir()
-            shutil.copytree(str(self._archiver.data_dir), str(into),
-                            copy_function=os.link)  # Use hardlinks
+            try:
+                shutil.copytree(str(self._archiver.data_dir), str(into),
+                                copy_function=os.link)  # Use hardlinks
+            except shutil.Error:
+                # Try again with full copy not links
+                shutil.copytree(str(self._archiver.data_dir), str(into),
+                                dirs_exist_ok=True)
```
ebolyen:
```suggestion
                                copy_function=qiime2.util.duplicate)
```

I don't know for certain that this will work, but it seems like it ought to. (and might inherit future symlink behavior).

qiime2/sdk/util.py
```diff
@@ -188,7 +188,7 @@ def get_available_usage_drivers() -> Dict[str, 'UsageDriver']:
     }
 
 
-@python_app
+@python_app(executors=['default'])
```
ebolyen, line 191: Should this actually be the executor of the current app?

ebolyen (commented):
A few questions and minor suggestions.

qiime2/sdk/util.py
Oddant1: I was wondering about that. I don't feel like it really matters? I only set it to default here to make it at least concretely defined.

qiime2/sdk/action.py
Oddant1: It is possible we have ended up in that state yeah... And yeah I'm not sure if we care about that if Context is no longer cleaning up

Oddant1: Ok so we are setting parent contexts when we call an action inside of a pipeline. This is used to identify when we have hit the end of the root pipeline so we know when to wait for our results.

ebolyen (approved):
Thanks @Oddant1!

PR #791, opened 2024-08-20
MAINT: update test URL to non-deprecated address

lizgehret:
The `test_init_metadata_from_url_error_on_non_metadata` test is failing due to https://www.qiime2.org/ no longer being an active URL. I've replaced the inactive link with the user docs URL.


lizgehret:
I'm going to go ahead and merge this because this is blocking new development builds for all distributions.

PR #797, opened 2024-09-10
API: adds `qiime.util.get_filepath_from_package`

### LINKED ISSUE #792: add `qiime2.util.get_filepath_from_package` API?
[This](https://github.com/qiime2/q2-fmt/blob/b8161decabb490628abc4b39ba468496fc793ce4/q2_fmt/_examples.py#L15) is a pretty useful utility function for usage examples. Should we add a `Usage.get_filepath_from_tests` API?

gregcaporaso:
fixes #792


qiime2/plugin/testing.py
```diff
@@ -110,16 +109,23 @@ def get_data_path(self, filename):
         ----------
         filename : str
             The name of the file to look up.
+        result_as_str : bool, True
```
lizgehret, line 112: This is really convenient 👌🏼

qiime2/util.py
```diff
@@ -114,3 +115,32 @@ def duplicate(src, dst):
...
+    FileNotFoundError
+        If the filepath does not exist.
+
+    """
+    fp = importlib.resources.files(package) / relative_filepath
+    if not fp.exists():
+        raise FileNotFoundError(
+            f'The requested data asset {fp} does not exist.')
+    else:
+        return fp
```
lizgehret, line 146: Is there a reason you didn't include the `result_as_str` param in this implementation? Seems like it would be convenient since you're returning all of the test cases above as str(fp).

lizgehret (commented):
overall this lgtm @gregcaporaso - just one small question inline!

qiime2/util.py
gregcaporaso: @lizgehret, i was doing that in the `TestPluginBase.get_data_path` just to maintain its API - I suspect that's the only one of these `get_data_path` methods that is being called from outside of the framework. 

My goal is that users of this new function get a `pathlib.FilePath` object back since I think that's the type that makes the most sense (hence no option for a string), but for all of the places where `TestPluginBase.get_data_path` is already used, it makes sense to have that continue returning a string so we don't break other code. 

Let me know if you want to discuss more.

lizgehret: Gotcha, that makes sense!

lizgehret (approved):

PR #798, opened 2024-09-17
Disallow List output registration

Oddant1:
Registering a `List` return type is meaningless because we will always write a `Collection` in the end. Disallowing this fell through the cracks before.

bokulich-lab/q2-moshpit#205


ebolyen (approved):
LGTM

For the sake of CI, let's merge this only after the update has happened in q2-moshpit

PR #800, opened 2024-09-17
API: adds `qiime.util.get_filepath_from_package`

lizgehret:
Reverts qiime2/qiime2#799

### DIFF
--- a/qiime2/metadata/tests/test_io.py
+++ b/qiime2/metadata/tests/test_io.py
@@ -8,7 +8,6 @@
 
 import collections
 import os.path
-import pkg_resources
 import tempfile
 import unittest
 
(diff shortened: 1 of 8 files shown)


lizgehret:
@gregcaporaso and I discussed offline about updating the implementation of this helper within `TestPluginBase` to catch the `FileNotFoundError` and proceed. This allows the associated unit tests in q2-types to still run through the validation step without failing on FileNotFound.

gregcaporaso:
After some discussion we decided to update the test to align with this API change. Our perspective is that raising the error is how the function _should_ work, and it's highly unlikely that tests are expecting the requested file not to exist outside of the framework or q2-types, so we're going to treat this as a bug fix.

gregcaporaso:
Depends on merge of https://github.com/qiime2/q2-types/pull/343 for integration tests to pass.

lizgehret:
Tests all passed on https://github.com/qiime2/q2-types/pull/343 and it's been merged into dev. I'll go ahead and merge this as well, thanks @gregcaporaso!

gregcaporaso:
Thanks @lizgehret!

PR #801, opened 2024-09-20
Only link in named pool if path still exists

### LINKED ISSUE #787: default recycle-pools are incompatible with simultaneous execution of the same action
I ran into this when running `qiime boots core-metrics` with a handful of different parameter settings. Two work-arounds are:

1) Specify the name of the recycle pool for each individual command, ensuring that the names of the recycle-pools are different (e.g., `--recycle-pool output-1` and `--recycle-pool output-2`). The catch here is the user is responsible for cleaning this up as QIIME 2 won't garbage collect user-named recycle pools.
2) Specify `--no-recycle` (in which case pipeline resumption won't be available).

Oddant1:
closes #787

### DIFF
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -1314,14 +1314,13 @@ class Cache:
         if self._named_pool_ is None:
             return None
 
-        if os.path.exists(self._named_pool_.path):
-            return self._named_pool_
-
-        if not self._named_pool_._warned_:
+        if not os.path.exists(self._named_pool_.path):
             warnings.warn("The named pool path"
                           f" '{self._named_pool_.path}' does not exist. It was"
                           " most likely removed by another QIIME 2 process")
-            self._named_pool_._warned_ = True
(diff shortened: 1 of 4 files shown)


Oddant1:
~~Blocking qiime2/q2cli#345~~

qiime2/core/cache.py
```diff
@@ -1194,7 +1194,11 @@ def _alias(self, uuid):
 
             # Named pool links are not aliased
             if self.named_pool is not None:
-                self.named_pool._make_symlink(uuid, uuid)
+                if not os.path.exists(self.named_pool.path):
+                    warnings.warn("The named pool path"
```
Oddant1: I don't think we actually want to set named pool to None after this warning because it has potential knock on effects later down the line when other bits of code are expecting there to not be a none for the named pool

colinvwood (commented):
Pending small refactor to `named_pool` property

colinvwood (approved):

PR #803, opened 2024-10-03
plugin-action

### LINKED ISSUE #802: require plugin name when specifying parallel Executor mappings
Currently only the action name is required, which may become a problem if different plugins define actions with the same name. 

So, an executor mapping like the following:

```python
em = {'summarize': 'tpool'}
```

or 

```toml
[parsl.executor_mapping]
summarize = "tpool"
```

would become something like:

```python
em = {'feature-table.summarize': 'tpool'}
```

or 

```toml
[parsl.executor_mapping]
feature-table.summarize = "tpool"
```

Oddant1:
Closes #802.

The `.` is a reserved character in toml, so I went with `plugin-action`

### DIFF
--- a/qiime2/sdk/tests/data/mapping_only_config.toml
+++ b/qiime2/sdk/tests/data/mapping_only_config.toml
@@ -2,3 +2,7 @@
 
 [parsl.executor_mapping.dummy_plugin]
 list_of_ints = "test"
+concatenate_ints = "test"
+
+[parsl.executor_mapping.other_plugin]
+concatenate_ints = "test"
--- a/qiime2/sdk/tests/data/mapping_only_config.toml
+++ b/qiime2/sdk/tests/data/mapping_only_config.toml
@@ -1,4 +1,4 @@
 [parsl]
 
 [parsl.executor_mapping]
-list_of_ints = "test"
+dummy_plugin-list_of_ints = "test"
--- a/qiime2/sdk/tests/data/mapping_only_config.toml
+++ b/qiime2/sdk/tests/data/mapping_only_config.toml
@@ -1,4 +1,4 @@
 [parsl]
 
-[parsl.executor_mapping]
-dummy_plugin-list_of_ints = "test"
+[parsl.executor_mapping.dummy_plugin]
+list_of_ints = "test"
--- a/qiime2/sdk/tests/data/mapping_only_config.toml
+++ b/qiime2/sdk/tests/data/mapping_only_config.toml
@@ -5,4 +5,4 @@ list_of_ints = "test"
 concatenate_ints = "test"
 
 [parsl.executor_mapping.other_plugin]
-concatenate_ints = "test"
+concatenate_ints = "default"
--- a/qiime2/sdk/tests/test_config.py
+++ b/qiime2/sdk/tests/test_config.py
@@ -37,7 +37,7 @@ class TestConfig(unittest.TestCase):
             'concatenate_ints': 'test'
         },
         'other_plugin': {
-            'concatenate_ints': 'test'
+            'concatenate_ints': 'default'
         }
     }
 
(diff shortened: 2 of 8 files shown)


gregcaporaso:
@Oddant1, @ebolyen - did you two discuss this one? Just want to make sure @ebolyen is good with the change but I'm happy to review after that.

gregcaporaso:
@Oddant1, the `-` doesn't feel super intuitive to me. Would `:` work instead? Or maybe these should involve an additional layer of nesting? For example:

```toml
[parsl.executor_mapping.dummy_plugin]
list_of_ints = "test"
```

@ebolyen, what do you think?

ebolyen:
Ah, that may be why this didn't happen the first time around. I like @gregcaporaso's suggestion of either nesting the object using TOML or using a colon in the key.

gregcaporaso:
I feel like nesting is the more intuitive solution - let's do that if you guys feel ok about it.

Oddant1:
@ebolyen and @gregcaporaso it now expects nested dicts of the form 

```toml
[parsl.executor_mapping.dummy_plugin]
list_of_ints = "test"
```

Which in Python becomes 

```python
{'dummy_plugin': {'list_of_ints': 'test'}}
```

So if you are providing a mapping in the Python API you just create a dictionary that looks like the above.

Oddant1:
I'm going to add some multi mapping to the dicts used in the tests. Gonna buff em up a bit and make it clear what things should look like with multiple plugins/actions.

Oddant1:
@ebolyen and @gregcaporaso should be gtg

qiime2/sdk/tests/test_config.py
```diff
@@ -123,10 +137,8 @@ def test_mapping_only_config(self):
         self.assertEqual(dict_execution_contexts, self.tpool_expected)
 
     def test_mapping_from_dict(self):
```
gregcaporaso, line 141: @Oddant1, what happens if you use the old format now (where the plugin name isn't specified). I think its worth testing that that throws an informative error message, just in case there are instances of that happening out in the wild (since this will be a backward incompatible change - but I think we bill it as a bug fix, since the previous behavior was that the same executor would be used for any `Pipelines` with the same name, regardless of what plugin they are in).

qiime2/sdk/tests/data/mapping_only_config.toml
```diff
@@ -1,4 +1,8 @@
 [parsl]
 
-[parsl.executor_mapping]
+[parsl.executor_mapping.dummy_plugin]
 list_of_ints = "test"
+concatenate_ints = "test"
+
+[parsl.executor_mapping.other_plugin]
+concatenate_ints = "test"
```
gregcaporaso:
```suggestion
concatenate_ints = "test3"
```

qiime2/sdk/tests/data/mapping_only_config.toml
```diff
@@ -1,4 +1,8 @@
 [parsl]
 
-[parsl.executor_mapping]
+[parsl.executor_mapping.dummy_plugin]
 list_of_ints = "test"
+concatenate_ints = "test"
```
gregcaporaso, line 5: Just so we can confirm the right executors are being associated with the right actions, these tests should reference different executors. 

```suggestion
concatenate_ints = "test2"
```

qiime2/sdk/tests/test_config.py
```diff
@@ -31,6 +31,16 @@ class TestConfig(unittest.TestCase):
     pipeline = plugin.pipelines['resumable_pipeline']
     method = plugin.methods['list_of_ints']
 
+    mapping = {
+        'dummy_plugin': {
+            'list_of_ints': 'test',
+            'concatenate_ints': 'test'
```
gregcaporaso, line 37:
```suggestion
            'concatenate_ints': 'test2'
```

qiime2/sdk/tests/test_config.py
```diff
@@ -31,6 +31,16 @@ class TestConfig(unittest.TestCase):
...
+        },
+        'other_plugin': {
+            'concatenate_ints': 'test'
```
gregcaporaso:
```suggestion
            'concatenate_ints': 'test3'
```

gregcaporaso (changes requested):
Thanks @Oddant1 - just a couple of minor things.

Oddant1:
@gregcaporaso using the old format should now raise an informative error, and I am now testing that concatenate_ints in dummy and concatenate_ints in other are properly being mapped to different executors as requested

ebolyen (approved):

gregcaporaso:
Thanks for wrapping it up @Oddant1 and for merging @ebolyen!

PR #812, opened 2024-11-14
TEST: Skip this test for now because it is holding up Mac builds. Revert later

Oddant1:
Mac builds are currently being held up by this test which fails with either a time out or seg fault on M series Macs. For now, we will skip this test.

The currently set deadline for figuring this out and making this test work is December 13th.

### DIFF
--- a/qiime2/sdk/tests/test_config.py
+++ b/qiime2/sdk/tests/test_config.py
@@ -9,8 +9,11 @@
 import os
 import tempfile
 import unittest
+import platform
+from datetime import datetime
 
 import parsl
+import pytest
 from parsl.executors.threads import ThreadPoolExecutor
 from parsl.errors import NoDataFlowKernelError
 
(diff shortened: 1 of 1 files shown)


qiime2/sdk/tests/test_config.py
```diff
@@ -215,6 +218,10 @@ def test_no_vendored_fp(self):
             self.assertEqual(list_execution_contexts, self.tpool_expected)
             self.assertEqual(dict_execution_contexts, self.tpool_expected)
 
+    @pytest.mark.skipif(platform.system() == 'Darwin' and
+                        datetime.today().strftime('%m-%d') < '12-13',
```
ebolyen: This will technically go back to passing in Jan

Oddant1: I figured we would have killed this by then, but I can change it if you would like

PR #815, opened 2024-11-25
REF: Plugin class typing change

### LINKED ISSUE #809: Typing issue with `Plugin` attributes
The `methods`, `visualizers`, and `pipelines` class attributes of the `Plugin` class are typed as classes ([here](https://github.com/qiime2/qiime2/blob/ab00603cdf39be5866ad58463492984b699d76a9/qiime2/plugin/plugin.py#L104)) but the methods called on these class attributes (e.g. `register_function`) are instance methods, and the attributes themselves are instances when used in `plugin_setup.py` files. Thus e.g. `Type['PluginMethods']` should probably just be typed as 'PluginMethods'.

colinvwood:
Addresses #809

### DIFF
--- a/qiime2/plugin/plugin.py
+++ b/qiime2/plugin/plugin.py
@@ -60,7 +60,7 @@
 import collections
 import inspect
 import types
-from typing import Any, Optional, Union, Type
+from typing import Any, Optional, Union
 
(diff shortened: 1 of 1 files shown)


gregcaporaso:
I think this is ready for merge based on our test of the documentation build - @ebolyen, is that right?

PR #818, opened 2024-11-28
REF: dont instantiate plugin manager in class definition

colinvwood:
Addresses qiime2/q2cli#334. Does not seem to resolve the command line help text lag however.

### DIFF
--- a/qiime2/core/archive/provenance_lib/replay.py
+++ b/qiime2/core/archive/provenance_lib/replay.py
@@ -76,7 +76,7 @@ class ReplayConfig:
         self.header = header
         self.verbose = verbose
         self.md_out_dir = md_out_dir
-        self.pm = PluginManager()
+        self.pm = PluginManager.reuse_existing()
 
 
 @dataclass
--- a/qiime2/core/archive/provenance_lib/replay.py
+++ b/qiime2/core/archive/provenance_lib/replay.py
@@ -28,7 +28,7 @@ from qiime2.sdk.util import camel_to_snake
 
 
 @dataclass
-class ReplayConfig():
+class ReplayConfig:
     '''
     Dataclass that stores various user-selected configuration options and
     other bits of information relevant to provenance replay.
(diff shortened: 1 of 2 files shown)


qiime2/core/archive/provenance_lib/replay.py
```diff
@@ -42,8 +42,6 @@ class ReplayConfig():
         disk in .tsv format.
     use_recorded_metadata : bool
         If True, replay should use the metadata recorded in provenance.
-    pm : PluginManager
-        The active instance of the QIIME 2 PluginManager.
```
ebolyen, line 46: I think we should probably keep this parameter, it can just be set to `None` by default, in which case it will instantiate. Otherwise, we would probably want to use the "find singleton" classmethod (I forget it's name), that way it will use the pre-instantiated manager. But I think the parameter is a lot clearer.

ebolyen (changes requested):
I think the test changes are probably fine, but we should put the parameter back on the config objects (tests should still work the same in the end).

qiime2/core/archive/provenance_lib/replay.py
colinvwood: I think the instantiation will return the preexisting instance if it exists:

```python
>>> pm = PluginManager()
>>> pm2 = PluginManager()
>>> id(pm) == id(pm2)
>>> True
```

ebolyen: I was thinking of  `qiime2.sdk.PluginManager.reuse_existing()` which will error if there isn't one. But I don't think we actually want that, so the constructor (like in your example) is all we need.

ebolyen: Ok, no I remember the thought. Either we leave the param in and instantiate the regular way if it's None.

OR

We remove the param, and use `PluginManager.reuse_existing()` that way the plugin manager has to be initialized up-front (i.e. we won't implicitly initialize).

No strong preference to be honest.

ebolyen (approved):

PR #820, opened 2024-12-03
Split Parallel Concerns onto new Contexts

Oddant1:


qiime2/sdk/proxy.py
```diff
@@ -95,6 +114,10 @@ def validate(self, level=NotImplemented):
 class ProxyArtifact(ProxyResult):
     """This represents a future Artifact that is being returned by a Parsl app
     """
+    def _view(self, view_type, recorder):
+        return self._get_element_(
```
ebolyen: I think this and `view()` should be calling `self.result()` right? Then we alias. Whereas right now, it'll be a view of the non-aliased version.

qiime2/sdk/parallel_context.py
```diff
@@ -0,0 +1,247 @@
...
+    # We add this future to the list and create a new proxy with its index as
+    # its future.
+    if isinstance(arg, Proxy):
+        futures.append(arg._future_)
+        mapped = arg.__class__(len(futures) - 1, arg._selector_)
+    # We do the above but for all elements in the collection
+    elif isinstance(arg, list):
+        mapped = []
+
+        for proxy in arg:
```
ebolyen: I know this is older code, but the guts of this for-loop and the dict can be replaced with a recursive comprehension:

```python
if isinstance(arg, Proxy):
    ...
elif isinstance(arg, list):
    mapped = [_map_arg(proxy, futures) for proxy in arg]
elif isinstance(arg, dict):
    mapped = {key: _map_arg(proxy, futures) for key, proxy in arg.items()}
```

qiime2/sdk/parallel_context.py
```diff
@@ -0,0 +1,247 @@
...
+            else:
+                mapped[key] = value
+    # We just have a real artifact and don't need to map
+    else:
+        mapped = arg
+
+    return mapped
+
+
+def _unmap_arg(arg, inputs):
```
ebolyen: Same deal here essentially.

qiime2/sdk/parallel_context.py
```diff
@@ -0,0 +1,247 @@
...
+            self.executor_name_type_mapping = parent.executor_name_type_mapping
+        else:
+            self.action_executor_mapping = \
+                PARALLEL_CONFIG.action_executor_mapping
+            self.executor_name_type_mapping = \
+                None if PARALLEL_CONFIG.parallel_config is None \
+                else {v.label: v.__class__.__name__
+                      for v in PARALLEL_CONFIG.parallel_config.executors}
+
+    def callable_action(self, *args, **kwargs):
```
ebolyen: This should probably be private, and since it would be called by the parent class, we might give it a special name like `_callable_action_()`.

qiime2/sdk/parallel_context.py
```diff
@@ -0,0 +1,247 @@
...
+        if self.cache.named_pool is not None and \
+                not _contains_proxies(*args, **kwargs) and \
+                (cached_results := self._check_cache(args, kwargs)):
+            return cached_results
+
+        # If we didn't have cached results to reuse, we need to execute
+        # the action.
+        return self.dispatch(*args, **kwargs)
+
+    def dispatch(self, *args, **kwargs):
```
ebolyen: Same here, re: special name.

qiime2/sdk/parallel_context.py
ebolyen: Oh. I don't think this needs to exist on this class right? The base-class appears to have the same implementation.

qiime2/sdk/asynchronous_context.py
```diff
@@ -0,0 +1,55 @@
...
+        # function's signature.
+        args = args[1:]
+
+        pool = concurrent.futures.ProcessPoolExecutor(max_workers=1)
+        future = pool.submit(_subprocess_apply, self, args, kwargs)
+        # TODO: pool.shutdown(wait=False) caused the child process to
+        # hang unrecoverably. This seems to be a bug in Python 3.7
+        # It's probably best to gut concurrent.futures entirely, so we're
+        # ignoring the resource leakage for the moment.
+        return future
```
ebolyen: Does this work? I would assume we would need a proxy here. Or otherwise, have some mechanism to roll the entire pipeline/callable into the single dispatch.

ebolyen:
This is so much cleaner, it's really incredible. Parsl now lives entirely in it's own little context (ignoring config).

qiime2/sdk/asynchronous_context.py
```diff
@@ -0,0 +1,55 @@
...
+    exe = ctx.action_obj._bind(
+        lambda: SerialContext(ctx), {'type': 'asynchronous'})
+    results = exe(*args, **kwargs)
+
+    return results
+
+
+class AsynchronousContext(Context):
+    def dispatch(self, *args, **kwargs):
+        # TODO handle this better in the future, but stop the massive error
```
gregcaporaso: Is there an issue or something else you could reference here to provide context on the error?

ebolyen: This was an ancient issue related to this: https://github.com/qiime2/qiime2/issues/73

It may not even exist anymore as a problem.

qiime2/sdk/parallel_context.py
Oddant1: Good call

qiime2/sdk/parallel_context.py
Oddant1: The only difference is the implementation on ParallelContext short circuits if it receives proxies as args. The version on base Context doesn't check for Proxies so it can remain unaware of all parallel stuff.

qiime2/sdk/asynchronous_context.py
Oddant1: Idk we never really supported async much, so I wasn't sure how far we wanted to go with it.

My understanding based on reading the implementation and talking to you about it was that it didn't actually parallelize anything and we didn't care for it to parallelize anything.

It was more of an "I'll execute it when I execute it" type thing. Idk.

qiime2/sdk/parallel_context.py
ebolyen: got it, ok then that's a good reason to override.


We should probably still change the name to `_callable_action_`

qiime2/sdk/asynchronous_context.py
ebolyen: Do we not test async pipelines? Blocking within (if not root) is fine, but I kind of expect a pipeline to break since `_dispatch_` lives inside the center of the returned callable, and it wouldn't expect a future in those circumstances.

ebolyen: Nevermind. Child contexts are synchronous after the first dispatch (from `_suproccess_apply`), so that's why this works. In which case, this is perfectly fine and matches the old API.

qiime2/sdk/asynchronous_context.py
ebolyen: @Oddant1 would you be able to throw a link to the issue as a comment here. I think we can drop the TODO, we're just documenting for posterity at this point.

ebolyen (commented):
Code is looking really great. Just 2 comments (and I need to pull this down and fuss around with it a bit).

ebolyen:
Actually, @Oddant1, how would you feel about creating a `qiime2.sdk.context` module with `__init__.py` being what is currently `context.py` and then we put sync, async, and parsl in there as subpackages? This gives them a nice place to live.

Oddant1:
@ebolyen I had been thinking about splitting the contexts into their own submodule, but I'm not sure I understand exactly how you're asking me to do it here? I did it the... I guess what I would consider to be the normal way in https://github.com/qiime2/qiime2/pull/820/commits/ec693f51b28555c8bae2f699551545e2c14852e1

ebolyen:
Looks about right. I would merge `context.py` into `__init__.py`, but it doesn't really matter.

ebolyen:
We also sometimes use a pattern like `context/base.py` which would also work instead of `context/context.py`

ebolyen (approved):
LGTM. Going to pull it down later today to mess around with.

ebolyen (approved):

PR #821, opened 2024-12-03
NEW: adds archive format v7

### LINKED ISSUE #747: create ArtifactFormat version 7.0
Add items to include here. (Some updates on 3 Dec 2024)

- [ ] cpu architecture and flags
- [x] conda environment info, if available
- [x] file size of each input artifact
- [x] forward compatibility marker of some sort (see semantic versioning discussion below) 
- [ ] support for integrating the following types of annotations (e.g., an `annotations` directory)
  - [x] @gregcaporaso and @ebolyen talked about adding arbitrary notes named something like note-001.md, note-002.md, ..., note-999.md. some reasonable limit would need to be placed on file size for each of these. 
  - [ ] support for adding a data license
  - [ ] support for including a source and/or creator
  - [ ] support for signing

lizgehret:


lizgehret:
Okay well this isn't normal run time:
Screen Shot 2025-01-29 at 7 04 44 PM

Seems weird that adding CPU flags would increase the run time by this much... will investigate further.

lizgehret:
To-do's:
- [x] update archive_parser with v7 additions
- [x] unit tests:
  - [x] semantic versioning
  - [x] conda-env.yaml file
  - [x] datadir-size in action.yaml

qiime2/core/archive/provenance_lib/parse.py
lizgehret, line 1: note that this isn't related to artifact format updates - but this was causing test failures due to nx.subgraph_view API changes.

lizgehret:
blocking https://github.com/qiime2/q2cli/pull/362

qiime2/core/archive/provenance.py
```diff
@@ -457,6 +457,7 @@ def make_transformers_section(self):
     def make_env_section(self):
         env = collections.OrderedDict()
         env['platform'] = platform.platform()
+
```
lizgehret:
```suggestion
```

qiime2/core/annotate.py
```diff
@@ -0,0 +1,322 @@
...
+
+from datetime import datetime
+
+
+class Annotation():
+    """General base class for all Annotation sub-classes.
+
+    Parameters
+    ----------
+    name : str\n
```
ebolyen: Does this cause any issues with numpydoc? I've not usually seen this.

qiime2/core/annotate.py
```diff
@@ -0,0 +1,322 @@
...
+        ValueError\n
+            If no `note.txt` file is found under the
+            corresponding annotation directory.
+
+        """
+        with open(os.path.join(filepath, 'metadata.yaml'), 'r') as fh:
+            meta_yaml = yaml.safe_load(fh)
+            annotation_type = meta_yaml['type']
+
+            if annotation_type == 'Note':
```
ebolyen, line 77: We'll need an `else` branch, otherwise 7.0 reading 7.1 will produce an `UnboundLocalError` on `annotation`.

Raising an error is fine.

qiime2/core/annotate.py
```diff
@@ -0,0 +1,322 @@
...
+        validate_name = name.replace('-', '_')
+        if not validate_name.isidentifier():
+            raise ValueError(f'Name "{name}" is not a valid Python identifier.'
+                             ' Keys may contain `-` characters but must'
+                             ' otherwise be valid Python identifiers. Python'
+                             ' identifier rules may be found here'
+                             ' https://www.askpython.com/python/'
+                             'python-identifiers-rules-best-practices')
+
+    def write_meta_yaml(self, annotations_dir,
```
ebolyen: We might want this to be private, as it will be visible on subclasses.

qiime2/core/annotate.py
```diff
@@ -0,0 +1,322 @@
...
+            os.mkdir(annotations_dir)
+
+        # create the unique dir for a particular annotation
+        annotation_uuid_dirname = \
+            os.path.join(annotations_dir, str(self.uuid))
+        os.mkdir(annotation_uuid_dirname)
+
+        metadata = {
+            'name': self.name,
+            'id': str(self.uuid),
```
ebolyen: I think these names should match to avoid confusion. I don't have a preference either way, we could go with `self.id` in the class or `'uuid'` in the yaml.

qiime2/core/annotate.py
```diff
@@ -0,0 +1,322 @@
...
+
+            with open(filepath, 'r') as fh:
+                # if we hit this branch, self.contents is now set to
+                # whatever is contained in the provided file
+                self.contents = fh.read()
+
+        # Construct Annotation class
+        super().__init__(name)
+
+    def write(self, annotations_dir, root_result_uuid, referenced_result_uuid):
```
ebolyen: this probably also should be private

qiime2/core/annotate.py
```diff
@@ -0,0 +1,322 @@
...
+        # across both input types
+        annotation_uuid_dirname = \
+            self.write_meta_yaml(annotations_dir,
+                                 root_result_uuid,
+                                 referenced_result_uuid)
+
+        # TODO: for now we're going to ignore the potential file system issues
+        # associated with writing from a user provided file, but this will need
+        # to be addressed prior to release via parsing/writing validation test
+        # of some sort
```
ebolyen: what kind of validation are you thinking? it seems like this will work fine, we don't really care what they are writing.

qiime2/core/archive/archiver.py
```diff
@@ -302,11 +303,21 @@ def _destroy_temp_path(cls, process_alias):
...
-        try:
-            imp, fmt_cls = cls._FORMAT_REGISTRY[version].split(':')
-        except KeyError:
-            return None
-        return getattr(importlib.import_module(imp), fmt_cls)
+        if '.' in version:
+            major, minor = version.split('.')
+            minor = int(minor)
+
+            for minor_version in range(minor, -1, -1):
```
ebolyen, line 310: neat! I've not seen this before, but it makes sense.

qiime2/core/archive/archiver.py
```diff
@@ -302,11 +303,21 @@ def _destroy_temp_path(cls, process_alias):
...
+                ver = f'{major}.{minor_version}'
+                if ver in cls._FORMAT_REGISTRY:
+                    imp, fmt_cls = cls._FORMAT_REGISTRY[ver].split(':')
+                    return getattr(importlib.import_module(imp), fmt_cls)
```
ebolyen, line 314: there's a branch here that's correct implicitly, but it may be worth writing it out anyway:

When the for loop fails to find a match, we will skip out of the if and dead-end with a default return of None.
The for-else construct, as dumb as it would look, would spell it out to future readers that this path exists and is intentionally None.

qiime2/core/archive/format/tests/test_util.py
```diff
@@ -81,3 +85,107 @@ def test_write_v4_archive(self):
...
+            'data/file1.txt',
+            'data/file2.txt',
+            'data/nested/file3.txt',
+            'data/nested/file4.txt',
+            'provenance/metadata.yaml',
+            'provenance/VERSION',
+            'provenance/citations.bib',
+            'provenance/conda-env.yaml',
+            'provenance/action/action.yaml',
+            f'provenance/annotations/{note.uuid}/metadata.yaml',
```
ebolyen: I had assumed `annotations` would be in the root. Wouldn't this cause the existing grandcestor logic (which maybe you've changed already) to duplicate annotations within the `provenance/artifacts` directory?

qiime2/core/archive/format/tests/test_util.py
```diff
@@ -81,3 +85,107 @@ def test_write_v4_archive(self):
...
+                zf.read(os.path.join(root_dir, 'provenance', 'annotations',
+                                     f'{note.uuid}', 'note.txt'))
+            checksums = \
+                zf.read(os.path.join(root_dir, 'provenance', 'annotations',
+                                     f'{note.uuid}', 'checksums.sha512'))
+        self.assertRegex(str(version), '^.*archive: 7.0.*$')
+        self.assertRegex(str(metadata), '^.*data-size: .*B.*$')
+        self.assertRegex(str(conda_env), '^.*dependencies:.*- .*$')
+        self.assertRegex(str(annotation_metadata),
+                         f'^.*id: {note.uuid}.*name: mynote.*type: Note.*$')
```
ebolyen: It _might_ be worth matching on newlines between these, which I mention only because `.*` is usually a jerk.

qiime2/core/archive/format/util.py
```diff
@@ -22,3 +25,12 @@ def artifact_version(version):
         yield
     finally:
         Archiver.CURRENT_FORMAT_VERSION = original_version
+
+
+def write_checksums(directory, checksum_file, checksum_type):
```
ebolyen, line 30: wow. I can't believe this didn't exist already, this is a good place for it.

qiime2/core/archive/format/v5.py
```diff
@@ -7,21 +7,26 @@
...
+        super().write(archive_record, type, format,
+                      data_initializer, provenance_capture)
 
-        checksums = md5sum_directory(str(archive_record.root))
-        with (archive_record.root / cls.CHECKSUM_FILE).open('w') as fh:
-            for item in checksums.items():
-                fh.write(to_checksum_format(*item))
-                fh.write('\n')
+        # make sure checksums are written last
+        cls.write_checksums(archive_record)
```
ebolyen, line 24: this is technically happening twice, once in v0, which looks up the implementation below, and then again here, which again looks up the implementation below. Net result is you run the checksums twice (the second time includes provenance as thats now written).

I would drop the call to write_checksums in v0's `write`.

qiime2/core/archive/format/v7_0.py
```diff
@@ -0,0 +1,159 @@
...
+            │   │   ├── metadata.yaml
+            │   │   ├── note.txt
+            │   │   ├── checksums.sha256
+
+        With each uuid representing an individual Annotation object that's been
+        attached to the Result object.
+
+        The `metadata.yaml` file within each Annotation sub-directory contains
+        the following details for a given Annotation:
+
```
ebolyen, line 54: FYI: Stale docs, as there's an `id` (or maybe it'll be `uuid` field as well)

qiime2/core/archive/format/v7_0.py
```diff
@@ -0,0 +1,159 @@
...
+        Annotations that can be added either via the Python API or the cli.
+
+        Supported Annotation sub-types in 7.0:
+
+            Note
+                Can contain inline text or the contents of a file.
+
+        `annotations` directory structure (containing example Notes):
+
+            provenance/
```
ebolyen: Tagging this pending discussion on root vs provenance.

qiime2/core/archive/format/v7_0.py
```diff
@@ -0,0 +1,159 @@
...
+    CHECKSUM_TYPE = CHECKSUM_FILE.split('.')[1]
+
+    # file size converter
+    @staticmethod
+    def _human_readable_size(num, suffix="B"):
+        for unit in ["", "Ki", "Mi", "Gi", "Ti", "Pi", "Ei", "Zi"]:
+            if abs(num) < 1024.0:
+                return f"{num:3.1f} {unit}{suffix}"
+            num /= 1024.0
+        return f"{num:.1f} Yi{suffix}"
```
ebolyen, line 111: lol

qiime2/core/archive/format/v7_0.py
```diff
@@ -0,0 +1,159 @@
...
+
+    @classmethod
+    def write(cls, archive_record, type, format,
+              data_initializer, provenance_capture):
+        # Pulling from the most recent write version that doesn't include
+        # checksums - this ensures that checksums are only calculated
+        # after all requisite files are present.
+        v1.ArchiveFormat.write(archive_record, type, format,
+                               data_initializer, provenance_capture)
```
ebolyen, line 120: V1 calls `super()` so this also suffers from the double-checksum problem. (It gets overwritten, so you don't notice it at the end, but if you throw a print in you'll see it).

qiime2/core/archive/format/v7_0.py
```diff
@@ -0,0 +1,159 @@
...
+
+            if conda_meta_dir.exists():
+                meta_files = \
+                    [file.stem for file in conda_meta_dir.iterdir()
+                     if file.is_file()]
+
+                with conda_fp.open(mode='w') as fh:
+                    fh.write('dependencies:\n')
+                    fh.writelines(f'- {filename}\n'
+                                  for filename in sorted(meta_files))
```
ebolyen: I _think_ you'll want to `filename.rsplit('-', 2)` to separate the versions to make it a technically installable environment.

qiime2/core/archive/format/v7_0.py
```diff
@@ -0,0 +1,159 @@
...
+
+        else:
+            with conda_fp.open(mode='w') as fh:
+                fh.write('error: no conda environment detected.\n')
```
ebolyen, line 143: would it be better to leave it empty, or otherwise drop the file? I'm not really sure what would make the most sense.

I'm fine with this if you prefer it. It's technically yaml, lol

qiime2/core/archive/format/v7_0.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # Add `data-size` field under top-level `metadata.yaml` file
+        data_fp = archive_record.root / cls.DATA_DIR
+        total_size = sum(path.stat().st_size for path in data_fp.iterdir()
```
ebolyen: This won't work when `data/` contains directories, as you'll only get the size of the filesystem's inode for a directory (which we don't care about), rather than the contents.

qiime2/core/archive/provenance_lib/archive_parser.py
```diff
@@ -1082,5 +1108,6 @@ class ParserV6(ParserV5):
     '3': ParserV3,
     '4': ParserV4,
     '5': ParserV5,
-    '6': ParserV6
+    '6': ParserV6,
+    '7.0': ParserV7
```
ebolyen, line 1126: Are we missing logic to backtrack the prov-replay for future versions, `7.1` etc?

qiime2/core/archive/tests/test_archiver.py
```diff
@@ -336,14 +352,27 @@ def test_archive_check(self):
...
-            'checksums.md5',
+            'checksums.sha512',
             'provenance',
             'VERSION'
         ])
 
         observed = set(file for file in archive.relative_iterdir())
         self.assertEqual(observed, expected)
 
+    def test_format_registry(self):
```
ebolyen, line 366: :100:

qiime2/core/util.py
```diff
@@ -118,49 +118,76 @@ def duration_time(relative_delta):
...
 
 
-def md5sum_native(filepath):
-    result = subprocess.run(['md5sum', str(filepath)],
-                            check=True, capture_output=True, text=True)
+def checksum_native(filepath, checksum_type):
+    if checksum_type == 'md5':
+        cmd = ['md5sum', str(filepath)]
+    elif checksum_type == 'sha512':
+        cmd = ['sha512sum', str(filepath)]
```
ebolyen, line 156: Do we want to handle `shasum` or whatever the binary is on OS X? We could leave that for the future I suppose as the python fallback will always work.

qiime2/core/archive/provenance_lib/_checksum_validator.py
```diff
@@ -104,31 +110,36 @@ def validate_checksums(
...
+    if float(archive_version) < 5.0:
         return ChecksumDiff({}, {}, {})
 
+    checksum_ext = _parse_checksum_ext(zf)
+
     root_dir = pathlib.Path(get_root_uuid(zf))
-    checksum_fp = str(root_dir / 'checksums.md5')
-    obs = md5sum_directory_zip(zf)
+    checksum_fp = str(root_dir / f'checksums.{checksum_ext}')
+    obs = checksum_directory_zip(zf, checksum_type=checksum_ext)
```
ebolyen, line 142: There's going to be an issue here, as the annotations are not exempted from the checksum list, so they will always look to be "added" files in the diff, and fail validation.

qiime2/sdk/result.py
```diff
@@ -108,6 +110,8 @@ def load(cls, filepath):
                    type(result).__name__))
 
         result._archiver = archiver
+        result._read_annotations()
```
ebolyen: This isn't the only way to bring up an artifact, so the `_annotations` property won't always be set.

For example `cache.load('my-key')` would skip this line.

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+        """
+        self._annotations = []
+        annotations_dir = self._archiver.annotations_dir
+        # Not sure if there's a better way to do this, but this accounts for
+        # the fact that the annotations_dir will be None for all
+        # previous archive versions < 7.0
+        if annotations_dir and os.path.exists(annotations_dir):
+            for annotation in os.listdir(annotations_dir):
+                self._annotations.append(
+                    Annotation.load(os.path.join(annotations_dir, annotation))
```
ebolyen: We should probably be lazy here, either `Note` defers reading the note/content into memory, or `self._annotations` doesn't instantiate until requested. The latter is probably the best option for performance.

qiime2/sdk/result.py
ebolyen: How about you make `_annotations` a `@property` so that you know that an instance exists, and you can memoize the results, then running `_read_annotations()` in all the correct spots matters way less.

qiime2/core/util.py
```diff
@@ -406,11 +434,11 @@ def ref_constructor(loader, node):
     def cite_constructor(loader, node):
         return node.value
 
-    def metadata_constructor(loader, node):
-        # Use the md5sum of the metadata as its identifier, so we can tell
+    def metadata_constructor(loader, node, checksum_type):
```
ebolyen: We should have seen a failing test for this. The metadata_constructor is used a few lines under this and doesn't have any way to take `checksum_type` as an argument. Also, the metadata entries in provenance are still fixed to md5, so we can keep it md5 here as well.

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+                )
+
+    def add_annotation(self, annotation):
+        """Add an Annotation onto a Result object.
+        All Result-associated parameters are passed into the sub-class's
+        `write` method, while the Annotation instance handles everything else.
```
ebolyen: We probably don't need to include this in the user-facing docs, but putting it in a comment would be good.

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+        Notes
+        -----
+            In Archive Format 7.0, `referenced_result_uuid` is set to
+            the same value as `root_result_uuid`, but this will change
+            in future versions to allow for Annotations that may reference
+            a different Result than the one they are attached to.
+
+        See Also
+        --------
+        Annotation.write
```
ebolyen: unless you want users calling `Annotation.write` directly, I would leave this out and say nothing about it

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+
+        """
+        self._validate_annotation_support()
+        # Guard to ensure Annotation names are unique per Result object
+        for existing_annotation in self._annotations:
+            if annotation.name == existing_annotation.name:
+                raise ValueError(
+                    'Namespace collision occurred when attempting to add '
```
ebolyen: I would go with "duplicate name" or something less technical here.

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+        # now calculate checksums for all files within the newly minted
+        # annotation subdir
+        annotation_dir = \
+            pathlib.Path(self._archiver.annotations_dir) / str(annotation.uuid)
+        checksum_ext = self._archiver._fmt.CHECKSUM_TYPE
+        manifest = self._archiver._fmt.CHECKSUM_FILE
+
+        checksums = util.checksum_directory(str(annotation_dir),
+                                            checksum_type=checksum_ext)
+        checksums.pop(manifest, None)
```
ebolyen: Is there a situation where the annotation possesses a checksum already? I'm guessing its to do with moving a note between two artifacts?

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+            subdir = os.path.join(annotations_dir, entry)
+            meta_filepath = os.path.join(subdir, 'metadata.yaml')
+
+            if os.path.isfile(meta_filepath):
+                with open(meta_filepath, 'r') as fh:
+                    meta = yaml.safe_load(fh)
+
+                if meta.get('name') == name:
+                    annotation_disk_dir = subdir
+                    break
```
Oddant1: For this, we have already loaded the annotation and as such we should have its uuid right? So this should look something like 

`shutil.rmtree(os.path.join(annotations_dir, str(annotation_to_remove.uuid)))`

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+        manifest = self._archiver._fmt.CHECKSUM_FILE
+
+        checksums = util.checksum_directory(str(annotation_dir),
+                                            checksum_type=checksum_ext)
+        checksums.pop(manifest, None)
+
+        with (annotation_dir / manifest).open('w') as fh:
+            for item in checksums.items():
+                fh.write(util.to_checksum_format(*item))
+                fh.write('\n')
```
ebolyen, line 384: It seems like this could all go in `Annotation.write` was it so that you can use `CHECKSUM_TYPE`? You could probably make those arguments to write (presuming it is converted to private) or just hardcode them to sha512.

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+        Raises
+        ------
+        ValueError\n
+            If no Annotations are found associated with the Result.
+
+        """
+        self._validate_annotation_support()
+
+        if self._annotations:
+            return iter(self._annotations)
```
Oddant1: I'd rather this be a tuple than an `iter`, but I don't have strong feelings

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+            If no Annotation with the provided name is found.
+
+        """
+        self._validate_annotation_support()
+
+        for annotation in self._annotations:
+            if annotation.name == name:
+                return annotation
+
+        raise ValueError(f'No Annotation with name: "{name}" was found.')
```
ebolyen: This could be a KeyError if you wanted, but I don't care too much. It might suggest that `_annotations` would do better as a dict, so that you don't have to scan it each time looking for a given name, or a duplicate.

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+    # TODO: add support to filter by type
+    # once additional annotation types are added in 7.1
+    def iter_annotations(self):
+        """Constructs an iterable containing all Annotations associated with
+        the Result object.
+
+        Raises
+        ------
+        ValueError\n
+            If no Annotations are found associated with the Result.
```
ebolyen: This one might be better to not raise an error from, in which case you can implement it as:
```python
yield from self._annotations
```
and that's basically the whole method.

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+                'formats of 7.0 and above.'
+            )
+
+    def _read_annotations(self):
+        """
+        Append any existing Annotations to `self._annotations`.
+        Helper method for `add_annotation`, after a given Annotation
+        has been written to disk.
+        """
+        self._annotations = []
```
Oddant1: Given the amount of times we iterate over this looking for a given name, this should probably be a dict of {name: annotation}

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+                    meta = yaml.safe_load(fh)
+
+                if meta.get('name') == name:
+                    annotation_disk_dir = subdir
+                    break
+
+        # Guard against Annotation name not found on disk
+        if annotation_disk_dir is None:
+            raise ValueError('Unable to locate on-disk directory '
+                             f'for Annotation with name: "{name}"')
```
ebolyen, line 453: weird, but probably a good idea.

qiime2/core/archive/archiver.py
```diff
@@ -462,11 +477,17 @@ def save(self, filepath):
...
             return ChecksumDiff({}, {}, {})
 
-        obs = dict(x for x in md5sum_directory(str(self.root_dir)).items()
-                   if x[0] != self._fmt.CHECKSUM_FILE)
+        obs = \
+            dict(x for x in
+                 checksum_directory(str(self.root_dir),
+                                    checksum_type=self._fmt.CHECKSUM_TYPE)
+                 .items()
+                 if x[0] != self._fmt.CHECKSUM_FILE)
```
ebolyen: We also need to exclude any annotations that may be present, otherwise it will raise an error here with "added" files.

ebolyen (changes requested):
Hey @lizgehret, this looks pretty good in general. There are a few places where we need to change some things, but I don't think any of this is going to look terribly different once fixed.

Some issues that aren't tied to any particular location:

~~1. Provenance won't group the annotations, so they end up scattered within the ancestors.~~ _this has been fixed_
2. Result.validate needs to exempt annotations from its list of validated files. (This is a good reason to move annotations out of provenance, otherwise life gets painful.)

qiime2/sdk/result.py
```diff
@@ -282,6 +286,205 @@ def result(self):
...
+        """
+        Append any existing Annotations to `self._annotations`.
+        Helper method for `add_annotation`, after a given Annotation
+        has been written to disk.
+        """
+        self._annotations = []
+        annotations_dir = self._archiver.annotations_dir
+        # Not sure if there's a better way to do this, but this accounts for
+        # the fact that the annotations_dir will be None for all
+        # previous archive versions < 7.0
```
Oddant1: The implementation looks fine to me. I'd probably leave the comment just change its contents
```suggestion
        #  annotations_dir will be None for all archive versions < 7.0
```

qiime2/core/annotate.py
Oddant1: If they use some file that can't be parsed into text as their annotation then if we attempt to read those contents later we will get an error, so we should probably make sure we can actually read the file as text before we use it as an annotation

qiime2/sdk/result.py
lizgehret: @ebolyen says iter...

qiime2/core/annotate.py
lizgehret: discussed offline, will remove for now as we don't have time to test this before releasing

qiime2/core/annotate.py
lizgehret: i've updated this to `self.id` because this feels less confusing to an end user vs having `uuid`, `root_result_uuid`, and `referenced_result_uuid` all living in the same place

qiime2/sdk/result.py
lizgehret: Evan and i worked on this offline and have implemented the latter suggestion

qiime2/sdk/result.py
Oddant1: @ebolyen has drank the LISP kool-aid :face_exhaling:

qiime2/core/annotate.py
lizgehret: ok i've added file size and content validation - so it can't exceed 10 MB and must be able to be read as utf-8

qiime2/core/archive/format/tests/test_util.py
lizgehret: i've updated this to match on the first three lines & have adjusted write_meta_yaml to construct an orderedDict, so it's easier to match on multi-line content

qiime2/core/archive/format/v7_0.py
lizgehret: updated such that deps are now an installable env

qiime2/core/archive/format/v7_0.py
lizgehret: i think i'd lean towards keeping it like this just so it's extra clear that a conda environment wasn't detected (even though i have no clue how we'd get here without a conda env lol)

lizgehret: that's also why i formatted it like this, so it would still be valid yaml 🤪

qiime2/core/util.py
lizgehret: i'll leave it for now since the python fallback should always work in the context of using QIIME 2 (like.... how would anything work without a valid python in a user's env)

qiime2/sdk/result.py
lizgehret: we worked on this offline & this has been fixed with an _annotation property on the result object

qiime2/sdk/result.py
lizgehret: updated the whole docstring as a comment, so just leaving this as-is for us to reference in the future

qiime2/core/archive/provenance_lib/archive_parser.py
lizgehret: Not sure I understand what you mean @ebolyen, can you clarify?

qiime2/core/archive/format/v7_0.py
lizgehret: Hey @ebolyen yep you're totally right, I can def fix this - but I'm wondering if we ever expect to have subdirs under `data/`? I can't think of an example of where we'd see this off the top of my head. Or is this something we want to ensure is forwards compatible if this comes up in the future?

qiime2/core/util.py
lizgehret: i'll add the default arg as md5 here - re: test failures, i'd like to chat about this for a few mins about this so i can add anything that's been previously missing

qiime2/sdk/result.py
lizgehret: i think i gaslit myself on this when i was testing, which is what led me to add this - but i just tested again on a fresh result and this isn't needed

qiime2/sdk/result.py
lizgehret: this seemed like a reasonable place to add this since the checksum logic happens separately from the actual writing of the annotation - would you prefer it lives under `_write` though? if that feels more sensible i can move it there.

qiime2/core/archive/provenance_lib/_checksum_validator.py
lizgehret: ok this has been fixed - something that i'd also like to discuss though is why there were no test failures associated with this change

qiime2/core/archive/archiver.py
lizgehret: fixed from above comment re: annotations being excluded from checksum validation

qiime2/sdk/result.py
lizgehret: that's fair; i'll modify `_annotations` accordingly along w/the endpoints that handle it

qiime2/sdk/result.py
lizgehret: nice, yeah you're right Anthony. wrote this one too late in the day to be thinking clearly lol

qiime2/core/archive/format/v7_0.py
lizgehret: evan and i discussed offline, this occurs in metagenomics Results & i will factor this functionality out into a helper to effectively test

qiime2/core/archive/provenance_lib/_checksum_validator.py
lizgehret: evan and i discussed this offline, this is expected bc none of the tests include the addition of annotations, which is where we'd expect to see the difference (before this change was made). i am going to add a test to assert an empty diff for this change.

qiime2/core/util.py
lizgehret: evan and i chatted offline about this - there was a very unusual chain of confusion that's been resolved w/an added comment on how this constructor works wrt provenance's constructor. closing the case on this.

qiime2/sdk/result.py
lizgehret: evan and i discussed offline and are going to leave this here until a future time dictates this being more sensible to move elsewhere... bc decision fatigue lol

qiime2/core/archive/provenance_lib/archive_parser.py
lizgehret: discussed offline, similar logic to the get_format_class method within the archiver was added into the archive_parser within provlib

lizgehret:
Okay @ebolyen @Oddant1 this is ready for a (hopefully) final review!

qiime2/core/archive/archiver.py
```diff
@@ -487,10 +490,15 @@ def validate_checksums(self):
...
+                 if (x[0] != self._fmt.CHECKSUM_FILE and
+                     pathlib.Path(x[0]).parts[0] != 'annotations')
+                 )
         with open(self.root_dir / self._fmt.CHECKSUM_FILE) as fh:
             exp = dict(from_checksum_format(line) for line in
-                       fh.readlines())
+                       fh.readlines()
+                       if (pathlib.Path(from_checksum_format(line)[0]).parts[0]
+                           != 'annotations')
+                       )
```
ebolyen:
```suggestion
```

We don't need this bit I think, since we won't write them into the checksum file in general.

qiime2/core/archive/provenance_lib/_checksum_validator.py
```diff
@@ -144,7 +144,8 @@ def diff_checksums(zf: ZipFile) -> ChecksumDiff:
     exp = {}
     for line in zf.open(checksum_fp):
         fp, checksum = from_checksum_format(str(line, 'utf-8'))
-        exp[fp] = checksum
+        if pathlib.Path(fp).parts[0] != 'annotations':
+            exp[fp] = checksum
```
ebolyen:
```suggestion
        exp[fp] = checksum
```
Same here I think, we can trust the checksum file to not contain things we don't care about.

ebolyen: ~~The section above however (for `obs = checksum...`) does need some updated logic. But I probably haven't gotten to the function definition for it yet.~~  Found it!

qiime2/core/archive/provenance_lib/archive_parser.py
```diff
@@ -616,9 +616,9 @@ def get_parser(cls, artifact: Union[str, pathlib.PosixPath]) -> Parser:
             An ArchiveParser object for the version of the artifact. One of
             ParserV[0-6].
         '''
-        if type(artifact) is pathlib.PosixPath:
+        if isinstance(artifact, pathlib.PosixPath):
```
ebolyen, line 619: yeah, that's probably a good idea

qiime2/core/util.py
```diff
@@ -214,10 +214,18 @@ def checksum_directory_zip(zf: zipfile.ZipFile, checksum_type: str) -> dict:
...
-            file_parts = list(fp.parts)
-            fp_w_o_root_uuid = pathlib.Path(*(file_parts[1:]))
-            sums[str(fp_w_o_root_uuid)] = checksum_zip(zf, file, checksum_type)
+        if fp.name == f'checksums.{checksum_type}':
+            continue
+
+        file_parts = list(fp.parts)
+        internal_path_parts = file_parts[1:]
+
+        if internal_path_parts and internal_path_parts[0] == 'annotations':
```
ebolyen, line 223: Here's the logic I was looking for, looks good!

qiime2/sdk/result.py
```diff
@@ -445,48 +429,31 @@ def remove_annotation(self, name):
...
 
-        # Guard against Annotation name not found on disk
-        if annotation_disk_dir is None:
+        if not os.path.exists(annotation_disk_dir):
             raise ValueError('Unable to locate on-disk directory '
                              f'for Annotation with name: "{name}"')
 
         shutil.rmtree(annotation_disk_dir)
-        self._annotations.remove(annotation_to_remove)
+        del annotations[name]
```
ebolyen, line 456: This all got way easier!

ebolyen (approved):
Looks great! I think we can assume that `annotation/...` won't be in the root checksum, so there's not any need to ignore that path when loading the expected checksums from an archive. (This would also avoid logic that could somehow miss or ignore what the checksum says, which seems like something we would prefer not to do especially when signatures get involved.)

qiime2/sdk/result.py
Oddant1: Look at you with you fancy __ methods. @ebolyen drank the LISP kool aid with the yield and the OOP kool aid with the __. All the kool aid

lizgehret: i'll take this as a compliment @Oddant1

ebolyen (approved):

ebolyen (approved):

PR #828, opened 2025-01-11
IMP: adds util for subprocess calls

### LINKED ISSUE #654: imp: adds `run_commands` utility function to the framework
It would be nice to add a utility function that DRYs up the functionality found in many of our plugins to run a command list (used for wrapper plugins). Example found [here](https://github.com/qiime2/q2-dada2/blob/master/q2_dada2/_denoise.py#L25) and [here](https://github.com/qiime2/q2-dada2/blob/3612f0a5d2ee255dfd2ccfc421dce14536e1e462/q2_dada2/_denoise.py#L220) in q2-dada2.

lizgehret:
closes #654

### DIFF
--- a/qiime2/plugin/util.py
+++ b/qiime2/plugin/util.py
@@ -7,6 +7,7 @@
 # ----------------------------------------------------------------------------
 
 import psutil
+import subprocess
 
 from qiime2.core.transform import ModelType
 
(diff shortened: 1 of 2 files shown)


lizgehret:
Blocking:
- [x] https://github.com/qiime2/q2-composition/pull/141
- [x] https://github.com/qiime2/q2-cutadapt/pull/66
- [x] https://github.com/qiime2/q2-dada2/pull/176

colinvwood (commented):
Looks good but I think the a better location is `qiime2/plugin/util.py`, that seems to be where we put utilities that may be useful to plugin developers. It looks like `qiime2/util.py` also has some of these, but those are probably also in the wrong spot.

lizgehret:
Got it, I've moved it into `qiime2/plugin/util.py` @colinvwood. I've also updated the imports for the connected plugins to reflect this change.

colinvwood (approved):

PR #830, opened 2025-01-16
MAINT: `pkg_resources` deprecation

lizgehret:
first pass at removing `pkg_resources` from the framework and replacing it with the `importlib.*` equivalent actions.

### DIFF
--- a/qiime2/core/archive/provenance.py
+++ b/qiime2/core/archive/provenance.py
@@ -10,7 +10,8 @@ import os
 import time
 import collections
 import collections.abc
-import pkg_resources
+import platform
+import importlib.metadata
 import uuid
 import copy
 import shutil
@@ -373,8 +374,14 @@ class ProvenanceCapture:
         return ForwardRef('environment:plugins:' + plugin.name)
 
     def capture_env(self):
-        return collections.OrderedDict(
-            (d.project_name, d.version) for d in pkg_resources.working_set)
+        env = collections.OrderedDict()
+
+        for pkg_list in importlib.metadata.packages_distributions().values():
+            for pkg in pkg_list:
+                version = importlib.metadata.version(pkg)
+                env[pkg] = version
+
+        return env
 
     def transformation_recorder(self, name):
         section = self.transformers[name] = []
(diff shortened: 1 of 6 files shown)


qiime2/core/archive/provenance.py
```diff
@@ -373,8 +374,14 @@ def reference_plugin(self, plugin):
         return ForwardRef('environment:plugins:' + plugin.name)
 
     def capture_env(self):
-        return collections.OrderedDict(
-            (d.project_name, d.version) for d in pkg_resources.working_set)
+        env = collections.OrderedDict()
+
+        for pkg_list in importlib.metadata.packages_distributions().values():
```
ebolyen: We learned this is incomplete, so I'm going to poke around a bit more for an equivalent to `working_set`

qiime2/core/archive/provenance.py
```diff
@@ -374,7 +375,9 @@ def reference_plugin(self, plugin):
 
     def capture_env(self):
         return collections.OrderedDict(
-            (d.project_name, d.version) for d in pkg_resources.working_set)
+            (d.metadata["Name"], d.metadata["Version"]) for d in
```
lizgehret, line 378: Okay @ebolyen I think this works - I compared artifacts from a dev env before/after these changes and the list is the same now.

qiime2/core/archive/provenance.py
```diff
@@ -453,15 +456,21 @@ def make_transformers_section(self):
...
         # use literal formatting.
         env['python'] = LiteralString('\n'.join(line.strip() for line in
                                       sys.version.split('\n')))
         env['framework'] = self.make_software_entry(
             qiime2.__version__, qiime2.__website__, self._framework_citations)
         env['plugins'] = self.plugins
-        env['python-packages'] = self.capture_env()
+
+        # sort pkgs alphabetically, ignoring upper/lower casing
+        unsorted_packages = self.capture_env()
```
lizgehret, line 469: @ebolyen this could probably also go under capture_env() - I don't have strong feelings. But this actually does a better job of sorting than what we were doing before (since it ignores casing).

PR #839, opened 2025-03-19
Added dictionary conversion to Results class and test script

RishiP2006:
Brief summary of the Pull Request, including any issues it may fix using the GitHub closing syntax:

This update adds a to_dictmethod to theResults class, allowing conversion to a dictionary. A test script has been included to validate the changes.

https://help.github.com/articles/closing-issues-using-keywords/

Also, include any co-authors or contributors using the GitHub coauthor tag:

https://help.github.com/articles/creating-a-commit-with-multiple-authors/

---

Include any questions for reviewers, screenshots, sample outputs, etc.


qiime2/sdk/results.py
```diff
@@ -44,14 +45,22 @@ def __new__(cls, fields, values):
...
-        object.__setattr__(self, '_fields', fields)
+        object.__setattr__(self, "_fields", fields)
 
         # Attach field names as instance attributes.
         for field, value in zip(fields, values):
             object.__setattr__(self, field, value)
 
         return self
 
+    def __iter__(self):
```
ebolyen, line 56: This breaks our API almost everywhere. The default `__iter__` of tuple permits unpacking of results:
```python
a, b = resultsObject
```

ebolyen (commented):
Hi @RishiP2006, unfortunately it looks like your editor has auto-formatted the entire repository, so we can't accept this PR as the commit obscures the intended changes. Also, there is not really a discussion about this yet [see our contributing guide](https://github.com/qiime2/.github/blob/main/CONTRIBUTING.md), if there was, we could have told you that there already is a method for this:
https://github.com/qiime2/qiime2/blob/136693096f9791b0d357d357563a345cace09ef9/qiime2/sdk/results.py#L111-L112

You'll notice it appears private, but we are actually matching the `namedtuple` API in that we don't want to collide with a field. So there's nothing wrong with using `_asdict` it is a supported API on our end.

Also, in your future endeavors, I would strongly recommend using feature branches when creating pull requests with git, you'll find your life gets a lot easier that way.

PR #840, opened 2025-03-26
Refactor locking

Oddant1:
Create a minimal locking version of GC and refactor locking in the cache in general.

@misialq ran into an issue [on the forum](https://forum.qiime2.org/t/notlockederror-with-assemble-megahit-and-parsl/31831/9) that seems to be as a result of his very large cache on an NFS taking more than 10 minutes to run garbage collection.

That prompted this PR which seeks to reduce the amount of locking required by garbage collection and the cache in general. We will leave around the fully locked version of GC for deeper cleaning, but the version that runs frequently during normal execution will work on a more "best effort" basis removing some of the things that need removed while ensuring nothing that shouldn't be removed is deleted and the lock is held as briefly as possible.

### DIFF
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -1256,7 +1256,7 @@ class Cache:
         Returns
         -------
         set[str]
-            All of the keys in the cache. Just the names now what they refer
+            All of the keys in the cache. Just the names not what they refer
             to.
         """
         return set(os.listdir(self.keys))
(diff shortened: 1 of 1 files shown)


lizgehret:
Hey @Oddant1, @ebolyen and I spent some time discussing and think this might need a different approach. The issue is result.load will recycle data and we can't work out a way for that to be done consistently if the same artifact is already marked for deletion and is loaded from a zipfile on two separate threads. tl;dr we'll set up a time to all meet and brainstorm 🧠

Oddant1:
@lizgehret Yeah, definitely need some more design work on this. I've also considered how to resolve this and how many other issues could conceivably result from running gc without locking when other things might expect the cache to be locked...

Oddant1:
closed in favor of #845

PR #843, opened 2025-05-07
MAINT: fix futuristic version error for cache (#843)

Oddant1:
I originally dropped the ball on making this futuristic cache error work out properly. We do have a work around which is to make version 2 and on "v2" as opposed to just 2. This will cause older versions of QIIME 2 that only support "1" to say a cache with "v2" and on will fail to validate as a cache. It will not raise a helpful error on versions of QIIME 2 prior to this PR, but it will at least recognize that it can't read the cache.


Oddant1:
@ebolyen this is the PR I was telling you about that may be a good candidate for you to review with someone else

qiime2/core/cache.py
```diff
@@ -567,11 +575,39 @@ def is_cache(cls, path):
 
         regex = \
             re.compile(
-                r"QIIME 2\ncache: \d+\nframework: 20\d\d\.")
+                r"QIIME 2\ncache: v?\d+\nframework: 20\d\d\.")
```
```suggestion
                r"QIIME 2\ncache: v?(\d+)\nframework: (20\d\d\.\d+)")
```
If there was a capture group we could skip the yaml parsing and fh schnaedling (Liz says this is a word).

ebolyen (changes requested):
I think switching to regex will make this simpler.

(I understand the irony of that sentence)

qiime2/core/cache.py
Oddant1: So. I am adding a regex (gross) with reasoning based on a fake word (gross gross)

lizgehret (approved):
Using capture groups makes this a lot nicer; this lgtm @Oddant1! Thanks!

PR #845, opened 2025-06-12
IMP: Do not set_permissions pre removal for performance

Oddant1:
Simply running GC unlocked as in #840 opens a number of potential holes, this PR will make some small adjustments to make GC more efficient, and it will make sure the lock stays around as long as it needs to in order to prevent the issue where with a large cache on a slow NFS we can take so long running GC that the lock times out.

### DIFF
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -262,6 +262,30 @@ def monitor_thread(cache_dir, is_done):
         time.sleep(60 * 60 * 6)
 
 
+def lock_thread(flufl_lock, lifetime, is_done):
+    """It is conceivable that something could need to hold the lock for longer
+    than the lifetime, so this thread will periodically refresh the duration
+    of the lock provided we are still holding it. We still want to set a
+    limited lifetime on the lock so that if whatever is holding the lock
+    terminates abnormally the next process that comes along and wants the lock
+    will break the expired lock
+
+    Parameters
+    ----------
+    flufl_lock: flufl.lock.Lock
+        The lock we are refreshing the duration on.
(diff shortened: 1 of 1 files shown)


qiime2/core/cache.py
```diff
@@ -1130,8 +1127,6 @@ def _copy_to_data(self, ref):
                     shutil.copytree(
                         ref._archiver.path, self.data, dirs_exist_ok=True)
 
-                set_permissions(destination, READ_ONLY_FILE, READ_ONLY_DIR)
```
Oddant1, line 1133: So it would seem as though this `set_permissions` and the one in `_rename_to_data` ought to be removed to make the GC work without setting write permissions; however, these operations don't seem to be working properly as is? When I actually look at the directories I ran the `set_permissions` on here, they have 0x755.

Oddant1: Oh man I am blind. We no-oped set_permissions for any permissions that aren't 0x777

qiime2/core/cache.py
```diff
@@ -263,6 +262,33 @@ def monitor_thread(cache_dir, is_done):
...
+    while not is_done.is_set():
+        try:
+            flufl_lock.refresh()
+        except flufl.lock._lockfile.NotLockedError:
+            break
+        time.sleep(lifetime.seconds * .9)
+
+
 # This is very important to our trademark
 tm = object
```
lizgehret, line 293: 👍🏼

lizgehret (approved):
lgtm (@ebolyen and I reviewed sitting right next to you)

PR #849, opened 2025-07-02
Revert "Revert "IMP: Do not set_permissions pre removal for performance""

lizgehret:
hit em with a revert revert.

@Oddant1 assigning this back to you so we can investigate post-release.


Oddant1:
At the very least, it's easy(ish) to stop setting the permissions. The whole thread deal may be a bigger problem. The only potential issue with not setting the permissions is that, in theory, someone could still have an old cache with write-only entries in it.

Oddant1:
The permission changes were made in #875. The thread is still an open question

Oddant1:
thread change superceded by #890

PR #850, opened 2025-07-14
MAINT: try adding xdist for parallelizing unit tests

lizgehret:
cc @andrewsanchez 🙃

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -41,7 +41,7 @@ test:
   - qiime2
   commands:
   - QIIMETEST= python -c "import qiime2.plugins.dummy_plugin"
-  - QIIMETEST= py.test -n auto --pyargs --doctest-modules qiime2
+  - QIIMETEST= py.test -n 2 --pyargs --doctest-modules qiime2
 about:
   home: https://qiime2.org
   license: BSD-3-Clause
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -36,11 +36,12 @@ test:
   - pytest
   - tornado
   - notebook <7
+  - pytest-xdist
   imports:
   - qiime2
   commands:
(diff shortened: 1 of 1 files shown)


conda-recipe/meta.yaml
```diff
@@ -36,11 +36,12 @@ test:
   - pytest
   - tornado
   - notebook <7
+  - pytest-xdist
   imports:
   - qiime2
   commands:
   - QIIMETEST= python -c "import qiime2.plugins.dummy_plugin"
-  - QIIMETEST= py.test --pyargs --doctest-modules qiime2
+  - QIIMETEST= py.test -n 2 --pyargs --doctest-modules qiime2
```
andrewsanchez, line 44: Is `-n 4` or `-n auto` OK? On public GitHub Actions Ubuntu-hosted runners, pytest-xdist `-n auto` will spawn 4 worker processes, since those runners have 4 CPU cores, AFAICT.

lizgehret: I was testing out n=2 to see how that impacted the test failures since there were fewer failures on ubuntu compared to osx (and ubuntu defaults to 2 workers while osx defaults to 4).

andrewsanchez: Oh, gotcha! The [`--dist` flag](https://pytest-xdist.readthedocs.io/en/stable/distribution.html) might help with those failures. For example,
>--dist loadscope: Tests are grouped by module for test functions and by class for test methods. Groups are distributed to available workers as whole units. This guarantees that all tests in a group run in the same process. This can be useful if you have expensive module-level or class-level fixtures. Grouping by class takes priority over grouping by module.

gregcaporaso:
Closing this for `qiime2` as it didn't really give much performance boost for the tests. It was integrated in some plugin repos with long running tests. Thanks for the suggestion @andrewsanchez!

PR #852, opened 2025-07-21
NEW: Visualizer.make_report

ebolyen:
Just a gut-based first-take on this. 

Pairs with https://github.com/qiime2/q2view/pull/53


gregcaporaso:
TODO: Add error message (or other fix) if a report is provided for provenance parsing.

ebolyen:
Provenance parsing works fine it turns out, so this will be compatible with older qiime 2's knowledgeable about the 7.0 series. :)

qiime2/sdk/result.py
```diff
@@ -652,6 +654,101 @@ def data_initializer(destination):
...
+                }
+                subfigure_root = os.path.join(subfigures_dir, viz_uuid)
+                if not os.path.exists(subfigure_root):
+                    shutil.copytree(viz._archiver.data_dir, subfigure_root)
+
+                # if not for reports of reports, this would have been the end
+                # of the data_initializer, however we don't want nested
+                # reports to have an arbitrarily long path as this will break
+                # things. Also we get de-duplication for free if we hoist
+                # sub-subfgures into just subfigures of the outermost report
```
```suggestion
                # sub-subfigures into just subfigures of the outermost report
```

qiime2/sdk/result.py
```diff
@@ -652,6 +654,101 @@ def data_initializer(destination):
...
+                if viz.format is report:
+                    # reports which were providfed as part of the collection
```
```suggestion
                    # reports which were provided as part of the collection
```

qiime2/core/util.py
```diff
@@ -470,3 +471,107 @@ def create_collection_name(*, name, key, idx, size):
...
+def replace_bytes_in_file(filepath, old_bytes, new_bytes, buffer_size=None):
+    """
+    Replaces all occurrences of old_bytes with new_bytes in the specified file.
+
+    Parameters
+    ----------
+    filepath : str
+        Path to the file to process.
+    old_bytes : bytes
+        The byte sequence to find and replace. Must be at least 2 bytes long.
```
Oddant1: Why does this have to be at least two bytes? And if this is a hard requirement, should we raise an error if it isn't met?

qiime2/sdk/result.py
```diff
@@ -652,6 +654,101 @@ def data_initializer(destination):
...
+                    # will have sub-figures moved and index.json re-written
+                    # to a new flattened path
+                    to_reindex[viz_uuid] = set()
+                    child_figs = os.path.join(subfigure_root, 'subfigures')
+                    with open(os.path.join(child_figs, 'index.json')) as fh:
+                        inner_index = json.load(fh)
+                        index[key]['children'] = inner_index
+                        to_reindex[viz_uuid] = \
```
Oddant1, line 714: I'd probably leave a comment explaining this a bit, in particular, probably explaining what I expect the path I'm splitting and grabbing the -2 element from to look like.

qiime2/core/archive/provenance.py
```diff
@@ -661,3 +662,19 @@ def fork(self, name, alias):
         forked.alias = alias
         forked.add_ancestor(alias)
         return forked
+
+
+class ReportProvenanceCapture(ProvenanceCapture):
+    def __init__(self):
+        super().__init__()
+        self.inputs = OrderedKeyValue()
```
Oddant1, line 670: What do we get from using this OrderedKeyValue class? Especially since dicts are fundamentally ordered now

qiime2/core/util.py
```diff
@@ -470,3 +471,107 @@ def create_collection_name(*, name, key, idx, size):
...
+
+        # Write any remaining overlap buffer
+        if overlap_buffer:
+            dst.write(pattern.sub(new_bytes, overlap_buffer))
+
+    # Replace the original file with the modified file
+    os.replace(tempfile, filepath)
+
+
+def flatten_children(dictionary, child_key='children'):
```
Oddant1, line 572: :flatbread: :steam_locomotive:

qiime2/core/archive/provenance.py
ebolyen: I think it was for a specific YAML representation, we connect a custom type (list of single key-value pairs)

qiime2/core/util.py
ebolyen: I think it had more to do with this being a silly function to use if you only had a string of 1 or less to replace. None of the regex business is needed anymore in that case.

Also it's a `qiime2.core.util` so I didn't worry about the sharper surface of the API.

Let me know if you would want me to change this.

ebolyen: yeah it can handle 1 byte also, but this function is just an unreasonably expensive way to do it

ebolyen:
@Oddant1 I think I covered everything, thanks for the review!

qiime2/core/util.py
ebolyen: https://neal.fun/absurd-trolley-problems/

qiime2/core/util.py
Oddant1: That's about what I figured. I did take some time thinking about what mechanism might actually make it not work if it was 1 byte, and I couldn't come up with one because there isn't one. There should probably be a disclaimer saying, "This will work with 1 byte, it's just a silly way of doing it with 1 byte" if only to save anyone in the future from sitting down and trying to come up with a technical reason it would break with 1 byte.

PR #856, opened 2025-07-29
MAINT: Remove multiple aliases from process pools

Oddant1:
We should no longer be required to create many aliases of the same artifact in process pools now that we no longer enter Contexts as Scopes.

Supercedes #855 with a slightly reworked and better approach.

### DIFF
--- a/qiime2/core/archive/archiver.py
+++ b/qiime2/core/archive/archiver.py
@@ -291,8 +291,8 @@ class Archiver:
         from qiime2.core.cache import get_cache
 
         cache = get_cache()
-        path = cache.process_pool._allocate(uuid)
-        return path, cache
+        path, created = cache.process_pool._allocate(uuid)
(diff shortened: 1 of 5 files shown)


ebolyen:
Is the approach here to basically not worry about destruction of anything until process-end?

Oddant1:
Yes, but that is more or less the approach that has been taken ever since we stopped using Contexts as Scopes,

This PR is primarily trying to rework things such that we don't have unique aliases referring to a given Artifact anymore. Our one alias ought to hang around for the process lifetime now I believe.

Oddant1:
Blocking qiime2/q2cli#374

Oddant1:
It turns out that, unsurprisingly, there isn't a super-easy win to be had here

ebolyen:
Closing for now, but the implementation may inform future work.

PR #858, opened 2025-07-31
MAINT: No error on failing to load cached collection

Oddant1:
Currently, if we attempt to load a cached collection for pipeline resumption and it is malformed, we get an assertion error. This should instead just rerun the action to create the collection.


lizgehret (approved):
for posterity, @Oddant1 and I discussed this offline to address a couple of questions i had on pipeline resumption and the test methods.

with that said, this all lgtm!

PR #862, opened 2025-08-19
Should be self.path

Oddant1:
This should be checking that `self.path == temp_cache_path`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ lint:
 	flake8
 
 test: all
-	QIIMETEST= pytest -k test_load_complex_config
+	QIIMETEST= pytest --doctest-modules
 
 # for parallel, pip install pytest-xdist
 mystery-stew: all
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ lint:
 	flake8
 
 test: all
-	QIIMETEST= pytest --doctest-modules
+	QIIMETEST= pytest -k test_load_complex_config
 
 # for parallel, pip install pytest-xdist
 mystery-stew: all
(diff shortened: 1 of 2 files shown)


Oddant1:
@colinvwood was able to reproduce the test failure here ~50% of the time on his M3 Mac. It happens seemingly 100% of the time on the runners. He said when it passes, it does so instantly; when it fails, it hangs for a while first. It seems like something to do with threading on OSX isn't getting along with us.

#845 saw similar issues when we introduced a thread that works fine on Linux. In that case, it didn't fail CI on the PR, but it did consistently time out when we were building the package.

Oddant1:
Now failing to reproduce the failure on @colinvwood's M3 Mac even though it was happening before. Also failed to repro it on my wife's M1. The only thing I can think of is that the GitHub runners are running osx 13 while all Macs I've been able to run it on are running osx 15, but Colin said when he was able to produce the failure it was on 15

Oddant1:
ACTUALLY it looks like the test was failing on macos-latest and CI was then changed to macos-13, and it is now passing.

The test that it was failing on is the only test in the framework that instantiates and loads a HighThroughputExeutor and it was timing out on the interchange for the executor starting. Could be a race condition of some sort, but I'm not sure what would be causing that under the hood. The thread we are starting doesn't do anything particularly interesting.

It is possible the OS has some sort of thread/process limit that we are hitting? Seems unlikely, but who knows. The mac runners have always been a bit borked. My only concern is that Colin was able to reproduce the issue at one point.

Oddant1:
We're gonna merge this because it really seems like the test failures are down to the CI workers being bad.

PR #867, opened 2025-09-06
closes #833: remove crypt library

### LINKED ISSUE #833: There is no need to use crypt module in tests
**Bug Description**
In the Debian packaged qiime package we are using a [patch](https://salsa.debian.org/med-team/qiime/-/blob/master/debian/patches/replace-crypt-module-by-hashlib.patch?ref_type=heads) which avoids the usage of the crypt module in tests.  You might like to review this patch whether it should be included into your code.

**Computation Environment**
- OS: Debian
- QIIME 2 Release 2024.5.0 and 2024.10.1

Kind regards
     Andreas.

colinbrislawn:
Replaced crypt module with hashlib for password hashing as suggested by @tillea

Also, include any co-authors or contributors using the GitHub coauthor tag:

Co-authored-by: @tillea

### DIFF
--- a/qiime2/core/tests/test_cache.py
+++ b/qiime2/core/tests/test_cache.py
@@ -9,6 +9,7 @@
 import os
 import gc
 import pwd
+import hashlib
 import shutil
 import string
 import atexit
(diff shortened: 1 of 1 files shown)


Oddant1:
I genuinely can't remember what the purpose of that 22 was... But it doesn't matter the specifics of this password don't matter at all. Looks good to me.

PR #868, opened 2025-09-09
MAINT: Add headers to `urllib.request`

Oddant1:
We started getting a 403 Forbidden when using urllib.request.urlopen to get docs.qiime2.org. Adding some headers to the request clears that up


lizgehret:
I've also tested this locally and it works for me. Once CI passes I'll get this merged.

lizgehret (approved):

PR #872, opened 2025-09-24
NEW: Archive Format v7.1

### LINKED ISSUE #847: Archive v7.1
New features will include:

## Annotation[Signature]

Under `provenance/annotations/<annotation_uuid>`:

**`metadata.yaml`**
- contains usual annotation content (ie name, type, etc)
- NEW signature block with the algorithm, fingerprint, relpaths, sizes, digests, optional signer name/email ~~& whether the public key is embedded (tbd if this is required or optional)~~

**`manifest.txt`**
- canonical signed manifest that is the result of walking the Result directory (excluding /annotations/*) sorting all paths & emitting the sha512 and relpath for each file (along with some sort of header line like `q2-signature-manifest-v7.1`)

**`signature.txt`**
- likely using Ed25519 algorithm
- signature is computed either with a provided private key ~~or via a newly generated key pair~~ along with the manifest file

**"Helper" digests:**
- `manifest_digest = sha512(manifest.txt)`
- `key_id = sha512(public_key)` ie., _fingerprint_

### Signature verification after the fact:

- Rebuild the canonical manifest from the current files that exist under the given Result & compare its sha512 hash to what `metadata.yaml` says the manifest digest was
- Obtain the public key (via an externally provided key ~~TBD on if embedding public key is required~~) & check that its fingerprint matches the fingerprint that's present in `metadata.yaml`

### Open questions to be addressed:

1. Do we want to have some sort of 'allowlist' that we append to wrt fingerprint for verification? Not sure if this is helpful but I'm thinking about this wrt either common collaborators or classifiers who will have a common fingerprint.
  a. Keyring
2. Any objections to using Ed25519 for the signing algorithm? This seems like the best/most modern option that's widely used.
~~3. Should we support storing a newly generated private key either via a user specified location on their machine, or in a user's cache?~~
a. Not going to support generating a new key pair bc of man in the middle attacks that we aren't adequately equipped to deal with - will push people to generate their own key pairs using gpg directly
~~4. Should embedding the public key in the annotation be optional or required?~~

lizgehret:


qiime2/core/annotate.py
```diff
@@ -70,10 +78,12 @@ def load(cls, filepath):
             corresponding annotation directory.
 
         """
+        # TODO: there needs to be a better way to deal with these shared attrs
```
Oddant1: If you really want you could spin it off into a helper function like `add_shared_attrs` or something.

qiime2/sdk/result.py
```diff
@@ -455,6 +502,169 @@ def remove_annotation(self, name):
...
+            - sha512sum for root level checksums file matches checksum_digest
+            - gpg detached signature verification
+            - sha512sum checks for each file in signature-level checksums file
+        """
+        signature = None
+
+        for annotation in self.iter_annotations('Signature'):
+            if getattr(annotation, 'name') == signature_name:
+                signature = annotation
+                break
```
```suggestion
        signature = self.get_annotation(signature_name)
```

qiime2/sdk/result.py
```diff
@@ -327,6 +332,38 @@ def _validate_annotation_support(self):
...
+                if not line:
+                    continue
+                match = checksum_line_regex.match(line)
+                if match:
+                    yield (match.group(1), match.group(2))
+
+    # base report for signature verification
+    # keys will be updated as different checks pass, else will include the
+    # failure message in the details key pair
+    def _empty_report(self, signature=None):
```
Oddant1: If we change `verify` to use `get_annotation` it will raise an error if the requested signature doesn't exist, then I don't think this function needs to handle a `None` signature.

qiime2/core/annotate.py
```diff
@@ -94,6 +104,30 @@ def load(cls, filepath):
...
+                annotation.fingerprint = meta_yaml.get('fingerprint')
+
+                # Validate that `signature.gpg` exists
+                sig_fp = os.path.join(filepath, 'signature.gpg')
+                if not os.path.exists(sig_fp):
+                    raise ValueError(
+                        'Unable to load malformed Signature with name: '
+                        f'"{annotation.name}" due to missing '
+                        '`signature.gpg` file.'
+                    )
```
Oddant1: Should we similarly make sure that the checksum file exists?

qiime2/sdk/result.py
```diff
@@ -455,6 +502,169 @@ def remove_annotation(self, name):
...
+            if getattr(annotation, 'name') == signature_name:
+                signature = annotation
+                break
+
+        report = self._empty_report(signature)
+
+        if signature is None:
+            report['details'] = \
+                f'No Signature with name "{signature_name}" found.'
+            return report
```
Oddant1: Is there a reason we want to do this instead of raising an error if the signature doesn't exist?

qiime2/sdk/result.py
```diff
@@ -455,6 +502,169 @@ def remove_annotation(self, name):
         shutil.rmtree(annotation_disk_dir)
         del annotations[name]
 
+    def verify(self, signature_name):
```
Oddant1, line 511: I think most of the things this function checks for will produce errors if we try to load a signature that doesn't have them. Even with the current `iter_annotations` approach, the `Signature` is still being instantiated by `Annotation.load` which is going to explode if many of the conditions this is checking are violated before we even get here.

qiime2/sdk/result.py
```diff
@@ -409,14 +446,24 @@ def get_annotation(self, name):
...
+            return
+
+        if filter_by_type not in ANNOTATION_TYPE_LIST:
+            raise ValueError(f'Unknown annotation type: "{filter_by_type}". '
+                             'Supported annotation types are: '
+                             f'{ANNOTATION_TYPE_LIST}')
+
+        for annotation in self._annotations.values():
+            if getattr(annotation, 'annotation_type') == filter_by_type:
+                yield annotation
```
```suggestion

        if filter_by_type is None:
            yield from self._annotations.values()
        elif filter_by_type not in ANNOTATION_TYPE_LIST:
            raise ValueError(f'Unknown annotation type: "{filter_by_type}". '
                             'Supported annotation types are: '
                             f'{ANNOTATION_TYPE_LIST}')
        else:
            for annotation in self._annotations.values():
                if getattr(annotation, 'annotation_type') == filter_by_type:
                    yield annotation
```

I think it's easier to read like this because it's just an if/elif/else without the sneaky return and whatnot, but this is an opinion thing.

qiime2/core/annotate.py
```diff
@@ -351,3 +395,319 @@ def _write(self, annotations_dir, root_result_uuid,
...
+        fingerprint = key_info['fingerprint']
+        signer_name = key_info['chosen_uid']['name']
+        signer_email = key_info['chosen_uid']['email']
+
+        annotation_uuid_dirname = \
+            self._write_meta_yaml(annotations_dir, root_result_uuid,
+                                  referenced_result_uuid, algorithm,
+                                  checksum_digest, signer_name, signer_email,
+                                  fingerprint)
+
```
Oddant1: This will potentially leave us with a hanging broken annotation if we create this then fail to actually create the key. If we get an error at any step of this process, we want to make sure anything we wrote is deleted not hanging around malformed.

qiime2/core/annotate.py
```diff
@@ -351,3 +395,319 @@ def _write(self, annotations_dir, root_result_uuid,
...
+    for _ in range(6):
+        provenance = p / 'provenance'
+        meta_yaml = p / 'metadata.yaml'
+        if provenance.is_dir() and meta_yaml.is_file():
+            return p
+        elif p.parent == p:
+            break
+        p = p.parent
+    raise ValueError(f'Could not locate Result root starting from {start}.')
```
Oddant1: What you wrote here is something that would probably go in util, but we don't need it in most cases because USUALLY if we need this information, we can access `Result._archiver.path` which maddeningly enough you're one step too far removed to do here. For this specific use case though, I would probably do something like the following.

```suggestion
def _find_root_fp(annotations_dir, root_result_uuid):
    split_fp = annotations_dir.split(os.sep)
    root_result_uuid_index = split_fp.index(root_result_uuid)
    root_fp = os.sep.join(split_fp[0: root_result_uuid_index + 1])
    return root_path
```

We know the root path we are looking for is above the `annotations_dir` we have, and we know it terminates with the `root_result_uuid`, so we can leverage this to split the path we have apart and recombine it only up to the point we need. The way you did it would work even if we didn't have the root_uuid, but in this case we do.

We also theoretically in this case do know EXACTLY where we are relative to the root of the artifact we are annotating, and so we could just hop up one level if we want to assume the structure won't change.

qiime2/core/annotate.py
```diff
@@ -351,3 +395,319 @@ def _write(self, annotations_dir, root_result_uuid,
...
+            The uuid of the Result object where an Annotation is being added.
+
+        referenced_result_uuid
+            The uuid of the Result object that an Annotation is referring to.
+            Note that in 7.0 & 7.1, `root_result_uuid` and
+            `referenced_result_uuid` are the same (i.e. Annotations can only
+            refer to the same Result they are being attached to) but separate
+            root and referenced uuids will be supported in future versions.
+        """
+        root_fp = _find_root_fp(annotations_dir)
```
Oddant1:
```suggestion
        root_fp = _find_root_fp(annotations_dir, root_result_uuid)
```

Only if `_find_root_fp` is changed

qiime2/sdk/result.py
lizgehret: omg no duh. thanks. i was def staring at this for too long

qiime2/sdk/result.py
lizgehret: no this is totally fair- i don't have strong opinions on it, this was just the way i ended up thinking through this section; this is definitely more readable.

qiime2/core/annotate.py
lizgehret: this makes sense! and yeah, it was maddeningly just out of reach.

qiime2/sdk/result.py
lizgehret: yeah that's a good point - the rest of it will just be a waste of time in that case.

qiime2/core/annotate.py
lizgehret: ok yep i'm just going to add this to util.py and then bump all of the other helpers over there.

qiime2/core/annotate.py
lizgehret: yeah probably best to do that here and just always raise if that's not found. that'll simplify the report as well.

qiime2/core/annotate.py
```diff
@@ -351,3 +381,201 @@ def _write(self, annotations_dir, root_result_uuid,
...
+        # Ensure at least one of signer uid (name/email address) is provided
+        if not signer_uid and not fingerprint:
+            raise ValueError(
+                'No inputs provided to either `signer_uid` or `fingerprint`. '
+                'Please provide either signer ID (name and email) or '
+                'fingerprint for key pair identification.'
+            )
+
+        # Construct Annotation class
+        self.signer_uid = signer_uid
```
Oddant1: @lizgehret and I talked offline about this, we're gonna just require the fingerprint to uniquely identify a specific key. No signer_uid

qiime2/core/annotate.py
```diff
@@ -351,3 +381,201 @@ def _write(self, annotations_dir, root_result_uuid,
...
+    See Also
+    --------
+    Annotation
+    """
+    annotation_type = 'Signature'
+
+    # NOTE: in future versions, name will become optional & the default value
+    # will be the annotation's UUID (if name isn't provided by the user)
+    def __init__(self, name, *, signer_uid=None, fingerprint=None):
+        self.validate_name(name)
```
Oddant1: This should probably go at the top of `Annotation.__init__`, then it doesn't need to be at the top of the subclass.__init__ methods.

qiime2/core/annotate.py
```diff
@@ -351,3 +381,201 @@ def _write(self, annotations_dir, root_result_uuid,
...
+                'No signer identity available. `signer_uid` '
+                '(e.g. "Name <email>") or `fingerprint` must be available '
+                'within gpg when constructing Signature.'
+            )
+
+        key_info = gpg_find_key(keypair_id)
+        algorithm = format_algorithm(key_info)
+        fingerprint = key_info['fingerprint']
+        signer_name = key_info['chosen_uid']['name']
+        signer_email = key_info['chosen_uid']['email']
```
Oddant1: This `gpg_find_key` can be done, and the attrs algorithm, signer_name, and signer_email can all be retrieved and set in `Signature.__init__`. I think it would be clearer to do it there because these attrs feel like a fundamental part of the `Signature` itself. Obviously, the checksums and whatever don't come into play until the `Signature` is written to a `Result`.

qiime2/core/annotate.py
```diff
@@ -94,12 +97,29 @@ def load(cls, filepath):
...
-                annotation.annotation_type = meta_yaml['type']
-                annotation.created_at = meta_yaml['created_at']
+            # SIGNATURE
+            elif annotation_type == 'Signature':
+                # signature-specific attrs
+                annotation.algorithm = meta_yaml['algorithm']
+                annotation.checksum_digest = meta_yaml['checksum_digest']
+                annotation.signer_name = meta_yaml['signer_name']
+                annotation.signer_email = meta_yaml['signer_email']
+                annotation.fingerprint = meta_yaml.get('fingerprint')
```
Oddant1, line 107: We're gonna make this mandatory right?

qiime2/sdk/result.py
```diff
@@ -385,6 +422,11 @@ def add_annotation(self, annotation):
                 fh.write(util.to_checksum_format(*item))
                 fh.write('\n')
 
+        # this ensures additional attrs on Signature (signer name/email)
+        # are present when running Result.verify
+        loaded_annotation = Annotation.load(str(annotation_dir))
+        self._annotations[loaded_annotation.name] = loaded_annotation
```
Oddant1, line 406: We can get signer name and email and whatnot in `__init__` without doing this... HOWEVER, this needs to happen regardless to get the checksums and all that which are written above in `annotation._write`... As silly as this bit of code seemed to me at first, I think it has kind of grown on me. The `Result` now has an `Annotation` object that definitively belongs to it.

It's not too different from what we do all over the `Cache` where after we save a `Result` to the `Cache`, we return a version of that `Result` loaded from the `Cache` and backed by the `Cache` path.

https://github.com/qiime2/qiime2/blob/deba9d1b866ff6ea57056f5df76878cfef67721b/qiime2/core/cache.py#L906

qiime2/sdk/result.py
```diff
@@ -457,6 +507,142 @@ def remove_annotation(self, name):
...
+
+        report['ok'] = (
+            report['fingerprint_ok']
+            and report['checksum_digest_ok']
+            and report['gpg_signature_ok']
+            and report['file_checksums_ok']
+        )
+        if not report['ok'] and not report['details']:
+            report['details'] = 'One or more verification checks failed.'
+        return report
```
Oddant1: Raise an error here if the report is not ok because people will probably expect that this not erroring out means they're fine. The error can be the report dict itself, formatted out in a string, then they still get relevant info.

qiime2/core/annotate.py
gregcaporaso: We're not planning to make signing mandatory, if that's what you're asking @Oddant1. I think that would be a bridge too far for a lot of users in terms of the complexity of using QIIME 2. This is intended to be optional, and something that users doing sensitive things like training and distributing classifiers would be expected to do.

gregcaporaso:
@lizgehret, I tested that I'm able to sign an artifact. No issues! What else can I do to help with this one?

qiime2/core/annotate.py
```diff
@@ -394,15 +394,9 @@ class Signature(Annotation):
...
         Annotation name (validated like other Annotations).
 
-    signer_name : str, optional
-        Name associated with the key pair in GnuPG
-        that will be used for signing.
-        At least one of name/email must be provided for Signature creation.
-
-    signer_email : str, optional
-        Email associated with the key pair in GnuPG
+    fingerprint : str, optional
```
Oddant1: This isn't optional anymore is it?

qiime2/core/util.py
```diff
@@ -573,45 +572,51 @@ def gpg_find_key(key_selector):
...
+        # the fingerprint will be here
         if tag == 'pub':
             in_primary = True
-            length = parts[2] or '0'
-            algorithm_num = parts[3] or ''
-            curve = parts[15] if len(parts) >= 16 and parts[15] else None
-            key_info['length'] = int(length) if length.isdigit() else 0
+            length = parts[2] if len(parts) > 2 else '0'
+            algorithm_num = parts[3] if len(parts) > 3 else ''
+            curve = parts[15] if len(parts) > 15 and parts[15] else None
```
Oddant1, line 598: why the `and parts[15]` on this line when the others don't check the truthiness of the value at their index?

qiime2/core/annotate.py
lizgehret: whoops, thanks. it is not

qiime2/core/util.py
lizgehret: this is fair, tl;dr is it's bc the curve is optional in gpg - but let me add some comments in here so it's more clear, since future us (and everyone else) will certainly be confused about this down the road

qiime2/sdk/result.py
```diff
@@ -480,6 +530,144 @@ def merge_annotations(self, other):
...
+        sig_checksums_fp = annotation_dir / 'checksums.sha512'
+        signature_fp = annotation_dir / 'signature.gpg'
+
+        try:
+            keypair_selector = (getattr(signature, 'fingerprint') or
+                                getattr(signature, 'signer_uid')
+                                )
+            if not keypair_selector:
+                report['details'] = 'Signature is missing fingerprint/UID.'
+                return report
```
Oddant1: This circumvents the exception we're trying to raise on failure

qiime2/sdk/result.py
```diff
@@ -480,6 +530,144 @@ def merge_annotations(self, other):
...
+                report['fingerprint_ok'] = (
+                    normalize_fingerprint(found_keypair['fingerprint']) ==
+                    normalize_fingerprint(signature.fingerprint)
+                )
+            else:
+                report['fingerprint_ok'] = True
+        except Exception as e:
+            report['details'] = ('Signer key not found in local GPG keyring: '
+                                 f'{e}')
+            return report
```
Oddant1: Also circumvents the exception we're trying to raise on failure

qiime2/sdk/result.py
```diff
@@ -480,6 +530,144 @@ def merge_annotations(self, other):
...
+                                )
+            if not keypair_selector:
+                report['details'] = 'Signature is missing fingerprint/UID.'
+                return report
+            found_keypair = gpg_find_key(keypair_selector)
+            if getattr(signature, 'fingerprint'):
+                report['fingerprint_ok'] = (
+                    normalize_fingerprint(found_keypair['fingerprint']) ==
+                    normalize_fingerprint(signature.fingerprint)
+                )
```
Oddant1: `signer_uid` doesn't exist anymore

```suggestion
            fingerprint = getattr(signature, 'fingerprint') 
            if not fingerprint:
                report['details'] = 'Signature is missing fingerprint.'
                return report # Some other form of control flow that doesn't dodge the error
                
             found_keypair = gpg_find_key(fingerprint)               
             report['fingerprint_ok'] = (
                    normalize_fingerprint(found_keypair['fingerprint']) ==
                    normalize_fingerprint(fingerprint)
                )
```

qiime2/core/annotate.py
```diff
@@ -352,3 +382,183 @@ def _write(self, annotations_dir, root_result_uuid,
...
+        checksum_digest = sha512_file_hex(checksums_fp)
+
+        env = os.environ.copy()
+        # Apparently this is helpful on Unix for GPG to find
+        # the correct terminal
+        try:
+            if sys.stdin and sys.stdin.isatty():
+                env.setdefault('GPG_TTY', os.ttyname(sys.stdin.fileno()))
+        except Exception:
+            pass
```
Oddant1: Maybe we throw this off into a helper since it's used in two places and it's such a weird looking thing. Putting it in a helper with a good name and this comment in it would make it look less strange

qiime2/core/annotate.py
```diff
@@ -111,6 +131,7 @@ def annotation_type(cls):
         raise NotImplementedError
 
     def __init__(self, name):
+        self.validate_name(name)
```
Oddant1: This should be below the docstring

Oddant1:
Test failure is because `self.validate_name(name)` was moved to the top of `Annotation.__init__` and `super().__init__(name)` is at the bottom of `Note.__init__`, so the name validation is happening after a buncha other stuff is validated and since only a name is passed into that `Note`, a buncha other stuff fails

lizgehret:
Tests to add:

- [x] passing roundtrip (sign then validate)
- [x] failed signature creation due to invalid fingerprint
- [x] failed verification because of missing/invalid public key

qiime2/sdk/tests/test_result.py
```diff
@@ -646,5 +649,111 @@ def test_invalid_key_added(self):
...
+        "uids": [{"raw": "Test User <test@example.com>",
+                  "name": "Test User",
+                  "email": "test@example.com"}],
+        "chosen_uid": {"raw": "Test User <test@example.com>",
+                       "name": "Test User",
+                       "email": "test@example.com"}
+    }
+
+    # default behavior will be to succeed
+    key_lookup_behavior = {'mode': 'ok'}
```
Oddant1, line 669: I think my only question is, is there an implementation reason this is a dict, not just a string? If not, I'd probably just make it a string. Otherwise, this looks good to me. I know the tests feel weak, but we have successfully round-tripped it for real using GPG, and I think we are validating well enough that if GPG works, we work, at least for now.

PR #873, opened 2025-09-30
NEW: Add functionality that can capture values set inside of an action in provenance

### LINKED ISSUE #823: develop approach for storing random seeds in provenance
A known limitation of data provenance in QIIME 2 is that random seeds aren't stored, unless a seed value is generated externally and passed in as a parameter (which the corresponding action would need to support). 

[Here's an example](https://github.com/caporaso-lab/q2-boots/commit/9195f9cfac1694196e1dfcc68934301d9238b350) of our current best way of handling this. Note that b/c the seed is generated internally, the value stored in provenance will be `None` unless the user provides another value. 

We should address this to support storage of seeds when the random value is generated internally. 

This came up here: https://github.com/qiime2/q2-feature-table/pull/317

Oddant1:
Close #823


qiime2/core/testing/method.py
```diff
@@ -183,5 +187,13 @@ def varied_method(ints1: int, ints2: list, int1: int = None,
     return ints1, ints2, int1
 
 
+def random_seed_method(random_seed: qtype.CaptureHolder = None) -> int:
+    if random_seed.value is None:
+        random_int = random.randrange(sys.maxsize)
+        random_seed.set_value(random_int)
+
+    return random_seed.value
```
Oddant1, line 258: Should we throw an error if a value was never set on the CaptureHolder?

Oddant1:
If a value was passed into a capture parameter, do we want to treat that as the capture being set?

I think doing that would be nice, but the value they explicitly passed in may be a sentinel value for capturing. The default value for a Capture parameter may not be a sentinel, or it may not be the only sentinel.

qiime2/core/testing/plugin.py
```diff
@@ -1110,6 +1111,32 @@ def factory():
                 ' collections'
 )
 
+dummy_plugin.methods.register_function(
+    function=random_seed_method,
+    inputs={},
+    parameters={
+       'random_seed': Capture[Int]
```
ebolyen: Do we need a semantic type for this? I don't think it changes how any interface would interact with the action.

qiime2/core/type/primitive.py
```diff
@@ -621,6 +621,45 @@ def encode(self, value):
...
+        pass
+
+
+class CaptureHolder:
+    def __init__(self, name, value, type, provenance):
+        self.set = False
+        self.name = name
+        self.value = value
+        self.type = type
+        self.provenance = provenance
```
ebolyen: These should probably be private for the most part otherwise someone is going to be clever and load/modify their own provenance.

qiime2/core/type/signature.py
```diff
@@ -401,6 +401,11 @@ def transform_and_add_callable_args_to_prov(self, provenance,
                     self._transform_and_add_input_to_prov(
                         provenance, name, spec, arg)
             else:
+                if spec.qiime_type.name == 'Capture':
```
ebolyen: I think we might look at the view type (for `CaptureHolder`) instead if there's no semantic type needed.

qiime2/core/testing/plugin.py
Oddant1: @ebolyen We don't need a semantic type no, but I do think this is the most pleasing way of implementing it from a user standpoint. I like the way "Capture[X]" looks and feels when registering it.

ebolyen: I'm not saying no to it necessarily, but we will need to change every interface to ignore/unwrap that outer type. Is that worth the effort?

Oddant1: No we probably don't :cry:

qiime2/core/testing/method.py
ebolyen: I like that idea and it's easier to relax that rule if we had to.

I think so long as `param.set_value(None)` is legal than there's no real downside.

ebolyen (commented):
Looks pretty good, but still need to pull it down and play with it.

One thing we'll need to do is add an import alias to `qiime2.plugin`. Also I think the actual implementation may want to be outside of `core/type/primitive.py`, but I don't actually have a nice place for it to live. Maybe a new `core/arguments.py` module?

qiime2/core/testing/method.py
Oddant1: @ebolyen How do we feel about rules for sentinel values? Are we ok with making the assumption there will only be one sentinel value? Do we feel ok with making the assumption that sentinel value will be the default value for the param? Makes a difference to how we determine if a value was set on the capture

Oddant1:
> but I don't actually have a nice place for it to live

@ebolyen Maybe it just goes in signature.py?

qiime2/core/testing/method.py
ebolyen: I see some of the issues here.

I like how simple the above code looks. Adding a rule about `set_value` would either mean you make defaults special to it, or have to do somewhat ugly things.

```python
    if random_seed.value is None:
        random_int = random.randrange(sys.maxsize)
    else:
        random_int = random_seed.value

    random_seed.set_value(random_int)
```

This just doesn't flow very nicely. So if we want to avoid setting what has already been set, then we need the default value to be treated as un-set, which is pretty much how we handle `None`, but if it was a value like `3` it gets weird. Although arguably, if `3` was the default, then there's nothing much to capture in the first place.


Given all of that, I think maybe the best approach is the default must be `None` and we treat that as un-set. Then to determine if `param.set_value()` has been called, we use a simple flag instead of the value itself. This permits `None` to be the value, but it requires the plugin to decide that rather than happening on accident and let's you skip `param.set_value` if it is anything other than `None`.

So then our logic would look like:

```python
def _is_set(self):
    return self._set_value_called or self.value is not None
```

ebolyen:
> Maybe it just goes in signature.py?

Yeah that sounds fine to me.
image
> `qiime2/core/type/signature.py` (1964 - Full Color)

qiime2/core/testing/method.py
Oddant1: Ok I did a buncha refactoring to follow these rules

qiime2/core/type/signature.py
```diff
@@ -779,3 +790,34 @@ def _make_hashable(self, collection):
...
+
+    @property
+    def value(self):
+        return self._value
+
+    def set_value(self, value):
+        if self._set:
+            raise ValueError(f'Value already set to {self._value}')
+
+        if value in self._type:
```
ebolyen: `None` is a special case here, it isn't in any primitive type, so we shuffle it along on a special track in signature. It is however a valid value here because the parameter is optional by definition (and always `None` as default).

```suggestion
        if value is None or value in self._type:
```

qiime2/sdk/action.py
```diff
@@ -157,6 +159,27 @@ def __init(self, callable, signature, plugin_id, name, description,
...
+
+        signature : Dict[string: ParamSpec]
+            The signature of this action
+
+        Raises : ValueError
+            If the default value of a CaptureHolder param is found to not be
+            the required default
+        """
+        for name, spec in signature.parameters.items():
+            if spec.view_type == qtype.CaptureHolder and spec.default != \
```
```suggestion
            if spec.view_type is qtype.CaptureHolder and spec.default != \
```

ebolyen (changes requested):
LGTM, just two tiny comments. Feel free to merge when ready.

PR #874, opened 2025-10-03
MAINT: Merge annotations in the cache

### LINKED ISSUE #869: Two of the same artifact with different annotations will collide in the cache
The cache is operating under the assumption that a uuid is enough to uniquely identify an artifact. This doesn't entirely hold true with annotations.

If you have an artifact with `uuidA` and `annotationA` in your cache then add an artifact with `uuidA` and `annotationB` the cache will have no way to distinguish between those two artifacts.

I propose we make the filepaths to the artifacts in the cache data-dir `uuid-<hash of annotation names>` to resolve this. 

EDIT: We may now just go out of our way to ensure that if `uuidA` is in the cache with `annotationA` and you add `uuidA` with `annotationB` to the cache you end up with `uuidA` in the cache with both `annotationA` and `annotationB`. We will need to document that the annotations are being combined and most likely also warn the user.

Oddant1:
Merge annotations in cache if an artifact is added to a cache twice with two different sets of annotations.

Prior behavior would be to just ignore the second addition of the artifact.

Closes #869

### DIFF
--- a/qiime2/core/annotate.py
+++ b/qiime2/core/annotate.py
@@ -101,6 +101,7 @@ class Annotation():
                 annotation.annotation_type = meta_yaml['type']
                 annotation.created_at = meta_yaml['created_at']
 
+        annotation._filepath = filepath
         return annotation
 
     # We never expect this to be hit as the base class for Annotations
@@ -332,11 +333,11 @@ class Note(Annotation):
 
         note_path = os.path.join(annotation_uuid_dirname, 'note.txt')
 
-        if self._filepath:
+        if self.contents:
+            contents = self.contents.encode('utf-8')
+        else:
             with open(self._filepath, 'rb') as fh:
                 contents = fh.read()
-        else:
-            contents = self.contents.encode('utf-8')
 
         # validation for max size and parsability
         max_size = 10 * 1024 * 1024
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -1182,9 +1182,6 @@ class Cache:
                     Result._from_archiver(Archiver.load_raw(destination, self))
                 existing.merge_annotations(ref)
 
-    # TODO: If multiple instances of the same artifact are put into the cache
-    # with different annotations, we merge the annotations. If there are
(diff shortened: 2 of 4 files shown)


qiime2/core/annotate.py
```diff
@@ -101,6 +101,7 @@ def load(cls, filepath):
                 annotation.annotation_type = meta_yaml['type']
                 annotation.created_at = meta_yaml['created_at']
 
+        annotation._filepath = filepath
```
Oddant1, line 104: This attr wasn't being put on the object at all if it was created this way.

qiime2/core/annotate.py
```diff
@@ -332,11 +333,11 @@ def _write(self, annotations_dir, root_result_uuid,
 
         note_path = os.path.join(annotation_uuid_dirname, 'note.txt')
 
-        if self._filepath:
+        if self.contents:
+            contents = self.contents.encode('utf-8')
```
Oddant1, line 337: I assume we would prefer getting the contents over reading the file because it potentially saves us from reading the file from disk again.

qiime2/core/cache.py
```diff
@@ -900,10 +900,15 @@ def save(self, ref, key):
...
+            try:
+                self._register_key(key, str(ref.uuid))
+                self._copy_to_data(ref)
+            except:  # noqa: E722
+                if key in self.get_keys():
+                    self.remove(key)
+                raise
 
-        return self.load(key)
+            return self.load(key)
```
Oddant1, line 924: I realized this needed to be made safer while testing the annotation merging. Before annotation merging was fully implemented, I encountered scenarios where I would add a duplicate artifact to the cache, and it would error out while adding it, but the new key would still be in the cache.

Oddant1:
Left some review comments explaining why I made a few changes that don't seem immediately relevant to the new functionality.

lizgehret (approved):
this all makes sense @Oddant1, thanks! could you resolve the merge conflicts? once that's done, i can get this merged.

Oddant1:
@lizgehret conflicts should be dealt with should be good to merge when ci passes

PR #875, opened 2025-10-09
MAINT: Only remove set permissions none o that fancy thread

Oddant1:
Only the removal of set permissions in #849 not the thread that is causing issues on MacOS.

The only potential issue with this is if someone has a very old cache that still has read-only permissions set on something. The cache has not set those permissions for some time because `set_permissions` has been no-oped on anything but `USER_GROUP_RWX` for some time. The only one of these calls that actually did anything was the one setting those permissions in GC to make sure it could be garbage collected.

### DIFF
--- a/qiime2/core/cache.py
+++ b/qiime2/core/cache.py
@@ -837,6 +837,12 @@ class Cache:
                     # We may not have permissions because old versions of
                     # QIIME 2 set entries in data to read-only. If we encounter
                     # that then set write permissions here and try again.
+                    #
+                    # This try-except will induce a slight performance overhead
+                    # in Python versions pre 3.11, but much less than running
+                    # set_permissions every time. In Python 3.11 and on,
+                    # try-except introduces no performance penalty if an
+                    # exception is not raised.
                     except PermissionError as e:
                         if e.errno == 13:
                             set_permissions(target, None, USER_GROUP_RWX)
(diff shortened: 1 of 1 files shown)


qiime2/core/cache.py
```diff
@@ -833,9 +832,16 @@ def garbage_collection(self):
...
+                    try:
+                        shutil.rmtree(target)
+                    # We may not have permissions because old versions of
+                    # QIIME 2 set entries in data to read-only. If we encounter
+                    # that then set write permissions here and try again.
+                    except PermissionError as e:
+                        if e.errno == 13:
+                            set_permissions(target, None, USER_GROUP_RWX)
+                            shutil.rmtree(target)
+                        raise e
```
Oddant1: This try-except will induce a slight performance overhead in Python versions pre 3.11, but much less than running set_permissions every time. In Python 3.11 and on, try-except introduces no performance penalty if an exception is not raised.

lizgehret (approved):
lgtm; also tested locally and everything passes. changes make sense to me (and we also discussed this in more depth offline, for posterity). thanks @Oddant1!

lizgehret:
I'll leave it up to you whether you want to changelog this but I'd guess this is fine to leave out (unless you want to mention the edge case on an older cache).

PR #876, opened 2025-10-14
IMP: add migration warning on `Action` class

lizgehret:
needed for https://github.com/bokulich-lab/RESCRIPt/pull/243
-------------------------
blocking https://github.com/qiime2/q2cli/pull/382
PR for user testing https://github.com/qiime2/q2-vizard/pull/83


lizgehret:
@gregcaporaso feedback - only required field be `to_plugin`, all others optional. Update warning message accordingly to handle differing output messaging.

qiime2/sdk/action.py
```diff
@@ -46,6 +49,46 @@ def _coerce_pipeline_outputs(ctx, outputs):
     return tuple(coerced_outputs)
 
 
+# TODO: validation on these params via data.yaml in distributions
```
Oddant1, line 52: Discussed at meeting, we don't want to do this

qiime2/sdk/action.py
```diff
@@ -46,6 +49,46 @@ def _coerce_pipeline_outputs(ctx, outputs):
...
+class MigrationInfo(TypedDict):
+    to_plugin: str
+    from_distro: str
+    to_distro: str
```
Oddant1, line 56: Discussed at meeting, make these optional

Oddant1 (commented):
We should probably add a `migrated_method` to the `dummy-plugin`

lizgehret:
@gregcaporaso feedback - only required field be `to_plugin`, all others optional. Update warning message accordingly to handle differing output messaging.

cherman2:
Liz and I talked about this offline, but when running `qiime vizard heatmap --help`, The warning placement is below the description of the method. It may make more sense for it to appear at the top or bottom of the help text because thats where users would expected cli warnings or errors.

qiime2/sdk/tests/test_action.py
```diff
@@ -181,3 +181,102 @@ def test_deprecation_warning(self):
 
     def test_docstring(self):
         self.assertIn('Method is deprecated', self.method.__call__.__doc__)
+
+
+class TestMigration(unittest.TestCase):
```
Oddant1, line 192: So you're gonna hate me, but we probably also want a "Missing the mandatory key" a "has an unallowed key" and a "is just set to some nonsense that isn't valid at all." Normally I wouldn't say we need all of the combinations of optional params... but since that changes the way the message is created and we want the message to be sane regardless of which are set that was probably a good idea, and it's already done so overall :+1: sorry for your pain.

qiime2/core/testing/method.py
```diff
@@ -55,6 +55,38 @@ def deprecated_method() -> dict:
     return {'foo': 43}
 
 
+def migrated_method1() -> dict:
```
Oddant1: Test failures are because there is a test that asserts the Actions registered to the dummy-plugin are the ones we expect, and also that the Methods registered to the dummy-plugin are the ones we expect, so these will need to be added to that test.

qiime2/sdk/tests/test_action.py
lizgehret: nope this all makes sense; tests have been added!

Oddant1 (approved):
lgtm @lizgehret. Is this ready for merge?

lizgehret:
> lgtm @lizgehret. Is this ready for merge?

Yep! Once this is merged I can get the q2cli PR passing and finalize tests there as well.

PR #884, opened 2025-11-19
MAINT: Refactor prov lib to natively use Results

### LINKED ISSUE #879: Refactor Provenance Lib to Load a Result as use the Result/Archiver API
**Improvement Description**
Provenance-lib was originally written as a separate entity from QIIME 2, and as a result, it handles the artifact as a raw filepath or a zipfile, not a Result.

**Current Behavior**
A bunch of functionality is reimplemented in provenance lib that is already implemented on Results ex: #878

There are also existing issues that are annoying to handle right now that could be much more cleanly handled if we were loading a Result ex: qiime2/q2cli#310 

**Proposed Behavior**

- [x] Load our filepath as a Result at some point early on in the process and consolidate code into the Result/Archiver API

Oddant1:
closes #879


Oddant1:
- [ ] Update all docstrings and whatnot to have consistent types once things are finalized

qiime2/core/archive/provenance_lib/archive_parser.py
```diff
@@ -199,36 +185,89 @@ def _parents(self) -> Optional[List[Dict[str, str]]]:
...
+        archiver: Archiver,
+        *args,
+        archive_version: str = None,
+        framework_version: str = None,
+        uuid: str | None = None
     ):
         '''
-        Constructs a ProvNode from a zipfile and the collected
-        provenance-relevant filepaths for a single result within it.
+        Constructs a ProvNode from a Result and the uuid of a Result within
```
colinvwood: From an `Archiver` not a `Result`

qiime2/core/archive/provenance_lib/archive_parser.py
```diff
@@ -199,36 +185,89 @@ def _parents(self) -> Optional[List[Dict[str, str]]]:
...
+        if archive_version is None or framework_version is None:
+            archive_version, framework_version = parse_version(archiver, uuid)
+
+        self._archive_version = archive_version
+        _archive_version = float(archive_version)
+        self._framework_version = framework_version
+
+        #  Set up the base path we are looking under for files
+        self._uuid = uuid if uuid else str(archiver.uuid)
+        if uuid is None:
```
colinvwood, line 227: Do we use `uuid=None` to indicate the root node(s)?

qiime2/core/archive/provenance_lib/archive_parser.py
```diff
@@ -199,36 +185,89 @@ def _parents(self) -> Optional[List[Dict[str, str]]]:
...
+                action_path = \
+                    base_path / 'provenance' / 'action' / 'action.yaml'
+            self.action = _Action(action_path)
+
+        # Parse the root metadata
+        metadata_path = base_path / 'metadata.yaml'
+        self._result_md = _ResultMetadata(metadata_path)
+
+        # Parse the citations
+        if _archive_version >= 4:
```
colinvwood, line 247: Why the version guards here, they weren't needed before?

qiime2/core/archive/provenance_lib/archive_parser.py
```diff
@@ -600,58 +627,74 @@ def parse_prov(self, cfg: Config, data: Any) -> ParserResults:
...
+        # NOTE: I would love to set result, archive_version, and
+        # framework_version as instance state here; however, to maintain the
+        # legacy API, I am not
+        archive_version, _ = parse_version(artifact_data)
+
+        if archive_version in FORMAT_REGISTRY:
+            return FORMAT_REGISTRY[archive_version]()
+
+        # Minor versions should more or less support future minor versions
+        if '.' in archive_version:
```
colinvwood, line 649: This is all suspiciously similar to `Archiver.get_format_class`

qiime2/core/archive/provenance_lib/archive_parser.py
```diff
@@ -946,136 +832,76 @@ def parse_prov(self, cfg: Config, archive: str) -> ParserResults:
...
-                if node_uuid in archive_contents:
-                    continue
-
-                # different artifact versions have different expected files
-                if 'artifacts' in fp.parts:
-                    #      0         1          2        3
-                    # /root-uuid/provenance/artifacts/node-uuid
-                    nested_path = pathlib.Path(*fp.parts[1:4])
-                    archive_version, _ = parse_version(zf, nested_path)
-                    parser = FORMAT_REGISTRY[archive_version]
```
colinvwood, line 988: didn't I just write this smh

qiime2/core/archive/provenance_lib/replay.py
```diff
@@ -777,6 +777,14 @@ def build_action_usage(
     action = node.action.action_name
     plg_action_name = ns.uniquify_action_name(plugin, action)
 
+    # Determine if the action we are looking at is present in the environment
+    # we are using
+    plugin_obj = cfg.pm._plugin_by_id.get(node.action.plugin)
+    if plugin_obj:
+        node.set_action_present(node.action.action_name in plugin_obj.actions)
+    else:
+        node.set_action_present(False)
```
colinvwood: Seems like we can just make a call to the `ProvNode.action_present` property and it will handle all this logic

qiime2/sdk/usage.py
```diff
@@ -1661,6 +1662,62 @@ def factory(name=param_name):
...
+                var_type = 'artifact'
+        else:
+            # If we are a pipeline we need to recurse until we find our root
+            alias_uuid = action_yaml['action']['alias-of']
+            alias_path = node._archiver.provenance_dir / 'artifacts' \
+                / alias_uuid / 'action' / 'action.yaml'
+
+            with open(alias_path) as fh:
+                alias_yaml = yaml.safe_load(fh)
+                var_type = self._find_var_type_from_prov(node, alias_yaml)
```
colinvwood: if this is deindented we won't potentially keep many files open at once (idk how far this can actually recurse)

qiime2/core/archive/provenance_lib/archive_parser.py
Oddant1: The version guard still basically existed only in a different form. Prior to this, we were collecting all the files in the archive then dispatching logic here based on which files we found, so we would only try to create citations here if we saw a citations file.

qiime2/core/archive/provenance_lib/archive_parser.py
Oddant1: Yes, I need to finish that docstring. Knew I was forgetting some. Sorry the dev cycle for this PR was a mess what with me getting sick and all

qiime2/core/archive/provenance_lib/archive_parser.py
Oddant1:
```suggestion
        Constructs a ProvNode from an Archiver and the uuid of a Result within
```

Oddant1: Rly wanted this to let me commit from here

qiime2/sdk/usage.py
Oddant1: Arbitrarily far. We can nest pipelines inside of pipelines arbitrarily, and every nesting level will come with an alias we need to dereference. Doesn't mean this can't be optimized in some way though. We don't need to keep the old file open once we've read the alias.

qiime2/core/archive/provenance_lib/archive_parser.py
Oddant1: Yeah, I felt bad about deleting it 😮‍💨, but I do prefer ditching the whole "expected files" system and just trying to parse the files that a given archive version should have without going through the whole "does the file exist" process first. We'll know an expected file is missing if we try to parse it and can't. Feels a lot more straightforward to me

qiime2/core/archive/provenance_lib/archive_parser.py
Oddant1: It is; however, that method returns the class itself, and then I would need to parse the version string I want out of it

qiime2/core/archive/provenance_lib/replay.py
Oddant1: I think you're right... but my god does it look weird. I feel like I accidentally engineered something kinda clever without realizing how clever it was then failed to use it properly.

qiime2/core/archive/provenance_lib/tests/test_parse.py
colinvwood: This is one of the last two remaining usages of this context manager that has to hack around with zip files. Is it worth refactoring this one to use the archiver and `os.remove` (and the one other usage) just to be rid of it?

colinvwood: You removed the other four usages in this file

colinvwood (approved):

colinvwood (approved):

PR #885, opened 2025-11-20
Checksum caching

colinvwood:


colinvwood:
The `diversity core-metrics` pipeline with moving pictures tutorial data is ~15% faster with these changes (58.8s vs 69.5s, ten iterations each).

colinvwood:
The `boots core-metrics` pipeline with moving pictures tutorial data and `n=10` is ~20% faster (348s vs 440s, ten iterations each).

colinvwood:
#883 first

qiime2/core/archive/format/v5.py
```diff
@@ -6,8 +6,12 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
+from pathlib import Path
+import os
+
 import qiime2.core.archive.format.v4 as v4
-from qiime2.core.archive.format.util import write_checksums
```
Oddant1, line 10: I believe this function is now entirely unused and can be removed.

qiime2/sdk/result.py
```diff
@@ -1097,11 +1107,128 @@ def values(self):
...
+
+            if path.is_relative_to(pathlib.Path("provenance/artifacts")):
+                key = path.relative_to(pathlib.Path("provenance/artifacts"))
+            elif path.is_relative_to(pathlib.Path("provenance")):
+                key = (
+                    pathlib.Path(str(artifact.uuid)) /
+                    path.relative_to(pathlib.Path("provenance"))
+                )
+            else:
+                continue
```
Oddant1, line 1220: Maybe put a comment explaining this continue is here because paths that are not under the above are not included in nested provenance. Just lowers the mental overhead of reading this a bit.

qiime2/sdk/result.py
```diff
@@ -778,6 +782,12 @@ def validate(self, level='max'):
 
         self.format.validate(self.view(self.format), level)
 
+    def get_checksums(self) -> dict[str, str]:
+        return self._archiver.get_checksums()
+
+    def validate_checksums(self):
```
Oddant1, line 788: Why is this called `validate_checksums` when all it does is call `validate`

qiime2/core/archive/format/v5.py
```diff
@@ -25,8 +29,39 @@ def write(cls, archive_record, type, format,
...
+                if checksum_string is None:
+                    checksum_string = checksum(str(path), cls.CHECKSUM_TYPE)
+
+                checksums[str(archive_relative_path)] = checksum_string
+
+        with (Path(archive_record.root) / cls.CHECKSUM_FILE).open('w') as fh:
+            for item in checksums.items():
+                # always ignore annotations dir when writing checksums
+                # so we dont self-invalidate when adding/removing annotations
+                if not item[0].startswith('annotations'):
```
Oddant1: Should we skip even calculating the checksums for annotations? Right now it looks like we calculate the checksums for the annotations then never write them or do anything with them, because as the comment says annotations are mutable.

qiime2/core/archive/archiver.py
```diff
@@ -303,7 +304,7 @@ def _destroy_temp_path(cls, process_alias):
         cache.process_pool.remove(str(process_alias))
 
     @classmethod
-    def get_format_class(cls, version):
+    def get_format_class(cls, version) -> ArchiveFormat | None:
```
Oddant1, line 307: What is this? C? You think type annotations count for more than thoughts and prayers here 😮‍💨

qiime2/sdk/tests/test_checksum_cache.py
```diff
@@ -0,0 +1,121 @@
...
+    def setUp(self):
+        ChecksumCache().cache = {}
+
+    @unittest.mock.patch(
+        'qiime2.core.util.checksum_python', wraps=checksum_python
+    )
+    @unittest.mock.patch(
+        'qiime2.core.util.checksum_native', wraps=checksum_native
+    )
+    def test_checksum_cache_fewer_checksum_calls(
```
Oddant1, line 39: We may also want a test proving the cache works across multiple action executions.

qiime2/sdk/result.py
ebolyen: Drive-by comment:

It looks like this is used by ResultCollection.validate() to avoid the expensive Artifact.validate, and instead dispatches to Result.validate.

That said, since this is Python, we could always just call classmethod directly as `Result.validate(result)` in `ResultCollection` and avoid the whole situation.

Also do any of these need to go on IResult? Who is the main user of ResultCollection.validate_checksums()?

ebolyen:
This is cool!

qiime2/core/archive/format/v5.py
colinvwood: it is tested unfortunately

colinvwood: I'm changing the test

colinvwood: done

qiime2/sdk/result.py
colinvwood: It calls the parent's validate which only validates checksums. Contrast with `self.validate` which validates checksums and its format.

qiime2/core/archive/archiver.py
colinvwood: 🙏🏻

qiime2/core/archive/format/v5.py
Oddant1: > it is tested unfortunately

Tests are unfortunate this is true

qiime2/core/archive/format/v5.py
colinvwood: Good point 👍🏻

qiime2/sdk/tests/test_checksum_cache.py
colinvwood: actually if we cache the same artifact twice we just overwrite the entries... Idk if this worth optimizing though because ultimately the point of the checksum cache is just to avoid calling the checksum algorithm which takes 4ever

colinvwood: I suppose I could cache artifact uuids in parallel and check that cache before potentially re-caching an artifact but I don't think all that is worth it

qiime2/sdk/result.py
colinvwood: @ebolyen correct, I made `Result.validate_checksums` to be used by `ResultCollection.validate_checksums` which in turn is used in `bound_callable` to validate inputs and just to have a consistent api between `Artifact` and `ResultCollection`. I'm unsure what the purpose of `IResult` is, so I can't answer whether any of these new methods need to go on there...

PR #890, opened 2025-12-18
MAINT: Bring back lock refresh thread

Oddant1:
Thread that refreshes lock periodically for long running locking sections.


ebolyen (approved):
Still LGTM. Is it worth reducing the lifetime of the lock if we have a stable way to keep it alive? I could see 2 minutes being almost imperceptible in the context of an HPC job.

Oddant1:
Maybe we can mess with the lock time after merging this as is and seeing if it works as intended at all? It really ought to but y'know...

PR #891, opened 2026-01-06
BUG: handle old artifact formats in checksum cache

colinvwood:
Fixes a bug where artifacts of version < 5 caused the checksum cache to crash due to a missing `CHECKSUM_TYPE`. The test added in this PR is apparently the only test that runs an action on a pre-v5 artifact, which is concerning. I only noticed the bug when running feature table tests which download old artifacts in the examples.

### DIFF
--- a/qiime2/sdk/tests/test_checksum_cache.py
+++ b/qiime2/sdk/tests/test_checksum_cache.py
@@ -6,15 +6,21 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
+from pathlib import Path
+import tempfile
 import unittest
 import uuid
 
+import qiime2
 from qiime2 import Artifact
 from qiime2.sdk.result import ChecksumCache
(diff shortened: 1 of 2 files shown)


Oddant1:
This looks good to me pending @ebolyen deciding how we want to handle this in general

qiime2/sdk/tests/test_checksum_cache.py
```diff
@@ -119,3 +125,22 @@ def test_checksum_cache_handles_checksum_algorithm_boundaries(self):
...
+
+    def test_checksum_cache_handles_artifacts_with_no_checksums(self):
+        '''
+        Tests that artifacts with a format that does not contain checksums are
+        handled by the checksum cache.
+        '''
+        v3_artifact_dir = (
+            Path(qiime2.__file__).parent / 'core' / 'archive' /
+            'provenance_lib' / 'tests' / 'data' / 'concated-ints-v3'
+        )
```
Oddant1, line 137: Clever

ebolyen:
Thanks guys!

PR #895, opened 2026-01-09
Adds a QIIME2 warning class

Macabe222:
Adds a QIIME2 warning class in order to aid in making certain warnings visible. 
~~This is blocking https://github.com/qiime2/q2cli/pull/387~~

### DIFF
--- a/qiime2/core/exceptions.py
+++ b/qiime2/core/exceptions.py
@@ -15,6 +15,9 @@ class ImplementationError(Exception):
     pass
 
 
-class QIIME2Warning(UserWarning):
-    """Custom QIIME2 warning that will always be displayed"""
+class RachisWarning(UserWarning):
+    """
+    A custom warning that will always be displayed in the CLI, whether or not
+    the --verbose flag is set.
+    """
     pass
--- a/qiime2/core/exceptions.py
+++ b/qiime2/core/exceptions.py
@@ -13,3 +13,8 @@ class ValidationError(Exception):
 
 class ImplementationError(Exception):
     pass
+
+
+class QIIME2Warning(UserWarning):
+    """Custom QIIME2 warning that will always be displayed"""
+    pass


ebolyen:
This will end up being an issue with the Rachis migration. Most of our objects actually don't say QIIME2 anywhere, which has been nice. Can we come up with a name like `VisibleWarning` or `InterfaceWarning` which can be used sensibly and survive the absence of "QIIME 2" as a useful name?

gregcaporaso:
@ebolyen, what do you think about `RachisVisibleWarning` or `RachisWarning`? Or do you feel strongly that it shouldn't have the framework name it? Just seems like it could be helpful to have that context on where it's coming from. 

Update: @colinvwood and @Oddant1 are discussing now - feels like if people see that it's coming from our framework they may be more inclined to pay attention to it. My preference would be `RachisWarning`. If you feel strongly though, we're all ok with alternatives.

colinvwood:
I'm going to merge as `RachisWarning` but we just rename if we want later

ebolyen:
I suppose that's fine, but it'll be a breaking change since this seems to be for plugins right?

My feelings are it's not really Rachis saying anything, rather it's an output stream for plugins to communicate "concerns" with the user. It could just as well be an open file descriptor that interfaces and plugins know to use. `PluginWarning` or `RachisPluginWarning` I think might be better.

Oddant1:
I think `RachisWarning` was settled on mostly because tech illiterate people will see that and know clearly what Rachis is so they will know this is specifically the piece of software they are using trying to tell them something

gregcaporaso:
@ebolyen, if you feel strongly and want to change it to `PluginWarning` or `RachisPluginWarning`, no objections from this end. Of those, I would lean toward `RachisPluginWarning`, because I feel like it indicates that the system they're working with is trying to communicate something to them (as opposed to a more generic warning, like `UserWarning`, ...) but you can make the call.

ebolyen:
It's probably not worth the effort. Worst case this can be a base class that gets differentiated if needed.

PR #897, opened 2026-01-13
MAINT: Skip the parse perms test if we are root

Oddant1:
This will skip the perms test for parsing archives if we are root. I am not convinced this REALLY addresses the root cause of this test failing in the prepare, but it will prevent it from running in the prepare, and that will stop it from failing. This isn't a very important test, so that's fine. I would honestly also be ok with entirely deleting it. This way the test will still run if someone runs the test locally.


lizgehret:
This seems fine enough- I agree that maybe it's ultimately better to delete in the long term unless we can find a way to actually get this kind of permission error to be raised again, which seems much more informative than the current outcome.

In any case, I ran locally to confirm it does still run - so I'll go ahead and merge this and see if this fixes the failing test in the prepare.

PR #898, opened 2026-01-14
add dummy method that raises RachisWarning

colinvwood:
needed for qiime2/q2cli#387

### DIFF
--- a/qiime2/core/testing/method.py
+++ b/qiime2/core/testing/method.py
@@ -272,4 +272,4 @@ def _underscore_method() -> int:
 
 def raises_rachis_warning() -> int:
     warnings.warn('This is an important warning.', RachisWarning)
-    return 67
+    return 42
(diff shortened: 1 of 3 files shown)


qiime2/core/testing/method.py
```diff
@@ -266,3 +268,8 @@ def random_seed_method_never_set(
 
 def _underscore_method() -> int:
     return 42
+
+
+def raises_rachis_warning() -> int:
+    warnings.warn('This is an important warning.', RachisWarning)
+    return 67
```
Oddant1:
```suggestion
    return 42
```

I will not tolerate this

ebolyen: it's important to be inclusive, we don't want to turn away the youths

Oddant1: It is our responsibility to educate them, not coddle them

PR #899, opened 2026-01-20
MAINT: remove validate calls

colinvwood:
No longer performs checksum validation of inputs during actions, due to performance concerns with large input artifacts, e.g. those encountered in MOSPHIT.


ebolyen (approved):
Thanks @colinvwood!

PR #908, opened 2026-02-05
Validate usage output names (and just standardize the method we use to do that to a util)

Oddant1:
Validate that usage output names follow the same rules as every other name we use, valid Python id + '-', this prevents real weird stuff like qiime2/q2-feature-table#353. This ought to get things close enough to correct for any given interface that `to_interface_name` can handle the rest.

### DIFF
--- a/src/rachis/util.py
+++ b/src/rachis/util.py
@@ -161,7 +161,7 @@ def is_valid_python_identifier(given_string, var_name):
     ------
     ValueError
         If the string passed in is not a valid Python identifier.
-        """
+    """
     replaced_string = given_string.replace('-', '_')
     if not replaced_string.isidentifier():
         raise ValueError(f'{var_name.capitalize()} "{given_string}" is not a'
--- a/src/rachis/util.py
+++ b/src/rachis/util.py
@@ -147,7 +147,7 @@ def get_filepath_from_package(package, relative_filepath):
 
 
 def is_valid_python_identifier(given_string, var_name):
-    """Validates that the given strng is a valid Python idenitifier with the
+    """Validates that the given string is a valid Python idenitifier with the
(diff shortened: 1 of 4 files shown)


ebolyen:
Should q2cli be doing something similar to this:
https://github.com/rachis-org/rachis/blob/dev/src/rachis/plugins.py#L52-L53

i.e. for shells the typical `shlex.quote`?

gregcaporaso:
Just one fly-by comment we talked about yesterday - probably good to get this merged early in the dev cycle, just in case it triggers failures (because of broken usage examples, as in https://github.com/qiime2/q2-feature-table/issues/353) elsewhere through the ecosystem.

gregcaporaso:
@colinvwood, would you be comfortable reviewing this one?

colinvwood:
Can do 👍🏻

src/rachis/util.py
```diff
@@ -144,3 +144,29 @@ def get_filepath_from_package(package, relative_filepath):
             f'The requested data asset {fp} does not exist.')
     else:
         return fp
+
+
+def is_valid_python_identifier(given_string, var_name):
+    """Validates that the given strng is a valid Python idenitifier with the
```
colinvwood:
```suggestion
    """Validates that the given string is a valid Python idenitifier with the
```

colinvwood (approved):

Oddant1:
@colinvwood I made a couple changes here. Decided I like having methods on the classes that call the util.

1. It lets them vary up the functionality a bit between themselves if they need. If their validation needs more than just that t can stay in one spot easily.
2. Leaving those methods there means we aren't making any API changes

colinvwood (approved):

PR #909, opened 2026-02-09
TST: fix mock handling in Py3.13 

ebolyen:
Checking for Python 3.14 Linux tests.

### DIFF
--- a/src/rachis/sdk/tests/test_checksum_cache.py
+++ b/src/rachis/sdk/tests/test_checksum_cache.py
@@ -62,8 +62,8 @@ class TestChecksumCache(unittest.TestCase):
             calls_without_cache = get_call_count()
 
         # with cache
-        native_mock.call_count = 0
-        python_mock.call_count = 0
+        native_mock.reset_mock()
+        python_mock.reset_mock()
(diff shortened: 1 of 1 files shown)


ebolyen:
@colinvwood, it seems that zeroing the call count stops working in Python 3.13+ on Linux. I have no _nice_ explanation for this, but using the reset_mock method seems to restore the intended functionality.

ebolyen:
Extra reverts added so you can see the behavior and [here's a cron with a larger text matrix](https://github.com/rachis-org/rachis/actions/runs/21773490210).

ebolyen:
I believe this is the reason for the behavior: https://github.com/python/cpython/pull/142744

It corresponds to:
Python 3.13.11 pass
Python 3.13.12 fail (released Feb 3rd)
Python 3.14.2 pass
Python 3.14.3 fail (released Feb 3rd)

This also lines up with the success of the original PR which ran on a build prior to 3.14.3.

ebolyen:
I've let [upstream](https://github.com/python/cpython/issues/142651#issuecomment-3872242970) know the issue, but we should still use the right method I think.

colinvwood (approved):
makes sense 👍🏻

PR #914, opened 2026-02-25
MAINT: update tuplize

ebolyen:
Bringing `tuplize` into the modern age.

(We can now annotate multiple return types correctly without breaking backwards compatibility.)

PR #916, opened 2026-03-10
DEP: remove decorator

ebolyen:
Took a _lot_ of back-and-forth with `codex`, but modern Python versions basically obviate the need for `decorator` entirely, which is nice.

I'm also pretty pleased that our own `Signature` can take over the job here (I had to prod codex to see that change).

### DIFF
--- a/src/rachis/core/type/signature.py
+++ b/src/rachis/core/type/signature.py
@@ -244,6 +244,21 @@ class PipelineSignature:
         collated_inputs.update(kwargs)
         return collated_inputs
 
+    def to_inspect_signature(self):
+        parameters = []
+        kind = inspect.Parameter.POSITIONAL_OR_KEYWORD
+        for name, spec in self.signature_order.items():
+            default = inspect.Parameter.empty
+            if spec.has_default():
+                default = spec.default
+            parameters.append(inspect.Parameter(
+                name, kind, default=default, annotation=spec.qiime_type))
+
+        output_types = tuple(spec.qiime_type for spec in self.outputs.values())
+        return_annotation = tuple[output_types]
+        return inspect.Signature(
+            parameters=parameters, return_annotation=return_annotation)
+
     def _assert_valid_inputs(self, inputs):
         for input_name, spec in inputs.items():
             if not is_semantic_type(spec.qiime_type):
--- a/src/rachis/plugin/tests/test_plugin.py
+++ b/src/rachis/plugin/tests/test_plugin.py
@@ -473,6 +473,5 @@ class TestPlugin(unittest.TestCase):
                             ' bad.'
             )
 
-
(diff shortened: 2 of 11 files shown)


src/rachis/sdk/action.py
```diff
@@ -389,23 +378,13 @@ def _set_wrapper_properties(self, wrapper):
...
-        output = []
-        for spec in self.signature.outputs.values():
-            output.append(spec.qiime_type)
-        output = tuple(output)
-
-        annotations["return"] = output
-
+        annotations = {
+            name: param.annotation
+            for name, param in self._wrapper_signature.parameters.items()}
```
ebolyen, line 387: Same idea as before, but signature is pre-computed and has the annotations we wanted now.

src/rachis/sdk/action.py
```diff
@@ -377,10 +364,12 @@ def parsl_wrapper(*args, **kwargs):
         return parsl_wrapper
 
     def _rewrite_wrapper_signature(self, wrapper):
-        # Convert the callable's signature into the wrapper's signature and set
-        # it on the wrapper.
-        return decorator.decorator(
-            wrapper, self._callable_sig_converter_(self._callable))
+        def signature_wrapper(*args, **kwargs):
+            bound = self._wrapper_signature.bind(*args, **kwargs)
+            bound.apply_defaults()
```
ebolyen, line 370:
```suggestion
            # ensures that provenance has all default arguments.
            bound.apply_defaults()
```

ebolyen:
Should be ready for review. cc @lizgehret (who is interested in this for seed-environment resolution).

src/rachis/core/type/signature.py
```diff
@@ -244,6 +244,21 @@ def collate_inputs(self, *args, **kwargs):
...
+        kind = inspect.Parameter.POSITIONAL_OR_KEYWORD
+        for name, spec in self.signature_order.items():
+            default = inspect.Parameter.empty
+            if spec.has_default():
+                default = spec.default
+            parameters.append(inspect.Parameter(
+                name, kind, default=default, annotation=spec.qiime_type))
+
+        output_types = tuple(spec.qiime_type for spec in self.outputs.values())
+        return_annotation = tuple[output_types]
```
ebolyen, line 258: This is why the tests needed to be updated. It's more of an oversight in #914 than anything else.

src/rachis/plugin/tests/test_plugin.py
```diff
@@ -473,6 +473,5 @@ def bad_function() -> list:
                             ' bad.'
             )
```
ebolyen, line 475:
```suggestion


```

lizgehret (approved):
this is great, nice to get rid of all this 10yo code. see ya decorator!

PR #917, opened 2026-03-17
Optional view types on pipelines

Oddant1:
A bit of cleanup and additional testing also enforce dict[str, ]


Oddant1:
Discussed out of band

src/rachis/sdk/tests/test_pipeline.py
```diff
@@ -15,8 +15,9 @@
 import rachis.sdk
 from rachis.core.testing.util import get_dummy_plugin
 from rachis.core.testing.type import IntSequence1, SingleInt, Mapping
-from rachis.plugin import Visualization, Int, Bool
+from rachis.plugin import  Visualization, Int, Bool
```
ebolyen:
```suggestion
from rachis.plugin import Visualization, Int, Bool
```

I'm suprised the linter didn't see this.

src/rachis/sdk/tests/test_pipeline.py
```diff
@@ -355,10 +356,98 @@ def test_nested_pipeline_parallel(self):
...
+                ('out6', Collection[Visualization])
+            ],
+            name='Uses only collection types.',
+            description='Uses only collection types.'
+        )
+
+        # The entire test with this pipeline is that it registered at all. We
+        # Don't care about calling it. Unregister it here so funny things don't
+        # happend when we audit what is registered to the dummy_plugin
+        self.plugin.pipelines.pop('annotations')
```
ebolyen: You could construct a `PipelineSignature` object and avoid mutating the plugin. It is mostly all the registration does in the first place.

src/rachis/sdk/tests/test_pipeline.py
```diff
@@ -355,10 +356,98 @@ def test_nested_pipeline_parallel(self):
 
         self.assertTrue(True)
 
+    def test_annotations_correct(self):
+        def annotations(
+                    ctx,
```
ebolyen, line 361: Should we allow annotations for this? We would probably need an interface type over in `rachis.plugin` but that might genuinely be a good thing, because the discoverability of context isn't great at the moment.

ebolyen: (especially in a world where an interface might genuinely swap out the context, having a more visible API contract would be a good thing for it)

ebolyen (approved):
Just a minor note about the test and question about context annotation.

It's good to go as is though.

src/rachis/sdk/tests/test_pipeline.py
Oddant1: Currently looking into this. I'm assuming we would make the annotation always just the base `Context` class, and based on your thoughts here the real point would be for people to actually see what Context is and what it does

ebolyen: yeah, I think that would be the goal. If we wanted an `IContext` I have no specific objection, but it seems over-complicated at the moment.

Any other context implementations should cover that minimal `Context` API and plugins really shouldn't use anything more than that.

src/rachis/sdk/tests/test_pipeline.py
```diff
@@ -368,16 +476,24 @@ def mixed_annotations(
                     "Pipelines must have annotations for all inputs,"
                     " parameters, and outputs or no annotations."
                 ):
-            self.plugin.pipelines.register_function(
-                function=mixed_annotations,
+            rachis.sdk.Pipeline._init(
```
ebolyen: Would `PipelineSignature` be easier? or did you need a real callable?

Oddant1: I suppose that's close enough

ebolyen:
Thinking about `Context` more. I think creating an `IContext` in `rachis.plugin` might actually be a good idea. We have had plugins import context and instantiate it themselves, which actually isn't really what we want (do you remember a handful of pipelines in diversity?)

But this interface object would not have an implementation and would posses only the API surface a plugin is allowed to use. Then interfaces can add whatever they want/need to their own Context object which implements `IContext`, and we would get eventual testing of that contract via static analysis once our plugins are a little more up to snuff on type things.

I think all of that can happen in a separate PR if you agreed.

PR #918, opened 2026-03-20
Add IContext and use as view_type for ctx

Oddant1:
Add IContext class as interface for Context, and use said interface as view type for ctx

### DIFF
--- a/src/rachis/core/type/signature.py
+++ b/src/rachis/core/type/signature.py
@@ -369,7 +369,7 @@ class PipelineSignature:
         TypeError
             If a view type annotation on an input or an output is invalid.
         """
-        self.annotated = self._assert_all_annotated_or_unannotated(
+        annotated = self._assert_all_annotated_or_unannotated(
             ctx, inputs, parameters, outputs
         )
 
@@ -391,7 +391,7 @@ class PipelineSignature:
             dict[str, rachis.sdk.Visualization]
         )
 
-        if self.annotated:
+        if annotated:
             if ctx.view_type not in VALID_CTX_ANNOTATIONS:
                 raise TypeError(
                     f'Context has the view type {ctx.view_type}. Valid view'
--- a/src/rachis/core/type/signature.py
+++ b/src/rachis/core/type/signature.py
@@ -24,6 +24,7 @@ from . import meta
 from .util import (is_semantic_type, is_collection_type, is_primitive_type,
                    parse_primitive)
 from ..util import ImmutableBase, checksum, create_collection_name
+from rachis.plugin.context import IContext
 
 
 class __NoValueMeta(type):
@@ -120,7 +121,7 @@ class PipelineSignature:
                              " to define the outputs, as the order is random."
                              % callable.__name__)
 
-        inputs, parameters, outputs, signature_order = \
+        ctx, inputs, parameters, outputs, signature_order = \
             self._parse_signature(callable, inputs, parameters, outputs,
                                   input_descriptions, parameter_descriptions,
                                   output_descriptions)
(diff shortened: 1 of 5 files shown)


src/rachis/plugin/context.py
```diff
@@ -0,0 +1,62 @@
...
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+import abc
+
+
+class IContext(metaclass=abc.ABCMeta):
+    @abc.abstractmethod
+    def _dispatch_(self, args, kwargs):
```
ebolyen: I don't think we want these methods here, since they aren't for plugin developers to use.

Is `IContext` perhaps too broad in retrospect? `CallContext`/`ActionContext`/`PluginContext`? None of those sit quite right.

src/rachis/plugin/context.py
```diff
@@ -0,0 +1,62 @@
...
+    def make_report(self, template, collection):
+        """Create a report based on a template and a collection of
+            visualizations
+        """
+
+    # NOTE: We end up with both the artifact and the pipeline alias of artifact
+    # in the named cache in the end. We only have the pipeline alias in the
+    # process pool
+    @abc.abstractmethod
+    def add_reference(self, ref):
```
ebolyen: We should also remove this one I think

src/rachis/core/type/signature.py
```diff
@@ -168,6 +170,14 @@ def _parse_signature(self, callable, inputs, parameters, outputs,
                 if builtin_args[0] != name:
                     raise TypeError("Missing builtin argument %r, got %r" %
                                     (builtin_args[0], name))
+                if name == 'ctx':
+                    view_type = ParameterSpec.NOVALUE
+                    if parameter.annotation is not parameter.empty:
+                        view_type = parameter.annotation
+                    ctx = ParameterSpec(
```
ebolyen: I don't think we need one of these, right? We could just have `_parse_signature` complain directly, and then everything else can stay the same and we don't need to route the `ctx` around (especially since the interface isn't the object we will actually use, it's just for the type-checker/plugin author to inspect).

ebolyen (changes requested):

src/rachis/plugin/context.py
Oddant1: Yeah I'm not sure. In principle I think we are going to have Context inherit IContext and then everything else inherit Context. I expect it's fine to have literally nothing on IContext... Only that really doesn't help with discoverability of the API. So I think just keeping plugin dev facing methods on IContext and keeping it as IContext or similar is a good idea

src/rachis/core/type/signature.py
Oddant1: Very true. I was just trying to keep all the whining in one place. In theory, we could also whine about all of it in _parse_signature... but I don't think that's a good idea.

ctx is a special case, so I suppose it is acceptable to fully treat it as one.

src/rachis/sdk/context/base.py
```diff
@@ -7,6 +7,7 @@
 # ----------------------------------------------------------------------------
 import warnings
 
+from rachis.plugin import IContext
```
ebolyen, line 10: I'm surprised this didn't cause a circular import. We may want to hoist to `core` it just to be safe.

ebolyen (approved):
This looks a lot cleaner, thanks @Oddant1!

src/rachis/sdk/context/base.py
Oddant1: I think hoisting it to core is what will create circular imports.

PR #919, opened 2026-03-27
IMP: polite user-agent for usage examples

ebolyen:
This should solve the issues we were seeing on Zenodo. Takes the same approach as in https://github.com/qiime2/q2doc/pull/19/ which was found to work without any special token.


lizgehret (approved):
looks good, thanks @ebolyen!

PR #920, opened 2026-03-31
Don't create a temp cache if there is a cache set

Oddant1:
If a cache is set, and the temp cache doesn't exist yet, then we currently create the temp cache even though we won't be using it. Don't do that.

### DIFF
--- a/src/rachis/core/cache.py
+++ b/src/rachis/core/cache.py
@@ -548,8 +548,6 @@ class Cache:
                 if not created_path:
                     # We own the temp_cache_path, so we can recreate it if
                     # there was something wrong with it
-                    print(self.path)
-                    print(temp_cache_path)
                     if self.path == temp_cache_path:
                         set_permissions(self.path, USER_GROUP_RWX,
                                         USER_GROUP_RWX, skip_root=True)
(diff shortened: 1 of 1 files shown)


src/rachis/core/cache.py
```diff
@@ -546,6 +548,8 @@ def __init(self, path=None, process_pool_lifespan=45):
                 if not created_path:
                     # We own the temp_cache_path, so we can recreate it if
                     # there was something wrong with it
+                    print(self.path)
+                    print(temp_cache_path)
```
colinvwood: 🤓

Oddant1: Oops. Must've added those in one of my drunken stupors. You could at least have suggested deleting them so I could just click a button to get rid of them

Oddant1:
```suggestion
```

Make @colinvwood's suggestion for him (he was too lazy to do it himself)

colinvwood: if you leave print statements in your PR you have to fix them yourself (qiime2 COC pg. 3)

colinvwood (approved):

PR #921, opened 2026-04-06
Update CaptureHolder

Oddant1:


src/rachis/core/type/signature.py
```diff
@@ -927,6 +927,44 @@ def _make_hashable(self, collection):
 class CaptureHolder:
```
ebolyen:
```suggestion
T_inner = TypeVar('T_inner')

class CaptureHolder(Generic[T_inner]):
```

src/rachis/core/type/signature.py
```diff
@@ -927,6 +927,44 @@ def _make_hashable(self, collection):
...
     CAPTURE_HOLDER_DEFAULT = None
 
+    @classmethod
+    def get_or_set(cls, instance, setter):
```
ebolyen:
```suggestion
    def get_or_set(cls, instance: None | T_inner | typing.Self, setter: typing.Callable[[], T_inner]) -> T_inner:
```

ebolyen (commented):
I think the above will address the typing concerns you had messaged about.

src/rachis/core/type/signature.py
ebolyen: These come from `typing`

src/rachis/core/type/signature.py
Oddant1: Thank you for bringing sanity back to this conversation before I had to fully learn what sys.settrace did

ebolyen:
It looks like maybe a circular import is happening, but I don't see why these changes would have induced that.

You can test locally via `uv run nox -t test-min`

Oddant1:
Annoying. I don't get that at all locally.

The failure is only happening in Python 3.10. I feel like we don't really care... but it is still strange.

ebolyen (changes requested):
It looks like `spec.view_type == CaptureHolder` doesn't work with 
`view_type = CaptureHolder[int]`

So we need a function like this:
```python
def contains_generic_base(base, expr):
    origin = typing.get_origin(expr)
    if origin is None:
        return issubclass(base, expr)
    if origin is typing.Union:
        return any([contains_generic_base(base, a) for a in typing.get_args(expr)])
    return issubclass(base, origin)

```

Oddant1:
Oh boy, now mystery_stew is unhappy

ebolyen:
Looks like my quickly written in ipython function is saying `True` to something a typemap makes for some reason.

src/rachis/core/type/util.py
```diff
@@ -283,6 +283,8 @@ def parse_primitive(t, value):
 
 
 def contains_generic_base(base, expr):
+    if expr is object or expr is None:
```
ebolyen, line 286: many things are a subclass of `object` it turns out

src/rachis/core/type/signature.py
```diff
@@ -935,7 +935,7 @@ class CaptureHolder(typing.Generic[T]):
     @classmethod
     def get_or_set(
             cls,
-            instance: typing.Self | T | None,
+            instance: 'CaptureHolder' | T | None,
```
ebolyen, line 938: `typing.Self` doesn't exist in py310. A string forward-reference is easy.

ebolyen (approved):

PR #922, opened 2026-04-15
Allow ~ in Collection keys

Oddant1:
Allow ~ in Collection keys. Currently needed for q2-PSEA. May or may not keep.


src/rachis/sdk/util.py
colinvwood: please update your error message sir

colinvwood:
```suggestion
                       '-, ., _, and ~. Offending keys include: '
```

Okay I'm feeling helpful

colinvwood (approved):

PR #929, opened 2026-04-30
IMP: handle large unions of monomorphic collections

ebolyen:
No AI used.

### DIFF
--- a/src/rachis/core/type/tests/test_collection.py
+++ b/src/rachis/core/type/tests/test_collection.py
@@ -152,6 +152,19 @@ class TestCollectionExpression(unittest.TestCase):
         json = Set[Int].encode(value)
         self.assertEqual(json, "[1, 2, 3]")
 
+    def test_collection_of_large_union(self):
+        """Tests functionality used by Signature._infer_type()
+        without all the overhead of creating that.
+        """
+        from rachis.core.type.grammar import UnionExp
+
+        # force a "constructive" definition of Foo (different id() but equal)
+        big = List[UnionExp([SemanticType('Foo')
+                             for _ in range(1000)]).normalize()]
+
+        self.assertEqual(repr(big), repr(List[SemanticType('Foo')]))
+
+
 
 if __name__ == '__main__':
     unittest.main()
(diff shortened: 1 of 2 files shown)


cherman2:
I think it makes sense for @Oddant1 to review this PR. 

@ebolyen is this ready for review? I was under the impression that it was but I don't want to jump the gun here.

src/rachis/core/type/tests/test_collection.py
```diff
@@ -152,6 +152,19 @@ def test_encode_decode_list(self):
...
 
+    def test_collection_of_large_union(self):
+        """Tests functionality used by Signature._infer_type()
+        without all the overhead of creating that.
+        """
+        from rachis.core.type.grammar import UnionExp
+
+        # force a "constructive" definition of Foo (different id() but equal)
+        big = List[UnionExp([SemanticType('Foo')
+                             for _ in range(1000)]).normalize()]
```
Oddant1, line 163: Are you sure 1000 was enough to make your point? 😆

ebolyen:
Yep this is ready!

src/rachis/core/type/tests/test_collection.py
ebolyen: No, I would be happy to increase this number. :)

Oddant1:
@ebolyen, when we run this PR with @cherman2's PR qiime2/q2-boots#60 it appears to strip all properties off the output.

With this PR:
image

Without:
image

ebolyen:
I think this may actually be because the annotation is on a pipeline, and the inner action doesn't produce the property. So when the pipeline checks its outputs it sees the wrong type. I can't quite tell because the traceback is missing.

But that's my guess. I'm going to check locally to see.

ebolyen:
Yeah that's definitely the problem.

So really this PR has nothing to do with solving the issue, it just makes the error less noisy (which is helpful I guess).

ebolyen:
This is still ready to review and fine, but it just isn't as interesting as we initially thought.

Oddant1:
I am going to merge this. I will review #932

PR #932, opened 2026-05-08
NEW: pipelines can now narrow their output types with properties

ebolyen:
Code generated by Codex GPT 5.5 under my direction and detailed review.

This should solve the underlying issue in https://github.com/qiime2/q2-boots/pull/60 which was discovered in #929

This works by allowing `Result._alias()` to optionally take a narrower type which will be used in the archiver instead of the original type. This means that pipelines can solve their outputs and then we can intersect those outputs with the underlying result to define what the new type ought to be. The guard in `Result._alias()` is because I wanted to make sure if it was called for some other reason, we don't get a completely nonsense type attached, but otherwise we should think of the intersection as being totally sufficient here.

### DIFF
--- a/src/rachis/sdk/action.py
+++ b/src/rachis/sdk/action.py
@@ -51,23 +51,15 @@ def _coerce_pipeline_outputs(ctx, outputs):
     return tuple(coerced_outputs)
 
 
-def _raise_incompatible_output_type(expected, output, context=''):
-    observed = output.type if hasattr(output, 'type') else type(output)
+def _intersect_pipeline_output_type(expected, output_type, output_name):
+    refined_type = output_type & expected
+
+    if (not refined_type.is_bottom()) and refined_type.is_concrete():
+        return refined_type
+
     raise TypeError(
-        "Expected output%s to be of type %r, received %r"
-        % (context, expected, observed))
-
-
-def _intersect_pipeline_output_type(expected, output, context=''):
-    try:
-        refined_type = output.type & expected
-    except TypeError:
-        _raise_incompatible_output_type(expected, output, context)
-
-    if refined_type.is_bottom() or not refined_type.is_concrete():
-        _raise_incompatible_output_type(expected, output, context)
-
-    return refined_type
+        "Expected output %s to be of type %r, received %r"
+        % (output_name, expected, output_type))
 
 
 # TODO: validation on these params via data.yaml in distributions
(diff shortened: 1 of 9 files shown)


ebolyen:
@Oddant1 and @cherman2, would you be able to try this again with the boot's PR? I think this will do what we actually want in a way that's pretty useful/intuitive.

Oddant1:
This doesn't seem to be working with https://github.com/qiime2/q2-boots/pull/60 out of the box.

image

Looks to be an issue with `alpha_average_action` as called by `core_metrics_phylogenetic`. This wasn't the only test failure, just the first.

ebolyen:
🙃 looks like the issue is that the function which creates the collection isn't a pipeline, so it is creating all of the inner artifacts and then passing those in. Since there's no pipeline boundary, there's no alias and refinement of the type.

So the error is correct and the problem is with the way the pipeline is written.

ebolyen:
Same deal as #929, this is maybe a nice feature, but doesn't help boots without a new inner action.

src/rachis/sdk/action.py
```diff
@@ -586,32 +605,36 @@ def _callable_executor_(self, ctx, view_args, output_types, provenance):
...
-        # have neither, or our types just don't match up, something bad
+        # have neither, or our types can't be reconciled, something bad
         # happened
         for output, (name, spec) in zip(outputs, output_types.items()):
-            if spec.qiime_type.name == 'Collection' and \
-                    output.collection in spec.qiime_type:
+            if spec.qiime_type.name == 'Collection':
+                if not isinstance(output, rachis.sdk.ResultCollection):
+                    _raise_incompatible_output_type(
+                        spec.qiime_type, output, " %r" % name)
```
Oddant1: Isn't this redundant? We have the return from the Action, which should already have been coerced into a ResultCollection if possible and raised an Exception if not.

src/rachis/sdk/action.py
```diff
@@ -51,6 +51,25 @@ def _coerce_pipeline_outputs(ctx, outputs):
     return tuple(coerced_outputs)
 
 
+def _raise_incompatible_output_type(expected, output, context=''):
+    observed = output.type if hasattr(output, 'type') else type(output)
+    raise TypeError(
+        "Expected output%s to be of type %r, received %r"
```
Oddant1: Why "output%s" with " " in the context and not just "output %s"? Seems to be in case context='', but context is provided in every place this is called, and it probably should always be provided because it's the name of the output which we will always have.

src/rachis/sdk/action.py
```diff
@@ -51,6 +51,25 @@ def _coerce_pipeline_outputs(ctx, outputs):
     return tuple(coerced_outputs)
 
 
+def _raise_incompatible_output_type(expected, output, context=''):
```
Oddant1: I don't think I like this overloading of the term "context." Maybe "output_name" or something?

src/rachis/sdk/action.py
ebolyen: I think there used to be a comma here, I'll fix that.

src/rachis/sdk/action.py
ebolyen: Ah yes you are right! I think that means I can simplify the `_intersect` helper and remove `_raise_incompatible` which will be nice. That helper started out too complicated to begin with.

Oddant1:
LGTM

PR #933, opened 2026-05-08
first draft of transitive search and multiple transformer composition

colinvwood:


src/rachis/core/transform.py
```diff
@@ -232,3 +236,178 @@ def validate(self, view, level=None):
...
+
+
+def wrap_transformer(file_type: type, sfdf_type: type):
+    '''
+    A transformer used to convert any `FileFormat` into its associated
+    `SingleFileDirectoryFormat`.
+    '''
+    def transformer(view):
+        sfdf = sfdf_type()
+        sfdf.file.write_data(view, file_type)
```
ebolyen, line 397: Just adding a note here for performance:

We should check that for owned files (rachis is the owner) we do something cheap like rename and for unowned files (user is the owner) we do something almost cheap like util.duplicate

src/rachis/core/transform.py
```diff
@@ -232,3 +236,178 @@ def validate(self, view, level=None):
...
+                    outstanding.insert(0, neighbor)
+
+    return None
+
+
+def compose_transformation(target: SearchNode | None, recorder=None):
+    if target is None:
+        return None
+
+    pm = sdk.PluginManager()
```
ebolyen, line 335: We probably want `PluginManager.reuse_existing()` but I'm not certain.

ebolyen (commented):
How will we handle the following scenerios:

## A) lossy-link
Transformers:
  X -> Y
  A -> Y
  B -> A (lossy)
  B -> X (complete)

Destination:
  B -> Y

## B) [pessimal-path](https://www.mipmip.org/tidbits/pasa.pdf)

Transformers:
  G -> H
  H -> I
  J -> K
  K -> Q
  G -> LMNOP
  LMNOP -> Q

Destination:
  G -> Q

---

Also how will we define search order? It will be somewhat coupled to registration order which can differ by filesystem.

colinvwood:
@ebolyen 

A) Good question, I'm not sure. We would have to attach metadata to transformers to be able to make a decision about the path we choose. Or, we could find all paths and warn/error if multiple are discovered. 

B) Great paper, right now we take the shortest path (breadth first) but I can refactor so we always take the longest path 😂.

In response to search order, we can sort the added neighbors in each iteration.

ebolyen:
Regarding A), something we could do is only error if the path isn't decidable. This is maybe more complicated, but we could declare that there if there is no unique decision than you need to write transformer which results in a shorter path than the previous redundant paths.

And if we can prove we take the shortest path then B) isn't an issue and A) reduces to a "duplicate transformer" registration error (which we already do).

However, some of our dependency issues come from trying to share transformers which would be better solved by allowing plugins to prefer their own transformers before falling back on the plugin manager. Doing the above locks us into no duplicate transformers or otherwise makes overriding transformers with first-party self-registered transformers really subtle. (Transformation paths become relative to the plugin interacting and how to overrides work in a pipeline calling another method? What if the redundancy is only visible from a 3rd plugin or only exists when an extra plugin is installed?)

ebolyen:
What if we did have metadata like you suggest:

```python
@plugin.register_transformer(upgrade=True)
def _0(data: OldType) -> NewType:
    return ...

@plugin.register_transformer(upgrade=False)
def _1(data: pd.Series) -> np.array:
    return ...
```
Then we only allow a transitive hop over a transformer that has a statement on lossy-ness. Upgrades are always used and downgrades are not preferred but used when there exists no other more direct connection.

I think this serves the general goal of trying to update a file-format as we are usually replacing it with something more capable, so most of these transitive hops are upgrades. Then we can imagine re-wiring the existing transformers in terms of the newer one and posses a single upgrade hop.

Then for other plugins which might use a constellation of older view types, we can retain those without necessarily rewriting transformers via a downgrade hop. Here we intentionally lose information, but are fine with that because downstream of those destinations that information never existed in the first place.

So we make these assumptions
```python
# lossy and not transitive, can only be invoked as a single step or at the terminal ends of the path.
# This is the default and what all current transformers are.
@plugin.register_transformer(upgrade=None)

# not lossy and allows transitive hops, can be chained indefinitely
# (SingleFileDirectoryFormat -> FF or vice-versa would be implicit examples of this)
@plugin.register_transformer(upgrade=True)

# lossy but transitive as a last resort for searching
@plugin.register_transformer(upgrade=False)
```

I think if we did this, then we only need to worry about duplicate paths when upgrade has been declared, which is less likely to accidentally happen as we're making intentional statements.

This also lets us describe a preferred direction for transformation.
O -> N (upgrade, good!)
N -> O (downgrade, avoid when possible)

colinvwood:
replaced by #959

PR #934, opened 2026-05-15
Add support for transforming from unions 

colinvwood:
Adds support for transforming from unions. Useful when a `TypeMatch` requires an action to return one of distinct view types. 

This mirrors what Michal added a few years ago to support transforming *into* a member of union, and factors the logic for both into a class method. 

Would be helpful for an addition @miiixi1in is making to `q2_feature_table.rename_ids` to allow us to rename feature IDs in any of `FeatureTable[*]`, `FeatureData[Taxonomy]`, `FeatureData[Sequence]`.

### DIFF
--- a/src/rachis/sdk/result.py
+++ b/src/rachis/sdk/result.py
@@ -705,7 +705,10 @@ class Artifact(Result):
 
         if cls._is_union_type(view_type):
             transformation, _ = cls._transform_to_or_from_union(
-                from_type=view_type, to_type=to_type, recorder=recorder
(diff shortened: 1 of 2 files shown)


ebolyen:
Could we perhaps move this into the transform module (both sides of the equation here)?

I feel like I should be able to use `util.transform` (I forget if its in plugin or sdk) and end up in the same place.

That would also make it easier to test (I think).

colinvwood:
replaced by #961

PR #935, opened 2026-05-18
IMP: cache save hardlink

gregcaporaso:
@Oddant1, I landed on this change while doing some benchmarking of `moshpit annotation classify-kraken2`. It doesn't seem to help much in my benchmarks, but seems like this is how we'd want this to work. Can you review and make sure I'm not missing some reason why we don't want to use `duplicate` here?


Oddant1 (approved):
I believe this ought to be good.

Oddant1 (approved):

gregcaporaso:
Thanks - any tests that should be added for this? This isn't really my neck of the code base. Otherwise, feel free to move forward with it @Oddant1 - thanks for looking!

PR #936, opened 2026-05-19
Change capture holder value

### LINKED ISSUE #924: Change already set parameter with CaptureHolder.
The current implementation of CaptureHolder lets you set a parameter that is not already set like a random seed. 
What is not possible it to change a parameter that is already set.
The PCA function of sk-learn lets you set the n_cmponents parameter with an int for how many components used, but you can also set a float from 0 to 1 and internally the number of components is used that leads to that explained variance. Then in the provenance all the user can see is the set float but not the number of components used. 
Would it be defensible to add a new method to CaptureHolder that lets you replace a parameter that was already set by a user?

Oddant1:
Closes #924

### DIFF
--- a/src/rachis/core/testing/plugin.py
+++ b/src/rachis/core/testing/plugin.py
@@ -1321,7 +1321,7 @@ dummy_plugin.methods.register_function(
        'random_seed': Int
     },
     outputs=[('seed', SingleInt)],
-    name='Run get_or_sett on the the CaptureHolder twice.',
+    name='Run get_or_set on the the CaptureHolder twice.',
     description='Runs get_or_set on the CaptureHolder object twice and asserts'
                 ' the value does not change.'
 )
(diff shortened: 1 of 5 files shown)


lizgehret:
Hey @Oddant1 is this ready for a review?

Oddant1:
@lizgehret is now

src/rachis/core/testing/plugin.py
```diff
@@ -1314,15 +1315,28 @@ def factory():
...
 
 dummy_plugin.methods.register_function(
-    function=random_seed_method_set_twice,
+    function=random_seed_method_get_or_set_twice,
     inputs={},
     parameters={
        'random_seed': Int
     },
     outputs=[('seed', SingleInt)],
+    name='Run get_or_sett on the the CaptureHolder twice.',
```
lizgehret:
```suggestion
    name='Run get_or_set on the the CaptureHolder twice.',
```

lizgehret (commented):
Thanks @Oddant1! The code changes all look reasonable to me, but i'd like to chat about this briefly during the eng meeting just to understand the underlying purpose.

Issue #864, opened 2025-08-21
large Artifacts inflate provenance size when used as Metadata


gregcaporaso:
This came up in a Whisper discussion on the forum. Briefly, if a large artifact is used as metadata, that large artifact will be stored in the provenance of all downstream results, making all of them large. 

A few possible solutions:
- create a `redact-metadata` tool that removes metadata from artifacts. this could be useful for other reasons (e.g., a user accidentally includes private information in metadata and doesn't want to redo their whole analysis to remove it). This is probably the path of least resistance. 
- create a tool that extracts ids from a Metadata-viewable artifact into a `ImmutableMetadata` artifact, so that that can be used in place of the original Artifact - this won't solve it for all cases (i.e., if something other than the ids are needed), but might be the most common related use case. The challenge here is retaining the link to provenance.

PR #941, opened 2026-06-12
Adds method: `redact_metadata`

Macabe222:
# Description
Finds and removes all metadata files in an artifact.
Addresses #864 
Linked to qiime2/q2cli#399

# AI Disclosure

- [X] NO AI USED.
- [ ] AI USED.


ebolyen:
A few ideas:

- Should this have an accompanying artifact annotation? It seems like we may want to have more granular selection (remove the metadata from this upstream action in my provenance) which would be well suited to the way the annotations work. I am imagining the annotation is a kind of meta-provenance here to record that a masking event happened and to whom. If they were annotations we would also have the option to apply the mask over sibling artifacts without having to pinpoint anything a second time. This isn't a feature yet, but I think it's in scope for how annotations could work and would be a good use case here (cc @lizgehret) 
- The functionality should probably go in Archiver as I think Result is better treated as an API surface for users/interfaces rather than possessing too much implementation on its own. Also if we ever had a zip-FUSE aware archiver, it would need to do a slightly different thing to mask the data.
- I think there's a checksum format parser already which you can use in core.archive.util or something nearby.

Painful things:
- Provenance replay will need to know about the masking. I wonder if it would be any easier to leave the metadata file but have it be empty. I think it will really depends on how prov-lib thinks about things. If it was empty we could better justify not bumping our format to a new major version.

lizgehret:
> A few ideas:
> 
> * Should this have an accompanying artifact annotation? It seems like we may want to have more granular selection (remove the metadata from this upstream action in my provenance) which would be well suited to the way the annotations work. I am imagining the annotation is a kind of meta-provenance here to record that a masking event happened and to whom. If they were annotations we would also have the option to apply the mask over sibling artifacts without having to pinpoint anything a second time. This isn't a feature yet, but I think it's in scope for how annotations could work and would be a good use case here (cc @lizgehret)
> 
> * The functionality should probably go in Archiver as I think Result is better treated as an API surface for users/interfaces rather than possessing too much implementation on its own. Also if we ever had a zip-FUSE aware archiver, it would need to do a slightly different thing to mask the data.
> 
> * I think there's a checksum format parser already which you can use in core.archive.util or something nearby.
> 
> Painful things:
> 
> * Provenance replay will need to know about the masking. I wonder if it would be any easier to leave the metadata file but have it be empty. I think it will really depends on how prov-lib thinks about things. If it was empty we could better justify not bumping our format to a new major version.

I agree that it would probably be the least painful thing to do to just wipe the metadata file contents and then leave that file itself in the structure (as @ebolyen said, avoiding the need to bump to a new major version). Then I'd imagine re-running the checksum calculation after an annotation gets added - maybe it's just a note that states that metadata was masked/wiped for performance or data privacy reasons.

Macabe222:
@ebolyen I was wondering if you preferred adding annotations to every upstream artifact, or adding one annotation to the root and referencing the upstream ones.

src/rachis/core/archive/archiver.py
```diff
@@ -494,9 +502,30 @@ def get_checksums(self):
...
 
+    def write_checksums(self, checksums):
+        with open(self.root_dir / self._fmt.CHECKSUM_FILE, 'w') as fh:
+            for k, v in checksums.items():
+                fh.write(f'{v}  {k}\n')
+
     def has_checksums(self):
         return hasattr(self._fmt, 'CHECKSUM_FILE')
 
+    def _add_annotation(self, annotation, annotation_dir):
```
Oddant1: Don't love the addition of this necessitated by putting redact on `Archiver` not `Result`

src/rachis/core/archive/archiver.py
```diff
@@ -521,3 +550,76 @@ def validate_checksums(self):
...
+
+        metadata_paths = []
+        relative_metadata_paths = []
+        for action_yaml in Path(self.provenance_dir).rglob('action.yaml'):
+            with open(action_yaml) as fh:
+                metadata = yaml.safe_load(fh)
+                metadatas = get_metadata_objects(metadata)
+                if metadatas is None:
+                    raise ValueError(
+                        'Cannot redact metadata from an artifact that does '
```
Oddant1:
```suggestion
                        'Cannot redact metadata from an artifact that does not '
```

src/rachis/core/archive/archiver.py
```diff
@@ -521,3 +550,76 @@ def validate_checksums(self):
...
+                    )
+                for metadata in metadatas:
+                    metadata_paths.append(
+                        action_yaml.parent / metadata.relative_fp
+                    )
+                    relative_metadata_paths.append(metadata.relative_fp)
+
+        # Empty metadata files
+        for metadata_path in metadata_paths:
+            open(metadata_path, 'w').close()
```
Oddant1, line 926: 🤓

Oddant1 (commented):
Initial quick review

@ebolyen, I understand wanting this functionality on `Archiver` not `Result`; however, it has created the need for some redundant code that I don't love. Maybe acceptable though.

Oddant1:
Maybe we do some refactoring of `Result.add_annotation` and `Archiver._add_annotation` to remove duplicated functionality and have `Result` instances call the method on their own `Archiver` instance.

src/rachis/core/archive/archiver.py
```diff
@@ -494,6 +503,37 @@ def get_checksums(self):
         with open(self.root_dir / self._fmt.CHECKSUM_FILE) as fh:
             return dict(from_checksum_format(line) for line in fh.readlines())
 
+    def extract_uuid_in_provenance(self, file):
+        return re.findall(r'provenance/artifacts/([0-9a-f-]{36})/', file)
+
+
+    def _add_annotation(self, annotation, reference_uuid = None):
```
Oddant1: I think we want to put the implementation of `Result.add_annotation` here if we're going to have `redact_metadat` live on `Archiver`. Then `Result.add_annotation` would just call this to maintain the API

Macabe222: Would this include moving other annotation methods to `Archiver` such as `get_annotation` and so forth?

Oddant1: Probably. Duplicating this functionality across the two classes doesn't sit right nor does spreading it, and because the `Archiver` doesn't have access to the `Result` if we want `redact_metadata` to live here, as @ebolyen suggested, then we need at least some of the functionality to live here, as you noticed. I really feel like if it needs to live here, it should all live only here.

Philosophically, the annotations are attached to the `Result`, but they live on disk, and the on-disk representation of the `Result` is managed by the `Archiver`

Oddant1:
@ebolyen @lizgehret thoughts on moving the functionality of annotation onto `Archiver` like this?

src/rachis/core/archive/archiver.py
lizgehret: @Oddant1 I don't think that's unreasonable and tbh I remember there being some difficulties in our initial implementation because things didn't live under the Archiver. I agree though that duplicating behavior doesn't sit right, and it's probably best for everything to live in one place.

ebolyen:
Just wanted to pop in real quick: let's move things to `Archiver` (as discussed in our meeting yesterday), but we should also keep methods which forward to that implementation on `Result` as well (same as many of the other methods on `Result`).

PR #950, opened 2026-07-01
MAINT: Rename `NP_RNG_SIZE` to `NP_RNG_BITS`

Oddant1:
# Description

Better name for this var.

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.

### DIFF
--- a/src/rachis/plugin/__init__.py
+++ b/src/rachis/plugin/__init__.py
@@ -9,7 +9,7 @@
 from .model import (TextFileFormat, BinaryFileFormat, DirectoryFormat,
                     ValidationError, SingleFileDirectoryFormat)
 from .plugin import Plugin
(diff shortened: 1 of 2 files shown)


lizgehret (approved):
Simple change - this is fine to merge as-is.

For posterity, this passed even though its partner PR failed- which made me realize that rachis PRs will never work with the `Paired-` functionality as the current CI doesn't run off of `lib-ci-dev` anymore. We may want to add this workflow back in to trigger on any branch names that match `Paired-` in order to support this in the framework.

lizgehret:
ok circling back to this - we don't need to care that the Paired functionality will never work for rachis, since it's not dependent on anyone else to begin with. so this is fine as-is; i'm going to merge this and its partner, thanks @Oddant1!

Issue #294, opened 2017-06-28
Identify when a format is empty


ebolyen:
If a given sniffer/validator for a format fails, it would be pretty handy to check to see if there is anything in the file and report whether it was empty or not. It would have made debugging [this issue](https://github.com/qiime2/q2-alignment/issues/34) much simpler.

This might tie into the more general notion of "why" a format has failed a check, but it seems like a common enough situation to potentially bake-in.

PR #956, opened 2026-07-21
Adds check for empty file in formats.

Macabe222:
# Description

Addresses #294 

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.

# AI Usage Details

### DIFF
--- a/src/rachis/core/format.py
+++ b/src/rachis/core/format.py
@@ -10,7 +10,7 @@ import rachis.core.path as qpath
 
 
 class FormatBase:
-    def __init__(self, path=None, mode='w'):
+    def __init__(self, path=None, mode='w', allow_empty=False):
         import rachis.plugin.model as model
         if path is None:
             if mode != 'w':
@@ -27,6 +27,7 @@ class FormatBase:
             self.path = qpath.InPath(path)
 
         self._mode = mode
+        self.allow_empty = allow_empty
 
     def __str__(self):
         return str(self.path)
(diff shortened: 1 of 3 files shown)


ebolyen:
I believe there is one or more formats in q2-types which expect empty files. Maybe we can add a parameter to the File property to `allow_empty` which will skip this warning.

I don't remember which formats this applies to any more unfortunately, so a bit of research is needed.

src/rachis/core/format.py
```diff
@@ -10,7 +10,7 @@
 
 
 class FormatBase:
-    def __init__(self, path=None, mode='w'):
+    def __init__(self, path=None, mode='w', allow_empty=False):
```
ebolyen, line 13: I think this shouldn't be in the constructor.

It isn't different from setting all file attributes to `optional=True` in the case of a directory format, and it's a little awkward for a single format (we could make a class-attribute if needed).

Since only the directory format validator is currently implicated, we should make it work similarly to `optional` on `File()` (I'll point it out below).

src/rachis/plugin/model/tests/test_directory_format.py
```diff
@@ -44,6 +44,9 @@ class OptionalDirFmt(model.DirectoryFormat):
     file3 = model.File(r'test_text3.txt', format=IntSequenceFormat,
                        optional=True)
 
+class SingleFileDirFmt(model.DirectoryFormat):
+    file1 = model.File(r'test.txt', format=IntSequenceFormat, optional=False)
```
ebolyen, line 48: This is where I would expect `allow_empty=True/False`

ebolyen (commented):
Question: Are we barking up the wrong tree here? Should we have a built-in validation on `FileFormat` which raises the warning?

Every file in the directory format should be collected and checked (unless it has `optional`), so we can let the format dispatch its own empty-ness logic. This could be set via a class-attribute or as the "pseudo-self" instance from the `model.BoundFile` via `model.File()` constructor. (I'm really sorry, the way these objects work to make the DSL is a bit rough.)

Issue #139, opened 2016-09-13
importing should ignore hidden files/dirs


jairideout:
It's common to have hidden files (e.g. .DS_Store) in a directory that is being imported. The importers should ignore these files (currently they error).

vitmy0000:
Hi, I'm a heavy Qiime user and want to contribute to qiime2.

Can you be more specific on this issue?
Is this a general python problem?

ebolyen:
Hey @vitmy0000, sorry for the confusion. This issue isn't about the python `import` statement at all. It is actually about "importing" data (sequences files, biom tables, etc...) into an QIIME 2 Artifact (all of this is very new vocabulary for QIIME, check out our [wiki](2.qiime.org) for more). 

The dotfiles are important to consider here because if you happen to open a folder in OS X Finder it will add a `.DS_Store` to the directory, which will confuse our framework machinery (as it sees a file it thinks shouldn't exist, therefore it isn't in the format it expects).

As far as getting involved, I would recommend joining our [slack team](https://qiime2-slackin.herokuapp.com/). Jumping into this particular issue is probably a little too deep to start out with, and we intend to heavily modify some of the machinery soon.

PR #957, opened 2026-07-21
MAINT: Ignore dotpaths during validation

colinvwood:
# Description
Addresses #139.

This is in line with the `_ZipArchive.save` logic --ignore all paths that contain a hidden directory. It's technically true that dotfiles and dotdirectories can hang around in the cache between `Artifact.import_data` and `Artifact.save`, not sure if we care about that.

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.


src/rachis/sdk/tests/test_artifact.py
```diff
@@ -388,6 +388,55 @@ def test_import_data_with_unrecognized_files(self):
...
+        dotfile = test_dir / '.dotfile'
+        dotfile.touch()
+
+        dotdir = test_dir / '.dotdir'
+        os.mkdir(dotdir)
+
+        nested = test_dir / '.dotdir' / 'nested'
+        nested.touch()
+
+        Artifact.import_data(IntSequence1, test_dir)
```
Oddant1, line 412: So we know the dotfiles were ignored because if they weren't this would have failed right? Maybe throw that in a comment above this `import_data` in both tests. Maybe also throw an `assert(True)` at the end of the test or something idk. We do those kind of dummy assertions in some places, but I don't think they're mandatory.


```suggestion
        # If dotfiles weren't ignored, this would fail.
        Artifact.import_data(IntSequence1, test_dir)
```

src/rachis/sdk/tests/test_artifact.py
```diff
@@ -388,6 +388,55 @@ def test_import_data_with_unrecognized_files(self):
...
+        dotfile = test_dir / '.dotfile'
+        dotfile.touch()
+
+        dotdir = test_dir / '.dotdir'
+        os.mkdir(dotdir)
+
+        nested = test_dir / '.dotdir' / 'nested'
+        nested.touch()
+
+        Artifact.import_data(SingleInt, test_dir)
```
Oddant1, line 438: ### 
```suggestion
        # If dotfiles weren't ignored, this would fail.
        Artifact.import_data(SingleInt, test_dir)
```

PR #959, opened 2026-07-28
Transitive transformers

Macabe222:
# Description

Makes transformers transitive.
For context read discussion in #933.

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.


Macabe222:
@ebolyen What are your thoughts on capturing the transformers in provenance? Currently it only lists the first and last transformer used. @colinvwood's thoughts are to list each step and the plugin from which that transformer came from.

ebolyen:
Yep! I think it should already be a yaml list in the action.yml to support this. So we just generate the path used, and in many cases it's just a single hop [start, end]. We can change the provenance transform recorder as needed, because that's an internal API.

colinvwood:
It looks like we use the `plugin` key to refer to the plugin in which the *transformed-to type* is registered. However sometimes we will make hops across types that are not registered (because they are not directory formats). Should we in such cases report the plugin the *transformer* is registered in instead?

src/rachis/core/transform.py
```diff
@@ -63,16 +63,16 @@ def make_transformation(self, other, recorder=None):
...
-                output_record=other._record
-            )
+        # if recorder is not None:
+        #     recorder(
+        #         target_node.record,
+        #         input_name=self._view_name,
+        #         input_record=self._record,
+        #         output_name=other._view_name,
+        #         output_record=other._record
+        #     )
```
colinvwood: 🧐

src/rachis/core/archive/tests/test_citations.py
```diff
@@ -16,6 +16,7 @@
 class TestCitationsTracked(unittest.TestCase):
     def setUp(self):
         self.plugin = get_dummy_plugin()
+        self.maxDiff = None
```
colinvwood: 🧐

ebolyen:
Yes, I think it should usually be the plugin that the transformer is registered to.

src/rachis/sdk/result.py
```diff
@@ -461,10 +461,6 @@ def view(self, view_type):
         return self._view(view_type)
 
     def _view(self, view_type, recorder=None):
-        if view_type is rachis.Metadata and not self.has_metadata():
```
colinvwood, line 464: What's this about?

Macabe222: Some paths might not have metadata now, for example `ANCOMBC2OutputDirFmt` -> `Metadata`

src/rachis/core/transform.py
```diff
@@ -5,36 +5,23 @@
...
+if TYPE_CHECKING:
+    from rachis.plugin.plugin import TransformerRecord
 
 from rachis import sdk
 from rachis.plugin import model
 from rachis.core import util
 
 
-def identity_transformer(view):
-    return view
```
Oddant1: I personally think our transformation system should get stuck in a loop of transforming things into themselves.

src/rachis/core/transform.py
```diff
@@ -146,89 +96,439 @@ def set_user_owned(self, view, value):
...
+                return True
+
+            current = current.parent
+
+        return False
+
+
+class NodeQueue:
+    def __init__(self):
+            self.nodes = []
```
Oddant1:
```suggestion
        self.nodes = []
```

src/rachis/core/archive/tests/test_citations.py
```diff
@@ -39,7 +39,7 @@ def test_import(self):
         obs = list(map(lambda item: (item[0], item[1].fields['title']),
                        archiver.citations.items()))
 
-        self.assertEqual(obs, expected)
+        self.assertEqual(sorted(obs), sorted(expected))
```
Oddant1, line 42: Why was this needed?

src/rachis/core/transform.py
```diff
@@ -146,89 +96,439 @@ def set_user_owned(self, view, value):
...
+            continue
+
+        if current.type_ == target:
+            return current
+
+        node_queue.insert_neighbors(current)
+
+
+def compose_transformation(
+    target: SearchNode | None,
```
Oddant1, line 404: Is this ever going to be called with None in practice? Seems like the only call site raises an exception if this is None

src/rachis/core/transform.py
```diff
@@ -146,89 +96,439 @@ def set_user_owned(self, view, value):
...
+        The transformation provenance recording function. None if caller does
+        not care about provenance, e.g. `Artifact.view`.
+
+    Returns
+    -------
+    Callable
+        The composed transformer closure.
+    '''
+    if target is None:
+        return None
```
Oddant1, line 426: Again, not sure if target will ever be None in practice, but if so this should probably return a no-op callable matching the transformer signature not a None since I expect any call site will be expecting a callable

src/rachis/core/archive/tests/test_citations.py
colinvwood: importing involves transformations which have citations, the transformation machinery changed so the order of citation recording changed I assume (Macabe made this change)

src/rachis/core/transform.py
colinvwood: Yeah seems like there's no reason for the none handling.

src/rachis/core/archive/tests/test_citations.py
Oddant1: I find it curious that the order would change but not the actual content, but yeah, that's probably it.

colinvwood: @Macabe222 do you remember why this change was needed?

PR #962, opened 2026-08-05
`replay-provenance` works on reports

### LINKED ISSUE #930: prov-replay does not work on reports
Currently prov-replay does not work on reports, because the report action does not have a plugin.

We should check the output type, and if its a report, we should just skip the action.

Macabe222:
# Description

Addresses #930 

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.


colinvwood:
@cherman2 can you user test this?

Issue #423, opened 2019-01-03
Provide brief information about an artifacts provenance from `peek`


wasade:
**Improvement Description**
Currently, `qiime tools peek` provides information about the semantic type of an artifact. However, this does not provide an indication from what plugin or action generated the artifact. For visualization artifacts, they do not have a semantic type leading the output of `peek` and presently only report that they are `Visualization`. 

The seed of this issue is that, with the current output of `qiime tools peek`, a user must rely on filenames (or the act of viewing with q2view) to determine what a file is. 

After discussing on Slack with @ebolyen and @thermokarst, we thought it may be useful to report the last action associated with an artifact (e.g., `q2_emperor.plot`). In general, this could help to provide sufficient information to the user to disambiguate what a file is. 

**Current Behavior**
The current behavior reports the type and data format (if a data artifact), but the last action performed is not provided. In the below example, it is not known whether `a_table.qza` came from `q2-deblur`, `q2-dada2`, `q2-feature-table`, or another method which can produce `FeatureTable[Frequency]`. Similarly, for the `a_plot.qzv`, the user cannot determine what type of plot this may represent in the absence of the last associated action.

```python
$ qiime tools peek a_table.qza
UUID:        75023169-d424-4092-b9c5-30472f803970
Type:        FeatureTable[Frequency]
Data format: BIOMV210DirFmt
$ qiime tools peek a_plot.qzv
UUID:        007ec0f6-b7cf-4f15-ad88-b3ec72baf117
Type:        Visualization
```

**Proposed Behavior**
A possible solution would be to include the last executed action performed which generated the artifact.

```python
$ qiime tools peek a_table.qza
UUID:        75023169-d424-4092-b9c5-30472f803970
Type:        FeatureTable[Frequency]
Data format: BIOMV210DirFmt
Last action: deblur.denoise_16s
$ qiime tools peek a_plot.qzv
UUID:        007ec0f6-b7cf-4f15-ad88-b3ec72baf117
Type:        Visualization
Last action: emperor.plot
```

PR #964, opened 2026-08-11
`peek` reports last used action

Macabe222:
# Description

Addresses #423 
Linked to qiime2/q2cli#408

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.

### DIFF
--- a/src/rachis/core/archive/archiver.py
+++ b/src/rachis/core/archive/archiver.py
@@ -31,7 +31,7 @@ from rachis.core.util import (
 )
 from rachis.core.archive.format.v0 import ArchiveFormat
 from rachis.core.archive.provenance import (
-    metadata_path_constructor, MetadataInfo
+    metadata_path_constructor, MetadataInfo, ZipActionYamlLoader
 )
 from rachis.core.annotate import Note, Annotation, ANNOTATION_TYPE_DICT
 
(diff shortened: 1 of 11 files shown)


src/rachis/core/archive/archiver.py
```diff
@@ -267,6 +272,13 @@ def mount(self, path):
         return ArchiveRecord(root, root / self.VERSION_FILE,
                              self.uuid, self.version, self.framework_version)
 
+    def load_action_yaml(self):
+        action_path = self.path / 'provenance' / 'action' / 'action.yaml'
+        with open(action_path) as fh:
+            prov = yaml.safe_load(fh)
+
+        return prov
```
Oddant1: Make this and the other `load_action_yaml` look the same by either returning from inside the with block like this from both of them, or saving to a variable and returning the variable in both of them.

```suggestion
            return yaml.safe_load(fh)
```

src/rachis/core/util.py
```diff
@@ -430,42 +430,34 @@ def touch_under_path(path):
                 pass
 
 
-def load_action_yaml(path):
-    """Takes a path to an unzipped Artifact and loads its action.yaml with
+def load_action_yaml_zipped(archive):
```
Oddant1: @colinvwood pointed out that this name is bad. It MIGHT be zipped, but it isn't necessarily (this function works either way though). This is why it's better to have these on the _Archiver classes, which requires changes to q2cli to use those methods not the functions here.

PR #968, opened 2026-08-17
NEW: Add the ability to not record prov for specified actions in pipelines

### LINKED ISSUE #963: Create a Mechanism for Not Recording Some Provenance in Pipelines
We are having issues with some pipelines, mostly the resample pipelines in [q2-boots](https://github.com/qiime2/q2-boots), creating a large amount of provenance we don't actually care about, e.g., storing all the prov for 1000 resampled tables and copying that prov across many other artifacts.

We need to create a mechanism to avoid storing this provenance. The current plan is to create an API that goes as follows:

```python
def pipeline(ctx, ...)
    # Add this "record_prov" keyword argument to get_action. Defaults to True
    foo = ctx.get_action('bar', 'foo', record_prov=False)
    
    # Now any calls to this executable do not record provenance
    results, = foo(...)
```

The underlying effects of invoking this API are still a bit up in the air. The ideal is that actions invoked in this way will produce stubs that only contain `uuid->data->...`. Just enough to actually load the data. At the very least, this API will create artifacts where `uuid->provenance->artifacts` is empty.

Oddant1:
# Description
Add the `record_provenance` flag to `Context.get_action` and the `NoOpProvenanceCapture` allowing for specific provenance to not be recorded.

Closes #963   

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.

# AI Usage Details

### DIFF
--- a/src/rachis/core/archive/provenance.py
+++ b/src/rachis/core/archive/provenance.py
@@ -694,6 +694,9 @@ class NoOpProvenanceCapture(ProvenanceCapture):
     def __init__(self, *args, **kwargs):
         pass
 
+    def _build_paths(self):
+        pass
+
     def add_ancestor(self, *args, **kwargs):
         pass
 
--- a/src/rachis/core/archive/provenance.py
+++ b/src/rachis/core/archive/provenance.py
@@ -327,6 +327,8 @@ class ProvenanceCapture:
             # NOTE: this implies that it is possible for an action.yaml file to
             # contain an artifact UUID that is not in the artifacts/ directory.
             return NoProvenance(artifact.uuid)
+        elif not os.path.exists(other_path):
+            return NoProvenance(artifact.uuid)
 
         destination = self.ancestor_dir / str(artifact.uuid)
         # If it exists, then the artifact is already in the provenance
--- a/src/rachis/core/archive/provenance.py
+++ b/src/rachis/core/archive/provenance.py
@@ -684,3 +684,63 @@ class ReportProvenanceCapture(ProvenanceCapture):
         action['inputs'] = self.inputs
 
         return action
+
+
+# TODO: Currently this is creating a metadata.yaml and a conda-env.yaml from
+# the archiver, and that's it. So the internal artifacts are in the returned
+# artifact's provenance/artifacts dir and just look like uuid/metadata.yaml +
+# conda-env.yaml
+class NoOpProvenanceCapture(ProvenanceCapture):
+    def __init__(self, *args, **kwargs):
+        pass
+
+    def add_ancestor(self, *args, **kwargs):
+        pass
+
+    def make_citation_key(self, *args, **kwargs):
+        pass
+
+    def make_software_entry(self, *args, **kwargs):
+        pass
+
+    def reference_plugin(self, *args, **kwargs):
+        pass
+
+    def capture_env(self, *args, **kwargs):
+        pass
+
+    def transformation_recorder(self, *args, **kwargs):
+        def recorder(*args, **kwargs):
+            pass
+
+        return recorder
+
+    def make_execution_section(self, *args, **kwargs):
+        pass
+
+    def make_transformers_section(self, *args, **kwargs):
+        pass
+
+    def make_env_section(self, *args, **kwargs):
+        pass
+
+    def write_action_yaml(self,*args, **kwargs):
+        pass
+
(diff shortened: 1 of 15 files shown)


src/rachis/plugin/context.py
```diff
@@ -10,7 +10,7 @@
 
 class IContext(metaclass=abc.ABCMeta):
     @abc.abstractmethod
-    def get_action(self, plugin: str, action: str):
+    def get_action(self, plugin: str, action: str, record_prov: bool=True):
```
ebolyen: It's probably worth spelling out, since it will be odd to see and somewhat rare: `record_provenance`

I would also make it a keyword only arg (add a bare `*,` after `action: str,`)

src/rachis/sdk/context/base.py
```diff
@@ -30,7 +30,9 @@ def _validate_collection(collection_order):
 
 
 class Context(IContext):
-    def __init__(self, action_obj=None, parent=None):
+    def __init__(self, action_obj=None, parent=None, record_prov=True):
+        self.record_prov = record_prov
```
ebolyen: I would probably make this private for now (no issue with `action.py` checking the state though).

src/rachis/sdk/context/base.py
```diff
@@ -188,3 +192,8 @@ def add_reference(self, ref):
 
         # Return an artifact backed by the data in the cache
         return new_ref
+
+    # PROV TODO: Another option for prov is to capture it and then drop it at
+    # the end of the pipeline
+    def drop_inner_provenance(self, result, provenance):
```
ebolyen: I think we won't need this, but if we did, the API would need to look a bit different (unless this is private).

src/rachis/sdk/action.py
```diff
@@ -273,8 +273,11 @@ def _bind(self, context_factory, execution_ctx={'type': 'synchronous'}):
         """
         def bound_callable(*args, **kwargs):
             ctx = context_factory()
-            provenance = self._ProvCaptureCls(
-                self.type, self.plugin_id, self.id, execution_ctx)
+            if ctx.record_prov:
+                provenance = self._ProvCaptureCls(
+                    self.type, self.plugin_id, self.id, execution_ctx)
+            else:
+                provenance = archive.NoOpProvenanceCapture()
```
ebolyen, line 280: This is pretty nice

src/rachis/core/archive/provenance.py
```diff
@@ -684,3 +684,63 @@ def make_action_section(self):
         action['inputs'] = self.inputs
 
         return action
+
+
+# TODO: Currently this is creating a metadata.yaml and a conda-env.yaml from
+# the archiver, and that's it. So the internal artifacts are in the returned
+# artifact's provenance/artifacts dir and just look like uuid/metadata.yaml +
+# conda-env.yaml
```
ebolyen, line 694: Should we change the ancestor behavior so that an alias of a NoOpProv result does nothing as well (i.e. provenance/artifacts/ isn't populated with the no-prov results)?

It seems like it might already do this via `fork()`, but I'm not certain.

ebolyen (changes requested):
Exciting!

src/rachis/core/archive/provenance.py
Oddant1: `fork()` does accomplish this.

Oddant1: I misread your comment. I don't think `fork()` fully accomplishes. I am looking back at this for not only boots but https://github.com/LadnerLab/q2-PSEA/pull/28. I'll see if I can sort out this behavior... it's not clear to me what we even want left of these artifacts at all... I suppose absolutely nothing.

Oddant1: @ebolyen ok looking back at it, we are most of the way there. When I run a pipeline with `record_provenance=False` on all actions, then look at /provenance/artifacts from an output from that pipeline the only things in provenance are

1. Full prov for all Artifacts input to the pipeline
2. Stubby prov (only metadata.yaml and conda.yaml) for the Artifact the returned Artifact is aliasing

Without no prov, there are a bunch of Artifacts from other steps in the pipeline in there

ebolyen: Awesome, that's definitely a substantial improvement!

